PR #18, opened 2018-04-13
Power Vertex - Accept Vertex as hyper parameter

GeorgeNash:
To be able to raise a DoubleVertex to the power of another DoubleVertex.


chris-major-improbable:
This looks like a breaking change to the API (change to the package for the power vertex).
Do we have a policy for anouncing/documenting changes that will break existing code on an upgrade?

gordoncaleb:
For the releases < 1.0.0, we'll write in the release notes that a non-backwards-compatible change happened and provide a recommended approach for upgrading. When we get to a point where we are fairly confident in the structure and direction of the code, we will go +1.x.x and non-backward-compatible changes will be accompanied by a major version bump. Until then, I'd like for us to have the power to change what we need to change. This implies that we wouldn't recommend running keanu in production until > 1.0.0.

PR #19, opened 2018-04-13
Extended randomFactory interface

dmculley:


GeorgeNash:
Nice one - as you've changed the interface you'll need to add these new methods to their implementations - `DoubleVertexFactory` and `RandomDoubleFactory`, that's why the build is breaking. 

Just chuck this method into both of the above.

`  @Override
    public DoubleVertex nextGaussian(double mu, DoubleVertex sigma) {
    return new GaussianVertex(mu, sigma, random);
    }`

For future reference you can see build status by running `./gradlew build` before committing or just look for the green tick or red cross next to each commit message: https://github.com/improbable-research/keanu/commits/feature/randomFactoryGeneralisation.

gordoncaleb:
Can you add a brief description of what the PR does? And why?

dmculley:
I'll extend the class as I need it inside Swayze and raise a JIRA ticket linking to it for inclusion in Keanu in due course

PR #29, opened 2018-04-17
Chi2 Vertex

GeorgeNash:
Bit of a strange one this.

Chi2 has only one parameter `k`, which is an integer. This is a bit of a nightmare for calculating the derivative (Can't be done without severe **mathematical 'estimation'**) , so I've left it out. 

Raises the bigger question of how do we want to handle the differential calculation of discrete variables. Project into continuous space?


GeorgeNash:
Thoughts on renaming this to ChiSquare?

PR #31, opened 2018-04-18
Point to Keanu 0.0.4 and actually use the data in the Starter Project

gordoncaleb:
The initial release included a starter project that loaded data and did nothing with it. This PR sets the starter project to use the externally released Keanu from Maven Central and updates the project to actually use the data loaded in the starter project.


GeorgeNash:
Looks like the `NetworkSamples results` has been taken out of `Model` in 0.0.4. Do you want to chuck it back in?

GeorgeNash:
Nice one.

PR #33, opened 2018-04-19
Fix bug in the Hamiltonian implemtation

gordoncaleb:
This PR fixes an issue with the leap frog calculation as well as an error in the accept/reject sample logic in the HMC implementation.

This also adds a simple histogram and scatter plot visualizer for debugging.


keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
```diff
@@ -28,185 +26,143 @@ public static NetworkSamples getPosteriorSamples(final BayesNet bayesNet,
...
+                    logOfMasterPBeforeLeapfrog,
+                    momentumBeforeLeapfrog
             );
 
-            if (shouldAccept(likelihoodOfLeapfrog, random)) {
-                addAsLatest(samples, position);
-            } else {
-                addAsLatest(samples, positionPreviousTimeStep);
+            if (shouldReject(likelihoodOfLeapfrog, random)) {
+                revertToPosition(latentVertices, positionBeforeLeapfrog);
```
GeorgeNash: As mentioned this morning, what do you think would work best?

1. Reverting the values by recalculating them
2. Store a 'cached' version of the BayesNet itself and swap it back on a reject

PR #35, opened 2018-04-23
Efficient set and cascade

gordoncaleb:
Previously, set and cascade of values would evaluate parts of the graph multiple times (potentially many many times). This PR implements an algorithm that ensures that the minimal amount of work is done. It does this by first traversing the graph and caching the relevant information for subsequent efficient graph evaluations. 

Also, probabilistic vertices no longer set their value from a sample upon creation. This avoids problematic lazy evaluation of uninitialised parent vertices and ensures that work is only done when the user specifies.


keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
```diff
@@ -55,14 +57,16 @@ public static NetworkSamples getPosteriorSamples(final BayesNet bayesNet,
...
+                                 final Map<String, Map<String, Integer>> setAndCascadeCache,
                                  final Random random) {
 
         final double affectedVerticesLogPOld = sumLogP(affectedVertices);
 
         final T oldValue = chosenVertex.getValue();
         final T proposedValue = chosenVertex.sample();
 
-        chosenVertex.setAndCascade(proposedValue);
+        Map<String, Integer> cascadeCache = setAndCascadeCache.computeIfAbsent(chosenVertex.getId(), (id) -> chosenVertex.exploreSetting());
```
GeorgeNash: My gut instinct is that we can do something a bit sneaky here and avoid having to `exploreSetting` for each Vertex. I see this as being quite a large compute if we have a load of vertices? What about...

1. `exploreSetting` for a Vertex, A, and store it in the map
2. `exploreSetting` for a second Vertex, B, that is `upstream` of A. (B is a parent of A). When we are exploring and find a Vertex that we have already computed, in this case A, can we take the already computed values of how many times vertices will be visited and just add them on to our exploration from B?
3. Add Vertex B's `exploredSetting` to the map 
3. `exploreSetting` for a third Vertex, C, that runs in to B. Use this saved propagation count...
4. etc

TLRD; Sort of like a cached routing problem. Once we hit a vertex we've already calculated the propagation count for, just use the values again as they will be the same from that point as previously computed?

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/SimulatedAnnealing.java
```diff
@@ -63,7 +65,7 @@ public static NetworkState getMaxAPosteriori(BayesNet bayesNet,
 
             double T = annealingSchedule.getT(sampleNum);
             Set<Vertex<?>> affectedVertices = affectedVerticesCache.get(chosenVertex);
-            logP = MetropolisHastings.nextSample(chosenVertex, logP, affectedVertices, T, random);
+            logP = MetropolisHastings.nextSample(chosenVertex, logP, affectedVertices, T, setAndCascadeCache, random);
```
GeorgeNash: See comment above on the Map of Map's

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/FitnessFunction.java
```diff
@@ -24,10 +28,11 @@ public MultivariateFunction fitness() {
 
     protected void setAndCascadePoint(double[] point) {
         for (int i = 0; i < point.length; i++) {
-            latentVertices
-                    .get(i)
-                    .setAndCascade(point[i]);
+            Vertex<Double> v = latentVertices.get(i);
```
GeorgeNash: nitpick, can we call this `vertex` not `v` to keep it consistent with the method below?

keanu-project/src/main/java/io/improbable/keanu/network/BayesNet.java
```diff
@@ -124,19 +125,19 @@ public boolean isInImpossibleState() {
...
+        for (Vertex<?> vertex : vertices) {
+            setValueFromSample(vertex);
+        }
+        VertexValuePropagation.cascadeUpdate(vertices, setAndCascadeCache);
     }
 
-    public static void cascadeValues(List<? extends Vertex<?>> vertices) {
-        vertices.forEach(BayesNet::cascadeValue);
+    private static <T> void setValueFromSample(Vertex<T> v) {
+        v.setValue(v.sample());
```
GeorgeNash: Same nitpick as above, this is called `vertex` throughout the rest of the class

keanu-project/src/test/java/io/improbable/keanu/vertices/VertexValuePropagationTest.java
```diff
@@ -0,0 +1,121 @@
...
+    @Test
+    public void doesPropagateAroundProbabilisticVertices() {
+        AtomicInteger n = new AtomicInteger(0);
+        DoubleVertex firstLayerStart = new GaussianVertex(0, 1, random);
+
+        DoubleVertex firstLayerEnd = addLinks(firstLayerStart, n, 1);
+
+        DoubleVertex secondLayerStart = new GaussianVertex(firstLayerEnd, 1, random);
+
+        DoubleVertex secondLayerLeft = sumVertex(secondLayerStart, firstLayerEnd, n, id -> log.info("OP on id:" + id));
```
GeorgeNash: Chuck a space after the `:` for formatting?

GeorgeNash:
Nice work!

PR #37, opened 2018-04-23
Lazy Eval Propagation

GeorgeNash:
Improvements to `lazyEval`.

Ensures that no vertices are evaluated more than once when calling `lazyEval`.

I stole a lot of the testing from: https://github.com/improbable-research/keanu/pull/35

 We can wait for that to be closed and then factor out the shared code before merging this in


keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
```diff
@@ -99,7 +99,27 @@ public double logDensityAtValue() {
...
+
+            Vertex<?> head = stack.peek();
+            Set<Vertex<?>> parents = head.getParents();
+
+            if (parents.isEmpty() || areParentsCalculated(hasCalculated, parents) || head.isProbabilistic()) {
+                Vertex<?> top = stack.pop();
+                top.updateValue();
+                hasCalculated.add(top);
+            } else {
+                parents.forEach(stack::push);
```
gordoncaleb: This can doubly add parents. Also, can we use a good ol for loop instead of the lambda? This line will be hit often enough that the lambda efficiency might be an issue.

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
```diff
@@ -98,7 +98,35 @@ public double logDensityAtValue() {
...
+        Deque<Vertex<?>> stack = new ArrayDeque<>();
+        stack.push(this);
+        Set<Vertex<?>> hasCalculated = new HashSet<>();
+
+        while (!stack.isEmpty()) {
+
+            Vertex<?> head = stack.peek();
+            Set<Vertex<?>> parents = head.getParents();
+
+            if (parents.isEmpty() || areParentsCalculated(hasCalculated, parents) || head.isProbabilistic()) {
```
gordoncaleb: Just realized that this iterates through the parents twice. Once in areParentsCalculated and then in the else branch. Ideally, the areParentsCalculated method would return the parents that aren't calculated instead of just a boolean that says there are un-calculated parents.

GeorgeNash: Nice call! I'll put that in

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
```diff
@@ -232,4 +258,16 @@ public int hashCode() {
     public Set<Vertex<?>> getConnectedGraph() {
         return DiscoverGraph.getEntireGraph(this);
     }
+
+    private Set<Vertex<?>> parentsThatAreNotCalculated(Set<Vertex<?>> calculated, Set<Vertex<?>> parents) {
+        Set<Vertex<?>> notCalculatedParents = new HashSet<>();
+        Iterator iterator = parents.iterator();
+        while (iterator.hasNext()) {
```
gordoncaleb: Why not a for loop?
```
for(Vertex<?> next: parents){
   if (!calculated.contains(next)) {
      notCalculatedParents.add(next);
   }
}
```

GeorgeNash: good spot, changed it

PR #50, opened 2018-04-30
Remove non log based density methods

gordoncaleb:
This PR removes density and dDensity from the base class Vertex. This makes logDensity and dLogDensity the preferred and required abstract methods. This was done because there's very little use cases where you would want density instead of the more numerically stable log density. Density and dDensity can be converted from the log versions if a user needs to but there's no use of it in our algorithms.

Also, the term logDensity (or density) has been replaced with logProb where appropriate. There's a disclaimer comment on the logProb that points out that in the continuous case this is actually logDensity but is accurate up to a constant.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/RejectionSampler.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/RejectionSampler.java
@@ -55,7 +55,7 @@ public class RejectionSampler {
 
     private static boolean matchesObservation(List<Vertex<?>> observedVertices) {
         return observedVertices.stream()
-                .allMatch(v -> v.densityAtValue() != 0.0);
+                .allMatch(v -> v.logDensityAtValue() != Double.NEGATIVE_INFINITY);
     }
 
     private static void takeSamples(Map<String, List<?>> samples, List<? extends Vertex<?>> fromVertices) {
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/RejectionSampler.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/RejectionSampler.java
@@ -55,7 +55,7 @@ public class RejectionSampler {
 
     private static boolean matchesObservation(List<Vertex<?>> observedVertices) {
         return observedVertices.stream()
-                .allMatch(v -> v.logDensityAtValue() != Double.NEGATIVE_INFINITY);
+                .allMatch(v -> v.logProbAtValue() != Double.NEGATIVE_INFINITY);
     }
 
     private static void takeSamples(Map<String, List<?>> samples, List<? extends Vertex<?>> fromVertices) {
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
@@ -18,62 +18,26 @@ public abstract class Vertex<T> implements Identifiable {
     private boolean observed;
 
     /**
-     * This is the value of the probability density at the supplied value.
-     *
-     * @param value The supplied value.
-     * @return The probability.
-     */
-    public abstract double density(T value);
-
-    /**
-     * Just a helper method for a common function
-     */
-    public double densityAtValue() {
-        return density(getValue());
-    }
-
-    /**
      * This is the value of the natural log of the probability density at the supplied value.
      *
      * @param value The supplied value.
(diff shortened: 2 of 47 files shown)


keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/RejectionSampler.java
```diff
@@ -55,7 +55,7 @@ private static void sampleLatents(List<? extends Vertex<?>> latents) {
 
     private static boolean matchesObservation(List<Vertex<?>> observedVertices) {
         return observedVertices.stream()
-                .allMatch(v -> v.densityAtValue() != 0.0);
+                .allMatch(v -> v.logProbAtValue() != Double.NEGATIVE_INFINITY);
```
GeorgeNash: Interesting!

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
```diff
@@ -18,62 +18,29 @@
     private boolean observed;
 
     /**
-     * This is the value of the probability density at the supplied value.
+     * This is the value of the natural log probability at the supplied value. In the
```
GeorgeNash: is it natural log prob or natural log of the prob?

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/probabilistic/Flip.java
```diff
@@ -8,41 +8,42 @@
 
 public class Flip extends ProbabilisticBool {
 
-    private final Vertex<Double> probTrue;
+    private final Vertex<Double> probabilityTrue;
```
GeorgeNash: Feels a bit odd to have `logProb` and now `probabilityTrue`. Let's pick `prob` or `probability` throughout?

keanu-project/src/main/java/io/improbable/keanu/vertices/intgr/probabilistic/FuzzyCastToIntegerVertex.java
```diff
@@ -184,10 +175,6 @@ private double dSdSigma(double x, double sigma) {
     }
 
     private double N(double mu, double sigma) {
-        return n(mu, sigma) / Math.sqrt(2 * Math.PI * sigma * sigma);
-    }
-
-    private double n(double mu, double sigma) {
-        return Math.exp(-(mu * mu) / (2.0 * sigma * sigma));
+        return Gaussian.pdf(mu, sigma, 0);
```
GeorgeNash: Nice tidy up

keanu-project/src/test/java/io/improbable/keanu/vertices/intgr/probabilistic/FuzzyCastToIntegerVertexTest.java
```diff
@@ -153,16 +153,16 @@ public void calculate_dP_dmu() {
...
         mu.setValue(mu2);
-        double density2 = fuzzy.densityAtValue();
+        double logDensity2 = fuzzy.logProbAtValue();
 
-        double expected_dPdmu = (density2 - density1) / delta;
+        double expected_dPdmu = (logDensity2 - logDensity1) / delta;
 
         log.info("Expected = " + expected_dPdmu + ", Actual = " + actual_dPdmu);
-        assertEquals(expected_dPdmu, actual_dPdmu, 1e-5);
+        assertEquals(expected_dPdmu, actual_dPdmu, 1e-4);
```
GeorgeNash: Was the test failing at e-5?

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
gordoncaleb: Changed to `This is the natural log of the probability at the supplied value.`

keanu-project/src/test/java/io/improbable/keanu/vertices/intgr/probabilistic/FuzzyCastToIntegerVertexTest.java
gordoncaleb: Yeah, the error from using the logProb is slightly higher. I think this is just due to the logProb being a bit steeper at this point than the density was. It can also be fixed by decreasing the size of delta.

GeorgeNash: Nice, just confirming

PR #54, opened 2018-05-03
Respect observed vertices as constants

gordoncaleb:
This PR fixes a few very inefficient bits of Keanu.

1. We calculate and return the gradient of an observed vertex with respect to itself. Instead, since the value of the vertex is fixed, the gradient should be zero.

2. For dual number calculation, an observed vertex should behave as a constant since it has no derivative with respect to anything.

3. In the case where we are doing max likelihood, it's possible that the likelihood function has no dependency on a particular latent vertex. In this case we were getting an NPE but it should just recognize that the gradient with respect to that latent is zero.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
@@ -23,7 +23,7 @@ public abstract class DoubleVertex extends ContinuousVertex<Double> implements D
      */
     protected abstract DualNumber calculateDualNumber(Map<Vertex, DualNumber> dualNumbers);
 
-    public DualNumber getDualNumber() {
+    public final DualNumber getDualNumber() {
         Map<Vertex, DualNumber> dualNumbers = new HashMap<>();
         Deque<DoubleVertex> stack = new ArrayDeque<>();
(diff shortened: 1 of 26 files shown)


keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
```diff
@@ -23,7 +23,7 @@
      */
     protected abstract DualNumber calculateDualNumber(Map<Vertex, DualNumber> dualNumbers);
 
-    public DualNumber getDualNumber() {
+    public final DualNumber getDualNumber() {
```
GeorgeNash: Nice. Is this to stop people implementing it by mistake instead of `calculateDualNumber`?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/ChiSquaredVertex.java
```diff
@@ -15,7 +15,6 @@
     public ChiSquaredVertex(IntegerVertex k, Random random) {
         this.k = k;
         this.random = random;
-        setValue(sample());
```
GeorgeNash: Good catch!

PR #68, opened 2018-05-15
Remove the need for random in vertex constructors

gordoncaleb:
Specifying a random in a vertex is only useful if you are using sample on that vertex. For MAP this isn't the case and as a result you end up specifying a random number generator and never actually using it. This PR moves the requirement of a RNG to the sample method and out of the vertex it's self. This causes a small problem when calling getValue on a probabilistic vertex that has no value yet. In this case, we can assume the caller is fine with using a globally defined RNG that lives on the Vertex base class. In the future, we may want to move this static RNG to a more appropriate place.


gordoncaleb:
Good question! I had it that way originally but then realized that meant we had a sample() method on every vertex that was potentially nondeterministic. If the user has set the seed manually on the default RNG then this might be what they want but they're more likely to use it and not know that it will be nondeterministic. Calling it sampleUsingDefaultRandom() makes it more likely the user gets what they expect.

PR #69, opened 2018-05-16
HMC with tensors

gordoncaleb:
Another fairly huge PR that adds a version of HMC that supports tensors.


keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
```diff
@@ -113,7 +113,11 @@ public void setValue(T value) {
     }
 
     public T getValue() {
-        return !hasValue() ? lazyEval() : value;
+        return hasValue() ? value : lazyEval();
```
GeorgeNash: nice

keanu-project/src/main/java/io/improbable/keanu/vertices/dbltensor/SimpleDoubleTensor.java
```diff
@@ -0,0 +1,319 @@
...
+        this.value = value;
+        this.shape = SCALAR_SHAPE;
+    }
+
+    public SimpleDoubleTensor(int[] shape) {
+        this.value = null;
+        this.shape = shape;
+    }
+
+    private SimpleDoubleTensor duplicate() {
```
GeorgeNash: Nice, let's get this on `Nd4jDoubleTensor` too. Add it to `DoubleTensor`?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbltensor/TensorShapeValidation.java
```diff
@@ -0,0 +1,108 @@
...
+     * the same shape of the proposal in question OR scalar.
+     *
+     * @param proposalShape the tensor shape being validated
+     * @param tensors       the tensors being validated against
+     * @throws IllegalArgumentException if there is more than one non-scalar shape OR if the non-scalar shape does
+     *                                  not match the proposal shape.
+     */
+    public static void checkTensorsMatchNonScalarShapeOrAreScalar(int[] proposalShape, Tensor... tensors) {
+
+        Set<TensorShape> nonScalarShapes = getNonScalarShapes(tensors);
```
GeorgeNash: `TensorShape` smashing it!

keanu-project/src/main/java/io/improbable/keanu/vertices/dbltensor/TensorShapeValidation.java
```diff
@@ -0,0 +1,108 @@
...
+
+        if (!nonScalarShapes.isEmpty()) {
+
+            boolean moreThanOneNonScalarShape = nonScalarShapes.size() > 1;
```
GeorgeNash: the variable assignment here seems a bit over the top. I think it's fine to have:

```
if (nonScalarShapes.size() > 1) {
            throw new IllegalArgumentException("More than a single non-scalar shape");
}
```

keanu-project/src/main/java/io/improbable/keanu/vertices/dbltensor/SimpleDoubleTensor.java
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    @Override
+    public DoubleTensor unaryMinusInPlace() {
+        value = -value;
+        return this;
+    }
+
+    @Override
+    public double[] getLinearView() {
```
GeorgeNash: I find `linearView` quite a strange way of expressing this flat concept. Is that what nd4j call it?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbltensor/SimpleDoubleTensor.java
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    @Override
+    public double scalar() {
+        return value;
+    }
+
+    @Override
+    public DoubleTensor reciprocal() {
+        return this.duplicate().reciprocalInPlace();
```
GeorgeNash: Why have we changed this to `duplicate` and then `op`? Is it faster?

If so, should we apply the `duplicate` then `op` logic to  `Nd4jDoubleTensor` operations to?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbltensor/TensorShapeValidation.java
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * This is a common function to check that tensors are either
+     * the same shape of the proposal in question OR scalar.
+     *
+     * @param proposalShape the tensor shape being validated
+     * @param tensors       the tensors being validated against
+     * @throws IllegalArgumentException if there is more than one non-scalar shape OR if the non-scalar shape does
+     *                                  not match the proposal shape.
+     */
+    public static void checkTensorsMatchNonScalarShapeOrAreScalar(int[] proposalShape, Tensor... tensors) {
```
GeorgeNash: I think you're fine to call this `checkTensorsMatchOrAreScalar`, it will read a bit nicer too

keanu-project/src/main/java/io/improbable/keanu/vertices/dbltensor/probabilistic/TensorUniformVertex.java
```diff
@@ -84,15 +85,15 @@ public double logPdf(DoubleTensor value) {
     public Map<Long, DoubleTensor> dLogPdf(DoubleTensor value) {
 
         DoubleTensor dlogPdf = DoubleTensor.zeros(this.xMax.getValue().getShape());
-        dlogPdf.applyWhere(value.getGreaterThanMask(xMax.getValue()), Double.NEGATIVE_INFINITY);
-        dlogPdf.applyWhere(value.getLessThanOrEqualToMask(xMin.getValue()), Double.POSITIVE_INFINITY);
+        dlogPdf.applyWhereInPlace(value.getGreaterThanMask(xMax.getValue()), Double.NEGATIVE_INFINITY);
```
GeorgeNash: Nice work with `applyWhereInPlace`. I can see a use case in which we'd want to have `applyWhere`. Add that in too?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbltensor/SimpleDoubleTensor.java
gordoncaleb: It's not necessarily faster. It's probably about the same, just less code. Good idea about doing the same in `Nd4jDoubleTensor`. I've updated the applicable methods there.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbltensor/SimpleDoubleTensor.java
gordoncaleb: Linear view is a 1d view of the underlying buffer in Nd4j, which is not what this actually is here. I've changed it to get getFlattenedView() and it returns an actual view (meaning editing it will actually edit the underlying tensor).

keanu-project/src/main/java/io/improbable/keanu/vertices/dbltensor/probabilistic/TensorUniformVertex.java
gordoncaleb: Good call. I've added a non-in-place version and changed the name to setWithMask in order to avoid confusion with the apply method you're adding in your Gamma PR.

PR #73, opened 2018-05-21
Use bayesian network interface for algorithms

gordoncaleb:
This PR turns the BayesNet class into a BayesianNetwork class that is continuous data type agnostic. This allows us to define a different BayesianNetwork extension for the tensor based algorithms and non-tensor algorithms. The algorithms that don't rely on continuous data types (e.g. Metropolis Hastings) will be able to use a BayesianNetwork that may or may not contain tensor vertices.


keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastingsTest.java
```diff
@@ -135,6 +176,8 @@ public void samplesFromPriorWithObservedDeterministic() {
 
         double postProbTrue = posteriorSamples.get(A).probability(v -> v);
 
+        System.out.println(postProbTrue);
```
GeorgeNash: Left over from debugging?

PR #86, opened 2018-06-01
Convert coal mining example to use tensor vertices

gordoncaleb:
Just some cleanup on the coal mining example. This converts it to use tensor vertices, which is a lot easier to read.

### DIFF
--- a/keanu-examples/coalMiningDisasters/src/test/java/com/example/coal/ModelTest.java
+++ b/keanu-examples/coalMiningDisasters/src/test/java/com/example/coal/ModelTest.java
@@ -16,7 +16,7 @@ public class ModelTest {
         Model model = new Model(coalMiningDisasterData);
         model.run();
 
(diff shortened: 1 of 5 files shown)


keanu-examples/coalMiningDisasters/src/test/java/com/example/coal/ModelTest.java
```diff
@@ -16,7 +16,7 @@ public void shouldFindCorrectSwitchYear() {
         Model model = new Model(coalMiningDisasterData);
         model.run();
 
-        int switchYear = model.results.get(model.switchpoint).getMode();
+        int switchYear = model.results.getIntegerTensorSamples(model.switchpoint).getIntegerMode();
```
joncoello83: Does 'getIntegerMode' add anything over 'getMode'? Seems very clear already that you're working with integers thanks to 'getIntegerTensorSamples'.

PR #96, opened 2018-06-19
Fix uniform int range and sampling bug as well as flip sample shape bug

gordoncaleb:
Fixes some minor bugs discovered recently.


keanu-project/src/main/java/io/improbable/keanu/vertices/intgr/probabilistic/UniformIntVertex.java
```diff
@@ -95,8 +95,8 @@ public double logPmf(IntegerTensor value) {
     @Override
     public IntegerTensor sample(KeanuRandom random) {
 
-        DoubleTensor delta = max.getValue().minus(min.getValue()).toDouble();
-        DoubleTensor randoms = random.nextDouble(delta.getShape());
+        DoubleTensor delta = max.getValue().toDouble().minus(min.getValue().toDouble());
```
GeorgeNash: Is this different from the previous?

PR #104, opened 2018-06-27
Add testing for numerical equals vertex

gordoncaleb:
This PR cleans up some tech debt. Some basic tests are added and a bug fix with DoubleTensor.absInPlace().


keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -504,7 +504,8 @@ public DoubleTensor unaryMinusInPlace() {
 
     @Override
     public DoubleTensor absInPlace() {
-        return duplicate().absInPlace();
```
GeorgeNash: Eesh, good spot

PR #105, opened 2018-06-28
Create eval and lazyEval function

gordoncaleb:
This PR renames the previously known method (Vertex).lazyEval() to (Vertex).eval(). This eval() method propagates all the way up the graph to a probabilistic vertex and evaluates on the way down. The new lazyEval() method propagates up to the point where the graph has values and then evaluates on the way down. This is being added in order to provide extra flexibility for efficient graph evaluation (specifically around computationally intense blackbox vertices).


keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/VertexValuePropagation.java
```diff
@@ -125,4 +125,89 @@ private static void cascadeUpdate(Deque<Vertex> stack, Map<Long, Long> explored)
 
         }
     }
+
+    public static void eval(Vertex... vertices) {
+        eval(Arrays.asList(vertices));
+    }
+
+    public static void eval(Collection<? extends Vertex> vertices) {
```
GeorgeNash: `eval` and `lazyEval` are so similar that I think it will actually be simpler to understand if we put them under one function and use a conditional for the `parentsThatAreNotCalculated` like:

```
 public static void eval(Collection<? extends Vertex> vertices, boolean checkValueOrVisited) {
        Deque<Vertex<?>> stack = new ArrayDeque<>();
        for (Vertex v : vertices) {
            stack.push(v);
        }

        Set<Vertex<?>> hasCalculated = new HashSet<>();

        while (!stack.isEmpty()) {

            Vertex<?> head = stack.peek();
            Set<Vertex<?>> parentsThatAreNotYetCalculated = checkValueOrVisited ?
                parentsThatAreNotCalculated(hasCalculated, head.getParents()) : parentsThatAreNotCalculated(head.getParents());

            if (head.isProbabilistic() || parentsThatAreNotYetCalculated.isEmpty()) {

                Vertex<?> top = stack.pop();
                top.updateValue();
                hasCalculated.add(top);

            } else {

                for (Vertex<?> vertex : parentsThatAreNotYetCalculated) {
                    stack.push(vertex);
                }

            }

        }
    }
```

PR #109, opened 2018-07-04
Remove the need for exploring graph when setting and cascading

gordoncaleb:
This PR removes the need to do a graph explore before propagating updates. This removes quite a bit of code where we were caching explore results or methods that allows the use of cached explore results. This explore method may come back one day but for now using vertex ids for topographical sorting is the clearest solution.


keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/VertexValuePropagation.java
```diff
@@ -5,118 +5,45 @@
...
-                if (child.isProbabilistic()) {
-                    continue;
-                }
-
-                if (!hasVisited.contains(child)) {
-                    stack.push(child);
-                } else {
-                    crossRoadCount.put(child.getId(), crossRoadCount.getOrDefault(child.getId(), 0L) + 1);
-                }
+        HashSet<Vertex> alreadyQueued = new HashSet<>(cascadeFrom);
```
GeorgeNash: Do we need this? Can't we use the `.contains` on the `PriorityQueue`?

PR #114, opened 2018-07-06
Add binomial distribution vertex

gordoncaleb:
This PR adds the binomial distribution vertex. Sampling and log probability is done element wise and not in the GPU. This will remain the case until there's a performance issue (if there ever will be).


migwellian:
There are several implementations of binomial that avoid the integer overflow you get from the factorial - see https://math.stackexchange.com/questions/202554/how-do-i-compute-binomial-coefficients-efficiently

migwellian:
Calculating binomial coefficients on the GPU with CUDA: https://research.ijcaonline.org/volume71/number2/pxc3888570.pdf

PR #116, opened 2018-07-10
Correct number of interpolation points for Non Gradient Optimizer

rory-improbable:
Previously the number of interpolation points for the BOBYQA algorithm was calculated using the number of latent vertices, which didn't take into account tensors.

Now the number of interpolation points is calculated based on the total number of latent dimensions, taking tensor vertices into account.

ALSO: refactored some optimizer functions into a shared abstract class

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/NonGradientOptimizer.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/NonGradientOptimizer.java
@@ -14,10 +14,9 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.function.BiConsumer;
 
-import static io.improbable.keanu.algorithms.variational.GradientOptimizer.currentPoint;
 import static org.apache.commons.math3.optim.nonlinear.scalar.GoalType.MAXIMIZE;
 
-public class NonGradientOptimizer {
+public class NonGradientOptimizer extends Optimizer {
 
     private final BayesianNetwork bayesNet;
     private final List<BiConsumer<double[], Double>> onFitnessCalculations;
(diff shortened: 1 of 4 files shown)


keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/Optimizer.java
```diff
@@ -0,0 +1,42 @@
...
+import io.improbable.keanu.vertices.Vertex;
+
+import java.util.Arrays;
+import java.util.List;
+
+public abstract class Optimizer {
+
+    private static final double FLAT_GRADIENT = 1e-16;
+
+    static double[] currentPoint(List<Vertex<DoubleTensor>> continuousLatentVertices) {
```
gordoncaleb: These should be `protected` and not `static` since it's an abstract class.

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/Optimizer.java
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        return point;
+    }
+
+    static long totalNumLatentDimensions(List<? extends Vertex<DoubleTensor>> continuousLatentVertices) {
+        return continuousLatentVertices.stream().mapToLong(FitnessFunction::numDimensions).sum();
+    }
+
+    static void warnIfGradientIsFlat(double[] gradient) {
```
gordoncaleb: I don't think this method should be part of the Optimizer interface since it's gradient specific. Can we just move it to the GradientOptimizer?

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/NonGradientOptimizer.java
```diff
@@ -84,6 +83,10 @@ public double optimize(int maxEvaluations, double boundsRange, List<Vertex> outp
         return pointValuePair.getValue();
     }
 
+    private int getNumInterpolationPoints(List<? extends Vertex<DoubleTensor>> latentVertices){
+        return (int)(2 * totalNumLatentDimensions(latentVertices) + 1);
```
gordoncaleb: Nice reuse of that `totalNumLatentDimensions(...)`.

PR #117, opened 2018-07-11
Feature/not

migwellian:


keanu-project/src/main/java/io/improbable/keanu/vertices/bool/BoolVertex.java
```diff
@@ -17,19 +18,23 @@
...
-    public final io.improbable.keanu.vertices.bool.BoolVertex and(Vertex<BooleanTensor>... those) {
+    public final BoolVertex and(Vertex<BooleanTensor>... those) {
         if (those.length == 0) return this;
         if (those.length == 1) return new AndBinaryVertex(this, those[0]);
         return new AndMultipleVertex(inputList(those));
     }
 
+    public static final BoolVertex not(Vertex<BooleanTensor> vertex) {
+        return new NotVertex(vertex);
+    }
```
GeorgeNash: If you fancy the challenge, want to extend this to work for multiple vertices? So we can do something like...

```
public static final BoolVertex not(Vertex<BooleanTensor>... those) {
   if (those.length == 0) // do nothing
   if (those.length == 1) return new NotVertex(those[0])
   return new NotMultipleVertex(inputList(those))
}
```

gordoncaleb: The not multiple would have multiple outputs?

keanu-project/src/test/java/io/improbable/keanu/vertices/bool/BoolVertexTest.java
```diff
@@ -50,7 +52,58 @@ public void doesAnd() {
         v1.setValue(true);
         v2.setValue(false);
 
-        assertTrue(!v3.eval().scalar());
+        assertFalse(v3.eval().scalar());
```
gordoncaleb: Nice testing!

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/BoolVertex.java
GeorgeNash: Eesh, good point, you can't collapse a not can you. Never mind!

PR #118, opened 2018-07-11
Linear regression with Matrix Multiply

gordoncaleb:
When trying to implement linear regression with the new matrix multiply vertex, I came across two bugs:

1. The summing done in the logProb for vertices is not being reflected in the dLogProb result. The bottom dimensions need to be summed over in order to match the behaviour of logProb.
2. The broadcast multiply in ND4J does not match what we expect or what is returned from numpy. The times operation on DoubleTensor now catches would be broadcast multiplies and does them correctly instead of letting ND4j take it. I'll raise this with the ND4J team as well.

Also, while trying to debug the problem I came up against some incorrect variable names in the distribution derivative code. I've refactored these to be sensible and consistent.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
@@ -1,5 +1,7 @@
 package io.improbable.keanu.tensor;
 
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+
 import java.util.Arrays;
 
 public class TensorShape {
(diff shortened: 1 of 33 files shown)


keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
```diff
@@ -1,5 +1,7 @@
 package io.improbable.keanu.tensor;
 
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
```
GeorgeNash: Unused import

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/PartialDerivatives.java
```diff
@@ -264,4 +236,28 @@ public PartialDerivatives clone() {
         }
         return clone;
     }
+
+    private static DoubleTensor increaseRankByAppendingOnesToShape(DoubleTensor lowRankTensor, int desiredRank) {
+        return increaseRankByPaddingOnes(lowRankTensor, desiredRank, true);
+    }
+
+    private static DoubleTensor increaseRankByPrependingOnesToShape(DoubleTensor lowRankTensor, int desiredRank) {
```
GeorgeNash: Should we move these to `TensorShape` as public static helpers?

Then we can change:

```
        if (!this.isObserved()) {
            dLogPdInputs.putWithRespectTo(getId(), dLogPdx.reshape(concat(SCALAR_SHAPE, dLogPdx.getShape())));
        }
```

to:

```
        if (!this.isObserved()) {
            dLogPdInputs.putWithRespectTo(getId(), increaseRankByPrependingOnesToShape(dLogPdx, dLogPdx.getRank() + 2));
        }
```

keanu-project/src/test/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensorTest.java
```diff
@@ -138,4 +134,59 @@ public void canApplySqrt() {
...
+        DoubleTensor rank4 = DoubleTensor.create(new double[]{1, 2, 3, 4, 5, 6, 7, 8}, new int[]{2, 2, 2, 1});
+        DoubleTensor matrix = matrixA.reshape(2, 2, 1, 1);
+        DoubleTensor expected = DoubleTensor.create(new double[]{1, 2, 6, 8, 15, 18, 28, 32}, new int[]{2, 2, 2, 1});
+
+        assertTimesOperationEquals(rank4, matrix, expected);
+        assertTimesInPlaceOperationEquals(rank4, matrix, expected);
+    }
+
+    @Test
+    public void canBroadcastMultiplyRank4ContainingMatrixAndMatrix() {
```
GeorgeNash: nice

migwellian:
We should have a chat about the `Diff` nested classes. I've nuked them as part of PR https://github.com/improbable-research/keanu/pull/122

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
migwellian: You can configure IntelliJ to remove these automatically: Editor - General - Auto Import

gordoncaleb: Wow I didn't know you could get IntelliJ to just magically optimize imports on the fly!

PR #121, opened 2018-07-11
Change behaviour to reshape on setValue and observe

gordoncaleb:
Previously, when you observed/setValue/setAndCascade a primitive or array then it would NOT reshape the vertex. This has been confusing to users (and me) and I propose we change the behaviour to be such that anything you set or observe is exactly reflected in the vertex. For example, setting or observing a scalar value would make the shape of the vertex scalar and if you observe an array then the vertex would would be reshaped as a vector.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/tensor/bool/SimpleBooleanTensor.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/tensor/bool/SimpleBooleanTensor.java
@@ -27,7 +27,8 @@ public class SimpleBooleanTensor implements BooleanTensor {
     private final int[] stride;
 
     public SimpleBooleanTensor(boolean[] data, int[] shape) {
-        this.data = data;
+        this.data = new boolean[(int) TensorShape.getLength(shape)];
+        System.arraycopy(data, 0, this.data, 0, this.data.length);
         this.shape = shape;
         this.stride = TensorShape.getRowFirstStride(shape);
     }
@@ -46,9 +47,9 @@ public class SimpleBooleanTensor implements BooleanTensor {
 
     public SimpleBooleanTensor(boolean constant, int[] shape) {
         this.data = new boolean[(int) TensorShape.getLength(shape)];
+        Arrays.fill(this.data, constant);
         this.shape = shape;
         this.stride = TensorShape.getRowFirstStride(shape);
-        Arrays.fill(this.data, constant);
     }
 
     @Override
--- a/keanu-project/src/main/java/io/improbable/keanu/tensor/bool/SimpleBooleanTensor.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/tensor/bool/SimpleBooleanTensor.java
@@ -26,6 +26,10 @@ public class SimpleBooleanTensor implements BooleanTensor {
     private final int[] shape;
     private final int[] stride;
 
+    /**
+     * @param data  tensor data used c ordering
+     * @param shape desired shape of tensor
+     */
     public SimpleBooleanTensor(boolean[] data, int[] shape) {
         this.data = new boolean[(int) TensorShape.getLength(shape)];
         System.arraycopy(data, 0, this.data, 0, this.data.length);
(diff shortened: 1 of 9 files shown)


keanu-project/src/test/java/io/improbable/keanu/vertices/intgr/IntegerVertexTest.java
```diff
@@ -66,4 +66,35 @@ public void doesLambda() {
...
+        int[] values = new int[]{1, 2, 3};
+        binomialVertex.setAndCascade(values);
+        assertArrayEquals(values, binomialVertex.getValue().asFlatIntegerArray());
+    }
+
+    @Test
+    public void canSetValueAsArrayOfValues() {
+        IntegerVertex binomialVertex = new BinomialVertex(0.5, 20);
+        int[] values = new int[]{1, 2, 3};
+        binomialVertex.setAndCascade(values);
```
migwellian: copy paste error? should be .set()?

keanu-project/src/test/java/io/improbable/keanu/vertices/intgr/IntegerVertexTest.java
```diff
@@ -66,4 +66,35 @@ public void doesLambda() {
...
+    }
+
+    @Test
+    public void canSetValueAsScalarOnNonScalarVertex() {
+        IntegerVertex binomialVertex = new BinomialVertex(new int[]{2, 1}, 0.5, 20);
+        binomialVertex.setValue(2);
+        assertArrayEquals(new int[]{2}, binomialVertex.getValue().asFlatIntegerArray());
+    }
+
 }
```
migwellian: add a `canSetAndCascadeAsScalarOnNonScalarVertex` method (as in the `DoubleVertexTest`)?

keanu-project/src/main/java/io/improbable/keanu/tensor/bool/SimpleBooleanTensor.java
```diff
@@ -46,9 +47,9 @@ public SimpleBooleanTensor(int[] shape) {
 
     public SimpleBooleanTensor(boolean constant, int[] shape) {
         this.data = new boolean[(int) TensorShape.getLength(shape)];
+        Arrays.fill(this.data, constant);
```
migwellian: What's the difference between `System.arraycopy` and `Arrays.fill`? Why are you using both in this class?

keanu-project/src/test/java/io/improbable/keanu/vertices/bool/BoolVertexTest.java
```diff
@@ -95,6 +94,36 @@ public void constantVertexWorksAsExpected() {
...
+        flip.setValue(values);
+        assertArrayEquals(new Boolean[]{true, false, true}, flip.getValue().asFlatArray());
+    }
+
+    @Test
+    public void canSetValueAsScalarOnNonScalarVertex() {
+        BoolVertex flip = new Flip(new int[]{2, 1}, 0.5);
+        flip.setValue(true);
+        assertArrayEquals(new Boolean[]{true}, flip.getValue().asFlatArray());
+    }
```
migwellian: add a canSetAndCascadeAsScalarOnNonScalarVertex method (as in the DoubleVertexTest)?

keanu-project/src/test/java/io/improbable/keanu/vertices/intgr/IntegerVertexTest.java
gordoncaleb: Good catch.

keanu-project/src/test/java/io/improbable/keanu/vertices/intgr/IntegerVertexTest.java
gordoncaleb: Yup. Good call.

keanu-project/src/main/java/io/improbable/keanu/tensor/bool/SimpleBooleanTensor.java
gordoncaleb: Fill is for filling a single value into a data buffer for an arbitrary shape e.g. a 2x2 of trues. Array copy is used to copy from the user provided boolean buffer to an internal buffer. This copy protects against the case where you pass in more data than the shape allows.

PR #127, opened 2018-07-16
Add accessors for divider/dividend on divide vertex

chris-major-improbable:


keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/binary/DivisionVertex.java
```diff
@@ -19,6 +22,16 @@
      */
     public DivisionVertex(DoubleVertex a, DoubleVertex b) {
         super(checkHasSingleNonScalarShapeOrAllScalar(a.getShape(), b.getShape()), a, b);
+        this.dividend = a;
```
gordoncaleb: I think we should rename `a` and `b` to `left` and `right` and then just return `left` in getDividend() and `right` in getDivisor(). Otherwise we have two references to the arguments `a`/`b` and `dividend`/`divisor`. This change would touch ALL of the DoubleBinaryOps.

chris-major-improbable:
This version is a bit more complete - Left and Right for all binary operators, and named getters for divide, power and possible others.

PR #129, opened 2018-07-18
Added access to internal representation of Nd4j tensors

chris-major-improbable:
This is to allow other code to do something like:
```java
if ( tensor instanceof Nd4jDoubleTensor){
  Nd4jDoubleTensor internal = (Nd4jDoubleTensor)internal;
  INDArray represntation = internal.getInternalRepresentation();
  // do something efficient
}else{
  // do something inefficient
}
```
Likely to be primarily in extensions such as new vertices that can be written to be more performant with direct access tot he Nd4j tensor represnetation.


gordoncaleb:
Allowing the user access to this object could violate some of the assumptions we make. We could do a defensive copy, which would avoid that but the real kicker is that by allowing the user to work around Keanu, we'll miss a lot of feature requests that would force us/user to do it properly and in a consistent way.

chris-major-improbable:
Opened a ticket (#137) detailing the thinking behind this PR

gordoncaleb:
Moved to github issue https://github.com/improbable-research/keanu/issues/137

PR #132, opened 2018-07-20
Feature/remove starter

christophernorth:
Remove the starter project from the main keanu repo.  Also done a little bit of cleanup in the documentation to reference the "latest" version of keanu.


keanu-docs/getting-started.md
```diff
@@ -61,7 +60,7 @@ in your gradle or maven build file.
 In your project's build.gradle:
 
 ```$groovy
-compile group: 'io.improbable', name: 'keanu', version: '0.0.8'
```
gordoncaleb: Good catch! This is supposed to be done as part of our release process but we seem to be falling down on that one. Ideally, we would automate this somehow. There might be a way to just link through to Maven Central to the latest version.

christophernorth: Yeah - I was thinking the same thing.  As discussed, I'll have a look in to the best ways to go about this over the weekend.

PR #134, opened 2018-07-20
Define interface and builders for optimizers

gordoncaleb:
This PR

- Defines an interface for Optimizers
- Adds Optimizer.of(BayesianNetwork net) that returns an appropriate optimizer (gradient or nongradient)
- Makes the BayesianNetwork less stateful in order to be useable across observation and unobservations


keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/Optimizer.java
```diff
@@ -26,7 +49,7 @@
         return point;
     }
 
-    protected long totalNumLatentDimensions(List<? extends Vertex<DoubleTensor>> continuousLatentVertices) {
+    static long totalNumberOfLatentDimensions(List<? extends Vertex> continuousLatentVertices) {
```
migwellian: `Vertex<?>` is preferable to `Vertex`

migwellian: Basically because `Vertex` is interpreted as `Vertex<Object>` which will lead to compiler errors elsewhere.

migwellian: also in https://github.com/improbable-research/keanu/pull/131/files

Issue #138, opened 2018-07-25
Extended function on DoubleTensor


chris-major-improbable:
**I'm trying to use `DoubleTesor` and feel it needs these functions:**
* `equalsWithEps`
This would allow the comparison between two `DoubleTensor`s with a specified epsilon.
* `zero`
This would set all elements in a `DoubleTensor` to 0. This may be a special case of a more general `setAll` function.

chris-major-improbable:
Sugested implementation is in a PR #139

GeorgeNash:
Closing as PR has been merged and released.

PR #139, opened 2018-07-25
Feature/double tensor equals with eps and zero issue 138

chris-major-improbable:
PR suggestion for issue #138

Adds implementation of the following to `DoubleTensor`
* `setAll`
* `equalsWithEps`


keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/ScalarDoubleTensor.java
```diff
@@ -637,6 +645,17 @@ public DoubleTensor standardizeInPlace() {
         throw new IllegalStateException("Cannot standardize a scalar");
     }
 
+    @Override
+    public DoubleTensor zero() {
```
migwellian: This could be the default method in the `DoubleTensor` interface

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
```diff
@@ -159,6 +159,8 @@ static DoubleTensor placeHolder(int[] shape) {
 
     double standardDeviation();
 
+    boolean equalsWithEps( DoubleTensor other , double epsilon );
```
migwellian: Can I suggest we call this `isCloseTo`? That aligns with the method name both in Hamcrest and Python.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
```diff
@@ -245,6 +247,10 @@ static DoubleTensor placeHolder(int[] shape) {
 
     DoubleTensor standardizeInPlace();
 
+    DoubleTensor zero();
```
migwellian: I think we should implement both `zero` and `zeroInPlace` so that we're calling out when the tensor will be modified.

migwellian: Also note that we already have a static `zeros` method

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
chris-major-improbable: I'd expect `isCloseTo` to be a distance mertic - where as this compares each dimension independently.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
chris-major-improbable: The static `zero` doesn't mutate - it creates.... I don't see why we'd create a non-static non-mutating zero?

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/ScalarDoubleTensor.java
chris-major-improbable: I'm not sure what you mean by this can you explain?

keanu-project/src/test/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensorTest.java
```diff
@@ -179,6 +180,36 @@ public void canBroadcastMultiplyRank5ContainingMatrixAndMatrix() {
...
+            6, 3, 2, 9, 3, 4, 7, 6, 6, 2, 5, 4, 0, 2, 1, 3
+        }, new int[]{2, 2, 2, 2, 2});
+        rank5.zero();
+        assertAllValuesAre(rank5,0.0);
+        rank5.setAll(0.8);
+        assertAllValuesAre(rank5,0.8);
+    }
+
+    @Test
+    public void canEqualsWithEspilon(){
```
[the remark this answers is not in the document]
chris-major-improbable: will fix

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/ScalarDoubleTensor.java
migwellian: Instead of having an identical `zero` method in both `ScalarDoubleTensor` and `Nd4jDoubleTensor`, you can put it in `DoubleTensor` and use the `default` keyword (introduced in Java 8).

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
migwellian: That's an interesting way of looking at it... if you consider the "Manhattan" distance metric then this is the definition of "close" in that metric!

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
migwellian: Agreed. To refine my request: can we call this method "zeroInPlace" so that users understand that it's mutating?

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
chris-major-improbable: Wouldn't Manhattan distance be the sum of differences - where this ensures that it's within a hypercube with sides of length `2*epsilon`...
Regardless the word `close` seems to imply a distance based metric - which isn't how the epsilon gets applied -and I'd prefer specific naming even if it diverges from other languages... However this is the syntax used in `junit` and `nd4j`

migwellian: The plot thickens: JUnit just has `assertEquals(expected, actual, delta)`. We probably wouldn't want to call our method `equals` though!

chris-major-improbable: it does - but the depreciation notice on `assertEquals(double,double)` calls it `epsilon` :)

I'm not to bothered, but wonder if it's worth getting other people's opinons

migwellian: I think your name is good, except I'd favour the slightly more long-hand `equalsWithinEpsilon`.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
```diff
@@ -245,6 +247,12 @@ static DoubleTensor placeHolder(int[] shape) {
 
     DoubleTensor standardizeInPlace();
 
+    default public DoubleTensor zero() {
```
gordoncaleb: This should be setAllToZero(). Or can we just use `setAll(0.0)`, which is both shorter and more clear?

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/ScalarDoubleTensor.java
```diff
@@ -353,6 +353,14 @@ public double standardDeviation() {
         throw new IllegalStateException("Cannot find the standard deviation of a scalar");
     }
 
+    @Override
+    public boolean equalsWithEps(DoubleTensor o, double epsilon) {
+        if (this == o) return true;
+        if (!this.hasSameShapeAs(o)) return false;
+        double difference = value - o.asFlatList().get(0);
```
gordoncaleb: Since o has the same shape, it's ok to just `value - o.scalar()`. The asFlatList() will do a lot more than you need.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -171,6 +171,24 @@ public double standardDeviation() {
...
+    @Override
+    public boolean equalsWithEps(DoubleTensor o, double epsilon) {
+        if (this == o) return true;
+
+        if (o instanceof Nd4jDoubleTensor) {
+            return tensor.equalsWithEps(((Nd4jDoubleTensor) o).tensor,epsilon);
+        } else {
+            if (this.hasSameShapeAs(o)) {
+                DoubleTensor difference = o.minus(this);
+                for (double v : difference.asFlatList()) {
```
gordoncaleb: `difference.getGreaterThanMask(DoubleTensor.scalar(epsilon)).sum() >= 1` or `difference.lessThan(epsilon).allTrue()`. Not sure which is more performant but either would be preferable here.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
gordoncaleb: I like the longer form name `equalsWithinEpsilon` but it looks like Numpy uses `isclose()`, which creates a stronger argument for `isCloseTo(...)`. I think `equalsWithinEpsilon(..)` is more like what we're doing, which is an element-wise compare as opposed to a multidimensional distance calculation.

keanu-project/src/test/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensorTest.java
```diff
@@ -179,6 +180,36 @@ public void canBroadcastMultiplyRank5ContainingMatrixAndMatrix() {
...
+            6, 3, 2, 9, 3, 4, 7, 6, 6, 2, 5, 4, 0, 2, 1, 3
+        }, new int[]{2, 2, 2, 2, 2});
+        rank5.zero();
+        assertAllValuesAre(rank5,0.0);
+        rank5.setAll(0.8);
+        assertAllValuesAre(rank5,0.8);
+    }
+
+    @Test
+    public void canEqualsWithEpsilon(){
```
gordoncaleb: Nice test!

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -171,6 +171,21 @@ public double standardDeviation() {
...
 
+    @Override
+    public boolean equalsWithEps(DoubleTensor o, double epsilon) {
+        if (this == o) return true;
+
+        if (o instanceof Nd4jDoubleTensor) {
+            return tensor.equalsWithEps(((Nd4jDoubleTensor) o).tensor,epsilon);
+        } else {
+            if (this.hasSameShapeAs(o)) {
+                DoubleTensor difference = o.minus(this);
```
gordoncaleb: Isn't this abs(diff) < eps?

PR #141, opened 2018-07-26
Implementation of Kernel Density Estimation with a Gaussian Kernel

johannespetrat:
Approximates the distribution of sum samples X=(x_0, ..., x_n) using a sum of Gaussians.

<a href="https://www.codecogs.com/eqnedit.php?latex=f_{KDE}(x)&space;=&space;\frac{1}{nh}&space;\sum_{i=0}^nK(\frac{x-x_i}{h}),&space;\quad&space;\text{&space;where&space;}&space;K(x)&space;=&space;\frac{1}{\sqrt(2\pi)}e^{-\frac{1}{2}x^2}" target="_blank"></a>

Relates to SOL-1980


migwellian:
Loving your use of a codecogs formula in the comments!

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/GaussianKDE.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class GaussianKDE {
+
+    public KDEVertex approximate(VertexSamples<DoubleTensor> vertexSamples){
+
+        List<Double> samples = vertexSamples.asList().stream()
+            .map(tensor -> tensor.scalar())
```
migwellian: Does this mean it only works on scalars?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/KDEVertex.java
```diff
@@ -0,0 +1,135 @@
...
+    private DoubleTensor samples;
+
+    public KDEVertex(DoubleTensor samples, double bandwidth){
+        this.samples = samples;
+        this.bandwidth = bandwidth;
+    }
+
+    public KDEVertex(DoubleTensor samples) {
+        this.samples = samples;
+        this.bandwidth = scottsBandwith();
```
migwellian: `this(samples, scottsBandwith())` is preferable, to avoid code duplication and potential for bugs if the code changes later

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/KDEVertex.java
```diff
@@ -0,0 +1,135 @@
...
+    private double getVariance() {
+        double mean = getMean();
+        return samples.minus(DoubleTensor.create(mean, samples.getShape())).powInPlace(2).sum() / samples.getLength();
+    }
+
+    private double getStandardDeviation() {
+        return Math.sqrt(getVariance());
+    }
+
+    private double scottsBandwith() {
```
migwellian: typo: scottsBandwidth

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/KDEVertex.java
```diff
@@ -0,0 +1,135 @@
...
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class KDEVertex extends ProbabilisticDouble {
+    public double bandwidth;
+    private DoubleTensor samples;
+
+    public KDEVertex(DoubleTensor samples, double bandwidth){
```
migwellian: Does this vertex have no parents?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/KDEVertex.java
```diff
@@ -0,0 +1,135 @@
...
+
+    private DoubleTensor gaussianKernel(DoubleTensor x) {
+        DoubleTensor exponent = x.pow(2.).times(-1. / 2.);
+        DoubleTensor power = DoubleTensor.create(Math.E, exponent.getShape()).powInPlace(exponent);
+        power.timesInPlace(1. / Math.sqrt(2. * Math.PI));
+        return power;
+    }
+
+    private double getMean() {
+        return samples.sum() / samples.getLength();
```
migwellian: Do we need to catch divide-by-zero error?

migwellian: also in `getVariance`

johannespetrat: good point, I'll instead throw an exception if an empty list of samples is provided to the constructor

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/KDEVertex.java
johannespetrat: It doesn't. This vertex should be used to approximate, for example, a posterior distributions based on samples drawn from it.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/KDEVertex.java
johannespetrat: 👍

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/GaussianKDE.java
johannespetrat: Yes, it only works with samples drawn from a univariate distribution.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/KDEVertex.java
johannespetrat: good point, I'll have to make scottsBandwidth() a static method

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/GaussianKDE.java
migwellian: In that case, you should probably throw an exception if the tensor is not scalar. The behaviour for a non-scalar is that it will just return the first value and throw away the rest.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/KDEVertex.java
migwellian: I see - so is it intended to be used as the root of a graph, i.e. it will have children?

johannespetrat: Yep, it may be used as an input to other vertices.

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/GaussianKDE.java
johannespetrat: Ok, i'll throw an `IllegalArgumentException` here. I also added a test for it.

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/GaussianKDE.java
```diff
@@ -0,0 +1,40 @@
...
+    public KDEVertex approximate(DoubleVertex vertex, Integer nSamples){
+        BayesianNetwork network = new BayesianNetwork(vertex.getConnectedGraph());
+        DoubleVertexSamples vertexSamples = MetropolisHastings.withDefaultConfig()
+            .getPosteriorSamples(network, Arrays.asList(vertex), nSamples).getDoubleTensorSamples(vertex);
+        return approximate(vertexSamples);
+    }
+
+    private DoubleTensor checkIfScalar(DoubleTensor tensor) throws IllegalArgumentException {
+        if (tensor.asFlatList().size()>1){
+            throw new IllegalArgumentException();
```
christophernorth: Just a minor comment - could we put a reason string in as part of constructing the Exception?  Hopefully will help with Debugging if people hit this case?

johannespetrat: good point, I was being lazy there.

PR #145, opened 2018-07-30
Allow specific bounds for each vertex to be provided to BOBYQA

rory-improbable:
Rather than just providing one bounds number for all the latent vertices passed to the optimizer, this allows you to pass in a map of latent vertices to bounds and therefore gives greater flexibility.

The default behaviour is still to pass in boundsRange as normal. If you don't pass in `vertexSpecificBounds` then this will revert to default behaviour

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/BoundsCalculator.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/BoundsCalculator.java
@@ -1,0 +1,75 @@
+package io.improbable.keanu.algorithms.variational;
+
+import io.improbable.keanu.tensor.TensorShape;
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+import io.improbable.keanu.vertices.Vertex;
+import lombok.RequiredArgsConstructor;
+import org.apache.commons.math3.optim.SimpleBounds;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+@RequiredArgsConstructor
+public class BoundsCalculator {
+
+    private final double boundsRange;
+    private final OptimizerBounds optimizerBounds;
+
+    public SimpleBounds getBounds(List<? extends Vertex<DoubleTensor>> latentVertices, double[] startPoint) {
+        List<Double> minBounds = new ArrayList<>();
+        List<Double> maxBounds = new ArrayList<>();
+
+        for (Vertex<DoubleTensor> vertex : latentVertices) {
+
+            if (optimizerBounds.hasBound(vertex)) {
+                validateBoundsForVertex(vertex, optimizerBounds);
+                addBoundsForVertex(vertex, optimizerBounds, minBounds, maxBounds);
+            } else {
+                int length = (int) TensorShape.getLength(vertex.getShape());
+                int startIndex = minBounds.size();
+                for (int i = 0; i < length; i++) {
+                    minBounds.add(startPoint[i + startIndex] - boundsRange);
+                    maxBounds.add(startPoint[i + startIndex] + boundsRange);
+                }
+            }
+        }
+
+        return new SimpleBounds(
+            minBounds.stream().mapToDouble(d -> d).toArray(),
+            maxBounds.stream().mapToDouble(d -> d).toArray()
+        );
+    }
+
+    private void addBoundsForVertex(Vertex<DoubleTensor> vertex,
+                                    OptimizerBounds bounds,
+                                    List<Double> minBounds,
+                                    List<Double> maxBounds) {
+
+        DoubleTensor lowerBound = bounds.getLower(vertex);
+        DoubleTensor upperBound = bounds.getUpper(vertex);
+
+        if (lowerBound.isScalar()) {
+            minBounds.addAll(DoubleTensor.create(lowerBound.scalar(), vertex.getShape()).asFlatList());
+        } else {
+            minBounds.addAll(lowerBound.asFlatList());
+        }
+
+        if (upperBound.isScalar()) {
+            maxBounds.addAll(DoubleTensor.create(upperBound.scalar(), vertex.getShape()).asFlatList());
+        } else {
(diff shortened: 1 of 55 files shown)


keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/NonGradientOptimizer.java
```diff
@@ -12,8 +12,7 @@
 import org.apache.commons.math3.optim.nonlinear.scalar.ObjectiveFunction;
 import org.apache.commons.math3.optim.nonlinear.scalar.noderiv.BOBYQAOptimizer;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.*;
```
migwellian: FYI - I've started documenting some IntelliJ settings at https://improbableio.atlassian.net/wiki/spaces/~616259109/pages/283574476/IntelliJ+settings 

Could you please enable the ones relating to import so that our commits are consistent on import order?

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/NonGradientOptimizer.java
```diff
@@ -89,30 +96,57 @@ private double optimize(List<Vertex> outputVertices) {
...
-            MAXIMIZE,
-            new InitialGuess(Optimizer.currentPoint(bayesianNetwork.getContinuousLatentVertices()))
-        );
+    private SimpleBounds getVertexSpecificBounds() {
+        List<Double> minBounds = new ArrayList<>();
+        List<Double> maxBounds = new ArrayList<>();
 
-        return pointValuePair.getValue();
+        vertexSpecificBounds.values().forEach(bounds -> {
+            minBounds.addAll(DoubleTensor.create(bounds.getLower()).asFlatList());
```
gordoncaleb: This bounds.getLower() will always be a single value for this vertex?

rory-improbable:
@gordoncaleb your new changes look great, I'm happy for this to be merged if you are

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/BoundsCalculator.java
```diff
@@ -0,0 +1,75 @@
+package io.improbable.keanu.algorithms.variational;
+
+import io.improbable.keanu.tensor.TensorShape;
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+import io.improbable.keanu.vertices.Vertex;
+import lombok.RequiredArgsConstructor;
+import org.apache.commons.math3.optim.SimpleBounds;
```
christophernorth: There was a previous comment mentioning that we shouldn't depend explicitly on this in Keanu - have we decided that it's ok to do that now?

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/OptimizerBounds.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    private Map<Vertex<? extends DoubleTensor>, VertexBounds> vertexBounds = new HashMap<>();
+
+    public void addBound(Vertex<? extends DoubleTensor> vertex, DoubleTensor min, DoubleTensor max) {
+        DoubleTensor minDup = min.duplicate();
+        DoubleTensor maxDup = max.duplicate();
+
+        vertexBounds.put(vertex, new VertexBounds(minDup, maxDup));
+
```
GeorgeNash: nit: white space

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/BoundsCalculator.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public SimpleBounds getBounds(List<? extends Vertex<DoubleTensor>> latentVertices, double[] startPoint) {
+        List<Double> minBounds = new ArrayList<>();
+        List<Double> maxBounds = new ArrayList<>();
+
+        for (Vertex<DoubleTensor> vertex : latentVertices) {
+
+            if (optimizerBounds.hasBound(vertex)) {
+                validateBoundsForVertex(vertex, optimizerBounds);
+                addBoundsForVertex(vertex, optimizerBounds, minBounds, maxBounds);
```
christophernorth: Should we drop the OptimizerBounds parameter from these two methods given we always just pass the member through to them?

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
```diff
@@ -70,6 +70,9 @@ static DoubleTensor placeHolder(int[] shape) {
     @Override
     DoubleTensor reshape(int... newShape);
 
+    @Override
+    DoubleTensor duplicate();
```
GeorgeNash: Feels like this should be on `Tensor` - I imagine boolean and generic implementations will want it?

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/BoundsCalculator.java
gordoncaleb: You good catch. It's a slip on the refactoring from pure functions.

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/BoundsCalculator.java
gordoncaleb: Good point. This class shouldn't be part of the public API and should be made package private and therefore just an implementation detail.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
gordoncaleb: It's on Tensor and GenericTensor but wasn't defined on the BooleanTensor interface. Added it.

PR #146, opened 2018-07-31
Make writer public

GeorgeNash:
So people can actually use it....

Thanks Chris!


christophernorth:
Seems to fix the issue - thanks for finding and fixing so quickly.

PR #148, opened 2018-08-01
pulled out Differentiable interface

migwellian:
also made getParents and getChildren return a List instead of a Set (which I will need later for refactoring the probabilistic vertex classes
also made getParents and getChildren return a defensive copy

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
@@ -2,7 +2,7 @@ package io.improbable.keanu.vertices;
 
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.HashSet;
+import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicLong;
@@ -20,8 +20,8 @@ public abstract class Vertex<T> {
     public static final AtomicLong ID_GENERATOR = new AtomicLong(0L);
 
     private long uuid = ID_GENERATOR.getAndIncrement();
-    private Set<Vertex> children = new HashSet<>();
-    private Set<Vertex> parents = new HashSet<>();
+    private Set<Vertex> children = Collections.emptySet();
+    private Set<Vertex> parents = Collections.emptySet();
     private T value;
     private boolean observed;
 
@@ -186,15 +186,15 @@ public abstract class Vertex<T> {
     }
 
     public Set<Vertex> getChildren() {
-        return ImmutableSet.copyOf(children);
+        return children;
     }
 
     public void addChild(Vertex<?> v) {
-        children.add(v);
+        children = ImmutableSet.<Vertex>builder().addAll(children).add(v).build();
     }
 
     public void setParents(Collection<? extends Vertex> parents) {
-        this.parents = new HashSet<>();
+        this.parents = Collections.emptySet();
         addParents(parents);
     }
 
@@ -203,16 +203,16 @@ public abstract class Vertex<T> {
     }
 
     public void addParents(Collection<? extends Vertex> parents) {
-        parents.forEach(this::addParent);
+        this.parents = ImmutableSet.<Vertex>builder().addAll(this.getParents()).addAll(parents).build();
+        parents.forEach(p -> p.addChild(this));
     }
 
     public void addParent(Vertex<?> parent) {
-        this.parents.add(parent);
-        parent.addChild(this);
+        addParents(ImmutableSet.of(parent));
     }
 
     public Set<Vertex> getParents() {
-        return ImmutableSet.copyOf(this.parents);
+        return parents;
     }
 
     @Override
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
@@ -1,14 +1,13 @@
 package io.improbable.keanu.vertices;
 
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.List;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicLong;
 
-import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 
 import io.improbable.keanu.algorithms.graphtraversal.DiscoverGraph;
 import io.improbable.keanu.algorithms.graphtraversal.VertexValuePropagation;
@@ -21,8 +20,8 @@ public abstract class Vertex<T> {
     public static final AtomicLong ID_GENERATOR = new AtomicLong(0L);
 
     private long uuid = ID_GENERATOR.getAndIncrement();
-    private List<Vertex> children = new ArrayList<>();
-    private List<Vertex> parents = new ArrayList<>();
+    private Set<Vertex> children = new HashSet<>();
+    private Set<Vertex> parents = new HashSet<>();
     private T value;
     private boolean observed;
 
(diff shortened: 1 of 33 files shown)


keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
```diff
@@ -1,21 +1,28 @@
...
-import java.util.concurrent.atomic.AtomicLong;
-
 public abstract class Vertex<T> {
 
     public static final AtomicLong ID_GENERATOR = new AtomicLong(0L);
 
     private long uuid = ID_GENERATOR.getAndIncrement();
-    private Set<Vertex> children = new HashSet<>();
-    private Set<Vertex> parents = new HashSet<>();
+    private List<Vertex> children = new ArrayList<>();
```
chris-major-improbable: Why are these `List`s rather than `Set`s? I can't see anything that specifies ordering in here, so a `set` seems more descriptive.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
```diff
@@ -1,19 +1,42 @@
...
 import io.improbable.keanu.vertices.dbl.nonprobabilistic.ConstantDoubleVertex;
 import io.improbable.keanu.vertices.dbl.nonprobabilistic.diff.DualNumber;
-import io.improbable.keanu.vertices.dbl.nonprobabilistic.operators.binary.*;
-import io.improbable.keanu.vertices.dbl.nonprobabilistic.operators.unary.*;
-
-import java.util.*;
-import java.util.function.Function;
-
-public abstract class DoubleVertex extends ContinuousVertex<DoubleTensor> implements DoubleOperators<DoubleVertex> {
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.operators.binary.AdditionVertex;
```
chris-major-improbable: Is there a reason to expand all the imports?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/Differentiable.java
```diff
@@ -0,0 +1,28 @@
...
+
+import io.improbable.keanu.vertices.Vertex;
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.diff.DualNumber;
+
+public interface Differentiable {
+
+    DualNumber calculateDualNumber(Map<Vertex, DualNumber> dualNumbers);
+
+    default DualNumber getDualNumber() {
+        return new Differentiator().calculateDual((Vertex & Differentiable) this);
```
chris-major-improbable: this `new` is likely to lead to memory allocation - and may be more efficient as a static method...

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/Differentiator.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import io.improbable.keanu.vertices.Vertex;
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.diff.DualNumber;
+
+public class Differentiator {
+    public <V extends Vertex & Differentiable> DualNumber calculateDual(V vertex) {
```
chris-major-improbable: could be `static`?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
migwellian: Philosophical answer: https://stackoverflow.com/questions/147454/why-is-using-a-wild-card-with-a-java-import-statement-bad
Practical answer: because SonarLint complains about it

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
migwellian: This is pre-empting a change I've made on another feature branch that requires it (a Vertex backed by a Distribution needs to know how its parent vertices correspond to its parameters like mu and sigma)

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/Differentiable.java
migwellian: I think the GC will happily take care of this kind of memory allocation, won't it? But you're right, in this case the object has no state so there's no reason not to make `calculateDual` static - I will change it.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/Differentiator.java
migwellian: Done

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
chris-major-improbable: isn't that why we have `getMu()` and `getSigma()`?

I really wouldn't rely on ordering in a list as it'll be non-obvious that the order matters when it's constructed.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
chris-major-improbable: That makes sense - I guess it's not much of an issue with a modern IDE, and I probably wouldn't have noticed if it wasn't such a large amount of changes.

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
migwellian: I'm going to revert this for now, since it's not actually relevant to the Differentiable interface. I will revisit later since I'm going to change how we create the Distribution-backed Vertices anyway.

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
```diff
@@ -180,7 +186,7 @@ public long getId() {
     }
 
     public Set<Vertex> getChildren() {
-        return children;
+        return ImmutableSet.copyOf(children);
```
gordoncaleb: Creating these copies of children and parents will happen a lot. Could we store it has an immutable collection and return that? And then if we ever need to change the children or parents we could swap out the entire collection.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/Differentiator.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Map;
+import java.util.Set;
+
+import io.improbable.keanu.vertices.Vertex;
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.diff.DualNumber;
+
+public class Differentiator {
+    private Differentiator() {}
+
+    public static <V extends Vertex & Differentiable> DualNumber calculateDual(V vertex) {
```
gordoncaleb: Nice! Pure static functions!

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
migwellian: That sounds reasonable.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/Differentiable.java
```diff
@@ -0,0 +1,28 @@
...
+
+public interface Differentiable {
+
+    DualNumber calculateDualNumber(Map<Vertex, DualNumber> dualNumbers);
+
+    default DualNumber getDualNumber() {
+        return Differentiator.calculateDual((Vertex & Differentiable) this);
+    }
+
+    static <V extends Vertex & Differentiable> List<V> filter(List<? extends Vertex<?>> vertices) {
```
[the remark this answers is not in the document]
migwellian: It's keeping only those that are Differentiable. So we will have `Differentiable.filter` and `Probabilistic.filter`

I could name it `keepOnlyDifferentiableVertices`if you like?

PR #150, opened 2018-08-01
Add factory method for linspace and arange

gordoncaleb:
This PR creates a shorthand for ND4js linspace and arange

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
@@ -63,8 +63,23 @@ public interface DoubleTensor extends NumberTensor<Double> {
         return Nd4jDoubleTensor.linspace(start, end, numberOfPoints);
     }
 
+    /**
+     * @param start start of range
+     * @param end   end of range (exclusive)
+     * @return a vector of numbers from start incrementing by one to end (exclusively)
+     */
     static DoubleTensor arange(double start, double end) {
         return Nd4jDoubleTensor.arange(start, end);
+    }
+
+    /**
+     * @param start    start of range
+     * @param end      end of range (exclusive)
+     * @param stepSize size of step from start to end
+     * @return a vector of numbers starting at start and stepping to end (exclusively)
+     */
+    static DoubleTensor arange(double start, double end, double stepSize) {
+        return Nd4jDoubleTensor.arange(start, end, stepSize);
     }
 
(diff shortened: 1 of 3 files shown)


chris-major-improbable:
All looks good, however I'm not sure these names  are particularly informative.

e..g `linspace`, I'm assuming is linear spacing? feels odd to specify max, min and number of steps, rather than max, min and interval (as you would in a for loop).

in the same way `arange` isn't clear what sort of range it is (from the tests it looks like it generates an inclusive integer sequence - but I wouldn't have guessed that from the name)

gordoncaleb:
I definitely agree that the names aren't ideal. The problem is that these names are so ubiquitous in other tools that I'd hesitate to give it the name it deserves due to users thinking it's somehow different. I'll ask around if anyone has any better names...

keanu-project/src/test/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensorTest.java
```diff
@@ -247,6 +247,20 @@ public void canBroadcastDivide() {
...
+    public void canLinSpace(){
+        DoubleTensor actual = DoubleTensor.linspace(0, 10, 5);
+        DoubleTensor expected = DoubleTensor.create(new double[]{0, 2.5, 5.0, 7.5, 10.0});
+        assertEquals(expected, actual);
+    }
+
+    @Test
+    public void canARange(){
+        DoubleTensor actual = DoubleTensor.arange(0, 5);
+        DoubleTensor expected = DoubleTensor.create(new double[]{0, 1, 2, 3, 4});
```
migwellian: should this return an IntegerTensor?

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
```diff
@@ -59,6 +59,14 @@ static DoubleTensor zeros(int[] shape) {
         }
     }
 
+    static DoubleTensor linspace(double start, double end, int numberOfPoints) {
+        return Nd4jDoubleTensor.linspace(start, end, numberOfPoints);
+    }
+
+    static DoubleTensor arange(double start, double end) {
```
GeorgeNash: Can we add a `step` or `interval` to this so that we can making things like:

`arange(0.0, 10.0, 2.0)` = `[0.0, 2.0, 4.0, 6.0, 8.0, 10.0]`?

https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.arange.html

PR #151, opened 2018-08-03
Dirichlet distribution

GeorgeNash:
Dirichlet: https://en.wikipedia.org/wiki/Dirichlet_distribution
Which is a multivariate: https://en.wikipedia.org/wiki/Beta_distribution

So a 2D dirichlet equates to a beta (as the sum of all dirichlet values must always be 1.)


keanu-project/src/main/java/io/improbable/keanu/tensor/Tensor.java
```diff
@@ -88,7 +88,11 @@ default boolean isScalar() {
     }
 
     default boolean isVector() {
-        return getRank() == 1;
+        if (getRank() == 2) {
```
icantsayrural: Can this be simplified to `getRank() == 2 && (getShape()[0] == 1 || getShape()[1] == 1)`?

GeorgeNash: Good spot - I've updated it

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/Dirichlet.java
```diff
@@ -0,0 +1,62 @@
...
+        final DoubleTensor gammaSamples = gamma.sample(concentration.getShape(), random);
+        return normalise(gammaSamples);
+    }
+
+    @Override
+    public DoubleTensor logProb(DoubleTensor x) {
+        if (Math.abs(x.sum() - 1.0) > EPSILON) {
+            throw new IllegalArgumentException("Sum of dirichlet values to calculate likelihood for must equal 1.");
+        }
+        final double sumConcentrationLogged = concentration.minus(1.).timesInPlace(x.log()).sum();
```
icantsayrural: This might be a stupid question, but can `x` have 0's here?

keanu-project/src/main/java/io/improbable/keanu/tensor/Tensor.java
```diff
@@ -88,7 +88,7 @@ default boolean isScalar() {
     }
 
     default boolean isVector() {
-        return getRank() == 1;
+        return getRank() == 2 && (getShape()[0] == 1 || getShape()[1] == 1);
```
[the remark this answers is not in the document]
GeorgeNash: Yep - good call.

Went for:

```
    /**
     * Returns true if the tensor is a vector. A vector being a 1xn or a nx1 tensor.
     *
     * (1, 2, 3) is a 1x3 vector.
     *
     * (1)
     * (2)
     * (3) is a 3x1 vector.
     *
     *
     * @return true if the tensor is a vector
     */
```

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -725,6 +725,21 @@ public double determinant() {
...
+        if (this.isScalar()) {
+            return this.scalar();
+        } else if (this.isVector()) {
+            if (this.getShape()[0] > 1) {
+                return tensor.prod(0).getDouble(0);
+            } else {
+                return tensor.prod(1).getDouble(0);
+            }
+        } else {
+            throw new IllegalArgumentException("Can only calculate product on vector or scalar tensorsa");
```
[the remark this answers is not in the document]
GeorgeNash: Nice - good call

christophernorth: "tensora" -> "tensor"

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/probabilistic/DirichletVertexTest.java
```diff
@@ -0,0 +1,237 @@
...
+        double fourDimDirichletPdf2 = dirichlet.logPdf(DoubleTensor.create(new double[]{0.7, 0.1, 0.1, 0.1}));
+        double fourDimDirichletPdf3 = dirichlet.logPdf(DoubleTensor.create(new double[]{0.25, 0.25, 0.25, 0.25}));
+
+        Assert.assertTrue(fourDimDirichletPdf1 == fourDimDirichletPdf2 && fourDimDirichletPdf2 == fourDimDirichletPdf3);
+    }
+
+    @Test
+    public void matchesScipyDirichletLogPdf() {
+        DirichletVertex dirichlet = new DirichletVertex(2, 2);
+        double scipyPdf = 0.8662499999999997;
```
christophernorth: This feels potentially fragile, but I can't really see a better way of dealing with this short of committing a script to generate the correct number at runtime.  If we have a different implementation, perhaps we could check a few more values rather than just one though?

migwellian: Is the problem that we don't have a Java library to check against, like apache?

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/Dirichlet.java
```diff
@@ -0,0 +1,62 @@
...
+        final double sumLogGammaConcentration = concentration.apply(org.apache.commons.math3.special.Gamma::gamma).logInPlace().sum();
+        final double logGammaSumConcentration = Math.log(org.apache.commons.math3.special.Gamma.gamma(concentration.sum()));
+        return DoubleTensor.scalar(sumConcentrationLogged - sumLogGammaConcentration + logGammaSumConcentration);
+    }
+
+    @Override
+    public Diffs dLogProb(DoubleTensor x) {
+        final DoubleTensor dLogPdc = x.log().
+            minusInPlace(concentration.apply(org.apache.commons.math3.special.Gamma::digamma)).
```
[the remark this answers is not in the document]
GeorgeNash: Nice - changed

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/Dirichlet.java
GeorgeNash: Yep x can be equal to zero but can't go negative. At zero it should go to negative infinity

keanu-project/build.gradle
```diff
@@ -39,6 +39,7 @@ dependencies {
     testCompile 'org.jfree:jfreechart:1.0.19'
     testCompile group: 'org.mockito', name: 'mockito-core', version: '2.15.0'
     testCompile 'org.hamcrest:hamcrest-library:1.3'
+    testCompile 'ca.umontreal.iro.simul:ssj:3.2.1'
```
migwellian: This library hasn't been touched in 6 years... perhaps we should use this one instead which is actively developed: https://github.com/stanfordnlp/CoreNLP/blob/master/src/edu/stanford/nlp/stats/Dirichlet.java

christophernorth: As discussed - that one doesn't look ideal either (some of the functions are just stubs).  The Dirichlet implementation seems sound (it agrees with Georges) so are we ok to stick with this?

migwellian: agreed

keanu-project/src/main/java/io/improbable/keanu/tensor/Tensor.java
```diff
@@ -87,8 +87,20 @@ default boolean isScalar() {
...
+     *
+     * (1)
+     * (2)
+     * (3) is a 3x1 vector.
+     *
+     *
+     * @return true if the tensor is a vector
+     */
     default boolean isVector() {
-        return getRank() == 1;
```
gordoncaleb: Having a rank of 1 is also a vector. Even if we don’t actually support rank 1 at the moment.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -725,6 +725,21 @@ public double determinant() {
         return new LUDecomposition(matrix).getDeterminant();
     }
 
+    @Override
+    public double product() {
+        if (this.isScalar()) {
+            return this.scalar();
+        } else if (this.isVector()) {
+            if (this.getShape()[0] > 1) {
+                return tensor.prod(0).getDouble(0);
```
gordoncaleb: Is there not a prod() that would work for both types of vectors? Can we make it work for arbitrarily ranked tensors as well?

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/probabilistic/DirichletVertexTest.java
```diff
@@ -0,0 +1,243 @@
...
+
+        Assert.assertTrue(fourDimDirichletPdf1 == fourDimDirichletPdf2 && fourDimDirichletPdf2 == fourDimDirichletPdf3);
+    }
+
+    @Test
+    public void matchesMontrealDirichletLogPdf() {
+        DirichletDist baseline = new DirichletDist(new double[]{2, 2});
+        DirichletVertex keanu = new DirichletVertex(2, 2);
+
+        Assert.assertEquals(Math.log(baseline.density(new double[]{0.175, 0.825})),
```
gordoncaleb: If we only assert against this baseline for the 2 x case, isn’t that the same as using the BetaDistribution from Apache math commons?

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
```diff
@@ -177,6 +177,8 @@ static DoubleTensor placeHolder(int[] shape) {
 
     double determinant();
 
+    double product();
```
gordoncaleb: Is this for the Dirichlet?

keanu-project/src/main/java/io/improbable/keanu/tensor/Tensor.java
GeorgeNash: As discussed, I've added the rank one check back in

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
GeorgeNash: Nice, added the zero constructor prod which works for all tensor shapes

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/probabilistic/DirichletVertexTest.java
GeorgeNash: Yeah fair enough - made this a 3d test

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
GeorgeNash: Didn't need it in the end but thought it would be worth keeping

PR #152, opened 2018-08-03
Renames

GeorgeNash:
The renames suggested in the feedback session.

Select -> Categorical
Flip -> Bernoulli
Pluck -> Take


migwellian:
Strictly speaking, we ought to have a `Bernoulli` class in `io.improbable.keanu.distributions.discrete` too, which `BernoulliVertex` calls on.

migwellian:
And similarly a `Categorical` distribution

migwellian:
Looks great :-)

PR #154, opened 2018-08-06
youCanObserveANonProbabilisticVariableAfterCreatingTheNetwork

migwellian:
There was a bug whereby the `BayesianNetwork` stores the latent and observed variables on construction only, which means if you observe a non-probabilistic vertex later it won't be added to this list.

The solution is to calculate the latent and observed variables dynamically. Of course this has an impact on performance in time and space, but we agreed that it's more important at this stage to be correct and to optimise later.

If we want to optimise later, we'd need some kind of callback mechanism whereby a `Vertex` notifies a `BayesianNetwork` when it gets observed.

### DIFF
--- a/keanu-project/src/test/java/io/improbable/keanu/network/BayesianNetworkTest.java
+++ b/keanu-project/src/test/java/io/improbable/keanu/network/BayesianNetworkTest.java
@@ -10,7 +10,7 @@ import org.junit.Before;
 import org.junit.Test;
 
 import io.improbable.keanu.vertices.bool.BoolVertex;
-import io.improbable.keanu.vertices.bool.probabilistic.Flip;
+import io.improbable.keanu.vertices.bool.probabilistic.BernoulliVertex;
 
 public class BayesianNetworkTest {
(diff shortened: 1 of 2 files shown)


keanu-project/src/test/java/io/improbable/keanu/network/BayesianNetworkTest.java
```diff
@@ -0,0 +1,50 @@
...
+        assertThat(network.getObservedVertices(), is(empty()));
+        input1.observe(true);
+        assertThat(network.getObservedVertices(), contains(input1));
+        input2.observe(true);
+        assertThat(network.getObservedVertices(), containsInAnyOrder(input1, input2));
+        input1.unobserve();
+        assertThat(network.getObservedVertices(), contains(input2));
+    }
+
+    @Test
```
migwellian: This test would fail without the above code changes in `BayesianNetwork`

migwellian:
Hehe, it’s a copy and paste from my test name

PR #155, opened 2018-08-06
Feature/standardise probabilistic vertex classes

migwellian:
* The main change here is that the classes `{Double,Integer,Bool}{Unary,Binary}OpVertex` now take a functional object as a constructor, so that each concrete child class can be declared simply and concisely
  * also another optional functional object if it's differentiable
  * for example:
```
public class MultiplicationVertex extends DoubleBinaryOpVertex {
    public MultiplicationVertex(DoubleVertex left, DoubleVertex right) {
        super(left, right,
            (l, r) -> l.times(r),
            (l, r) -> l.multiplyBy(r));
    }
}
```
  * there's no longer a need for `{Double,Integer,Bool}{Unary,Binary}OpLambda` as they did the same thing
* Moved `pow` up from `DoubleOperators` to `Operators`
  * Improved the implementation of `IntegerTensor#pow` (use Guava IntMath library)
* Moved lots of methods up from `DoubleTensor` to `NumberTensor`
* `DoubleVertex`, `IntegerVertex` and `BoolVertex` now all implement equality and comparison operators
  * future work: make this a Kotlin interface too?


keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/FloorVertex.java
```diff
@@ -16,16 +12,6 @@
...
-    }
-
-    @Override
-    protected DoubleTensor op(DoubleTensor a) {
-        return a.floor();
-    }
-
-    @Override
-    public DualNumber calculateDualNumber(Map<Vertex, DualNumber> dualNumbers) {
-        throw new UnsupportedOperationException();
```
GeorgeNash: So what happens now that we're not throwing an exception on Vertices that can't have their dual number calculated? Does it silently fail or will it still break?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/RoundVertex.java
```diff
@@ -16,16 +13,6 @@
...
-
-    @Override
-    protected DoubleTensor op(DoubleTensor a) {
-        return a.round();
-    }
-
-    @Override
-    public DualNumber calculateDualNumber(Map<Vertex, DualNumber> dualNumbers) {
-        throw new UnsupportedOperationException();
+        super(Tensor.SCALAR_SHAPE, inputVertex, DoubleTensor::round, null);
```
GeorgeNash: This looks a bit rogue, why are we setting a scalar shape here?

I think you want: 
```
public RoundVertex(DoubleVertex inputVertex) {
        super(inputVertex, DoubleTensor::round);
    }
```

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/SigmoidVertex.java
```diff
@@ -17,22 +14,16 @@
...
+        super(inputVertex,
+            a -> a.unaryMinus().expInPlace().plusInPlace(1).reciprocalInPlace(),
+            a -> op(a));
     }
 
-    @Override
-    public DualNumber calculateDualNumber(Map<Vertex, DualNumber> dualNumbers) {
-        DualNumber dualNumber = dualNumbers.get(inputVertex);
-        DoubleTensor x = dualNumber.getValue();
+    public static DualNumber op(DualNumber a) {
```
GeorgeNash: Can we move this to the `DualNumber` class as a `sigmoid` function?

keanu-project/src/main/java/io/improbable/keanu/vertices/intgr/IntegerVertex.java
```diff
@@ -117,6 +136,31 @@ public IntegerVertex slice(int dimension, int index) {
         return new IntegerSliceVertex(this, dimension, index);
     }
 
+
+    public BoolVertex equalTo(IntegerVertex rhs) {
```
GeorgeNash: I know they're mostly pass through, but it would be nice to have tests covering them just in case

keanu-project/src/main/java/io/improbable/keanu/vertices/intgr/nonprobabilistic/operators/binary/IntegerBinaryOpVertex.java
```diff
@@ -1,28 +1,50 @@
...
+        this(checkHasSingleNonScalarShapeOrAllScalar(a.getShape(), b.getShape()), a, b, op);
+    }
+
+    /**
+     * A vertex that performs a user defined operation on two input vertices
+     * @param shape the shape of the tensor
+     * @param a first input vertex
+     * @param b second input vertex
+     * @param op operation used to sample
+     */public IntegerBinaryOpVertex(int[] shape, IntegerVertex a, IntegerVertex b, BinaryOperator<IntegerTensor> op) {
```
GeorgeNash: Chuck a new line in here

keanu-project/src/main/java/io/improbable/keanu/vertices/intgr/nonprobabilistic/operators/binary/IntegerPowerVertex.java
```diff
@@ -0,0 +1,16 @@
+package io.improbable.keanu.vertices.intgr.nonprobabilistic.operators.binary;
+
+import io.improbable.keanu.tensor.intgr.IntegerTensor;
+import io.improbable.keanu.vertices.intgr.IntegerVertex;
+
+public class IntegerPowerVertex extends IntegerBinaryOpVertex {
+    /**
+     * Multiplies one vertex by another
```
GeorgeNash: Raises one vertex to the power of another

keanu-project/src/main/java/io/improbable/keanu/vertices/intgr/nonprobabilistic/operators/unary/IntegerUnaryOpVertex.java
```diff
@@ -1,33 +1,43 @@
...
     /**
      * A vertex that performs a user defined operation on a singe input vertex
-     *
-     * @param shape the shape of the resulting vertex
-     * @param inputVertex a vertex
-     */
-    public IntegerUnaryOpVertex(int[] shape, IntegerVertex inputVertex) {
+     * @param inputVertex the input vertex
+     * @param op operation used to sample
+     */   public IntegerUnaryOpVertex(IntegerVertex inputVertex, Function<IntegerTensor,IntegerTensor> op) {
```
GeorgeNash: Needs a quick format

keanu-project/src/test/java/io/improbable/keanu/vertices/bool/nonprobabilistic/GreaterThanOrEqualVertexTest.java
```diff
@@ -25,12 +24,12 @@ public void comparesDoubles() {
     }
 
     private void isGreaterThanOrEqual(int a, int b, boolean expected) {
-        GreaterThanOrEqualVertex<IntegerTensor, IntegerTensor> vertex = new GreaterThanOrEqualVertex<>(ConstantVertex.of(a), ConstantVertex.of(b));
+        BoolVertex vertex = ConstantVertex.of(a).greaterThanOrEqualTo(ConstantVertex.of(b));
```
GeorgeNash: cool!

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/binary/AdditionVertex.java
```diff
@@ -18,19 +11,9 @@
...
     public AdditionVertex(DoubleVertex left, DoubleVertex right) {
-        super(checkHasSingleNonScalarShapeOrAllScalar(left.getShape(), right.getShape()), left, right);
-    }
-
-    @Override
-    public DualNumber calculateDualNumber(Map<Vertex, DualNumber> dualNumbers) {
-        DualNumber leftDual = dualNumbers.get(left);
-        DualNumber rightDual = dualNumbers.get(right);
-        return leftDual.plus(rightDual);
+        super(left, right,
```
gordoncaleb: Converting the op and dualNumber calculation to a lambda adds complexity IMHO and increases the memory footprint. I don't see us getting much out of it other than removing the *OpLambdas and removing a couple of boilerplate lines. I do really like removing the calculateDualNumbers on the ops up to the op and then calling a method defined by the op implementation that just requires it's left and right dual number instead of the entire world of dual numbers. I also like but am not sure about having the default shape check be an element-wise operations shape check. On one hand it is definitely by far the most common case and removing the boilerplate is nice but on the other hand if you don't require a shape from the op then someone creating a new op might not know that this shape constraint has been introduced. 

Can we keep the abstract op method and create an abstract dualNumberOp method that represents these two lambdas? Also, lets keep the default shaping constructor (good call) as I think other shape checks are uncommon (i.e. matrix multiply).

keanu-project/src/main/java/io/improbable/keanu/vertices/intgr/IntegerVertex.java
```diff
@@ -117,6 +136,31 @@ public IntegerVertex slice(int dimension, int index) {
...
+        return new EqualsVertex<>(this, rhs);
+    }
+
+    public <T extends Tensor> BoolVertex notEqualTo(Vertex<T> rhs) {
```
gordoncaleb: Nice!

keanu-project/src/main/java/io/improbable/keanu/tensor/NumberTensor.java
```diff
@@ -1,14 +1,75 @@
 package io.improbable.keanu.tensor;
 
+import java.util.function.Function;
+
+import io.improbable.keanu.tensor.bool.BooleanTensor;
 import io.improbable.keanu.tensor.dbl.DoubleTensor;
 import io.improbable.keanu.tensor.intgr.IntegerTensor;
 
-public interface NumberTensor<T extends Number> extends Tensor<T> {
+public interface NumberTensor<N extends Number, T extends NumberTensor<N,T>> extends Tensor<N> {
```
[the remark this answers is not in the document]
migwellian: Yes, this is a cheeky little trick and it's not at all intuitive, but it is important. It enables you to define the interface in terms of the concrete class that implements it. 

Here's a simple example:

    class Scratch {

        interface Number<N extends Number> {
            Number<N> doGenericThing(Number<N> other);
            N doGenericThingBetter(N other);
        }

        static class RealNumber implements Number<RealNumber> {
            @Override
            public RealNumber doGenericThing(Number<RealNumber> other) {
                print(other);                   // compiler error: print(RealNumber) cannot be applied to Number<RealNumber>
                return other.doSpecialThing();  // compiler error: Cannot resolve method
            }

            @Override
            public RealNumber doGenericThingBetter(RealNumber other) {
                print(other);
                return other.doSpecialThing();
            }

            RealNumber doSpecialThing() { return new RealNumber(); }

        }

        static void print(RealNumber value) {
            System.out.println(value);
        }
    }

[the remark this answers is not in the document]
migwellian: Yes, this is a cheeky little trick and it's not at all intuitive, but it is important. It enables you to define the interface in terms of the concrete class that implements it. 

Notice that the methods in the interface return a `T`, not a `NumberTensor`. That means that when you call one of these methods on a concrete types (like `DoubleTensor`), you take in a `DoubleTensor` and you get back a `DoubleTensor` - not a `NumberTensor`. 

Here's a simple example:

    class Scratch {

        interface Number<N extends Number> {
            Number<N> doGenericThing(Number<N> other);
            N doGenericThingBetter(N other);
        }

        static class RealNumber implements Number<RealNumber> {
            @Override
            public RealNumber doGenericThing(Number<RealNumber> other) {
                print(other);                   // compiler error: print(RealNumber) cannot be applied to Number<RealNumber>
                return other.doSpecialThing();  // compiler error: Cannot resolve method
            }

            @Override
            public RealNumber doGenericThingBetter(RealNumber other) {
                print(other);
                return other.doSpecialThing();
            }

            RealNumber doSpecialThing() { return new RealNumber(); }

        }

        static void print(RealNumber value) {
            System.out.println(value);
        }
    }

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/FloorVertex.java
migwellian: It means we're enforcing it at compile time instead of runtime (i.e. this class does not implement `Differentiable`).

If you were to try to cast an object of this type to a `Differentiable` you'd get a `ClassCastException`.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/SigmoidVertex.java
migwellian: Why? Would it get called anywhere else?

keanu-project/src/main/java/io/improbable/keanu/tensor/NumberTensor.java
```diff
@@ -1,14 +1,75 @@
...
 
-    T sum();
+    N sum();
 
     DoubleTensor toDouble();
 
     IntegerTensor toInteger();
 
+
+    T reshape(int... newShape);
```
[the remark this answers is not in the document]
migwellian: Yes you're right - I've removed it, it was redundant.

keanu-project/src/main/java/io/improbable/keanu/tensor/NumberTensor.java
```diff
@@ -1,14 +1,75 @@
...
+
+    T setWithMaskInPlace(T mask, N value);
+
+    T setWithMask(T mask, N value);
+
+    T apply(Function<N, N> function);
+    
+    // In Place
+
+    T minusInPlace(T that);
```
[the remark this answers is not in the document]
migwellian: They are declared in Operators.kt. 

I could make `NumberTensor` extend `Operators<T>`  - that would make the association with the non-in-place versions clearer, and enforce that any class that implements `NumberTensor` also implements `Operators`.

keanu-project/src/main/java/io/improbable/keanu/tensor/intgr/ScalarIntegerTensor.java
```diff
@@ -295,7 +297,7 @@ public IntegerTensor powInPlace(IntegerTensor exponent) {
 
     @Override
     public IntegerTensor powInPlace(int exponent) {
-        value = (int) Math.pow(value, exponent);
+        value = IntMath.pow(value, exponent);
```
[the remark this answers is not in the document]
migwellian: I think the emphasis is on correctness, but it's also efficient.

From the javadoc:

_"Even if the result overflows, it will be equal to BigInteger.valueOf(b).pow(k).intValue(). This implementation runs in O(log k) time."_

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/binary/AdditionVertex.java
migwellian: Sure, I'm not too opposed to using an abstract protected method instead of a lambda function field.

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/nonprobabilistic/operators/multiple/OrMultipleVertex.java
```diff
@@ -1,26 +1,19 @@
...
             ),
-            input, OrMultipleVertex::or);
-    }
-
-    /**
-     * Returns true if one of the vertices is true
-     */
-    private static BooleanTensor or(BooleanTensor a, BooleanTensor b) {
-        return a.or(b);
+            input, (a,b) -> a.or(b));
```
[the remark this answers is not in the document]
migwellian: Done - and a bunch of others too.

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/BoolVertex.java
```diff
@@ -39,11 +43,19 @@ public static final BoolVertex not(Vertex<BooleanTensor> vertex) {
         return new NotVertex(vertex);
     }
 
+    public BoolVertex equalTo(BoolVertex rhs) {
```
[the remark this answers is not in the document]
migwellian: Done .- please take a look at `BoolVertexTest#doesEquals`

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/nonprobabilistic/operators/multiple/AndMultipleVertex.java
```diff
@@ -1,24 +1,17 @@
...
             input.stream().map(Vertex::getShape).collect(Collectors.toList())
-        ), input, AndMultipleVertex::and);
-    }
-
-    /**
-     * Returns true if all vertices are true
-     */
-    private static BooleanTensor and(BooleanTensor a, BooleanTensor b) {
-        return a.and(b);
+        ), input, (a, b) -> a.and(b));
```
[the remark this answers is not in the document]
migwellian: Done - and a bunch of others too.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/DoubleUnaryOpVertex.java
```diff
@@ -1,32 +1,76 @@
...
-     *
-     * @param shape       the shape of the resulting vertex
-     * @param inputVertex a vertex
+     * A vertex that performs a user defined operation on a single input vertex
+     * @param inputVertex the input vertex
+     * @param op operation used to sample
      */
-    public DoubleUnaryOpVertex(int[] shape, DoubleVertex inputVertex) {
-        super(new NonProbabilisticValueUpdater<>(v -> ((DoubleUnaryOpVertex) v).op(inputVertex.getValue())));
+    public DoubleUnaryOpVertex(DoubleVertex inputVertex, Function<DoubleTensor,DoubleTensor> op) {
```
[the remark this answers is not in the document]
migwellian: Done... I wish IntelliJ could do this on auto-save.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/RoundVertex.java
migwellian: thanks

keanu-project/src/main/java/io/improbable/keanu/vertices/intgr/IntegerVertex.java
migwellian: Done - please take a look at `IntegerVertexTest#doesEqualTo`

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/BoolVertex.java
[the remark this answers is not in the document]
GeorgeNash: nice

keanu-project/src/main/java/io/improbable/keanu/tensor/NumberTensor.java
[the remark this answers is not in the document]
GeorgeNash: Sweet man - thanks

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/DoubleUnaryOpVertex.java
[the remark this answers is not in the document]
GeorgeNash: RIP Eclipse

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/SigmoidVertex.java
GeorgeNash: It's just a convention we have, keeps the OpVertex fairly light and means you can the `DualNumber` class for any operations!

keanu-project/src/main/java/io/improbable/keanu/tensor/NumberTensor.java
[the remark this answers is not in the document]
migwellian: ^^^ done

PR #156, opened 2018-08-08
Ensure full range of integer is supported in IntegerTensor

gordoncaleb:
This PR has two improvements:

- The integer tensor was backed by a single precision float which doesn't give the integer tensor expected range as the java int. This is now backed by a double which does give it the expected range.

- Negative Integer division was wrong because flooring floating point operations isn't the same as integer math :(. For the ND4J backed IntegerTensor on division, the entire buffer will be brought back on heap and converted to integers before being sent back down in order to ensure the correct result.


keanu-project/src/main/java/io/improbable/keanu/tensor/intgr/Nd4jIntegerTensor.java
```diff
@@ -21,6 +21,10 @@
 
 public class Nd4jIntegerTensor implements IntegerTensor {
 
+    static {
+        System.setProperty("dtype", "double");
```
migwellian: it's probably worth adding a comment here about why we need this line

migwellian: The ND4J docs say that you can also set this with:

    Nd4j.setDataType(DataBuffer.Type.DOUBLE);

which is probably preferable

keanu-project/src/test/java/io/improbable/keanu/tensor/intgr/Nd4jIntegerTensorTest.java
```diff
@@ -383,4 +383,56 @@ public void canBroadcastDivide() {
...
+        assertDropsFractionCorrectlyOnDivision(5, 3);
+    }
+
+    @Test
+    public void doesNegativeDivisionCorrectly() {
+        assertDropsFractionCorrectlyOnDivision(-5, 3);
+    }
+
+    @Test
+    public void canStartAsMaxInteger() {
```
christophernorth: I think the tests could give us even more confidence if we didn't just concentrate on the max/min values in the integer range, but instead ran over a subset of the integer range (after all, the problem isn't that the single precision float didn't cover the integer range - it's more that it can't represent all of the numbers within that range).  I wrote a quick program and Double is able to represent every 32bit integer so we should be fine - but can't hurt to check a wider range?

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -25,11 +30,8 @@
 
 public class Nd4jDoubleTensor implements DoubleTensor {
 
-    static {
-        System.setProperty("dtype", "double");
-    }
-
     public static Nd4jDoubleTensor scalar(double scalarValue) {
+        Nd4j.setDataType(DataBuffer.Type.DOUBLE);
```
GeorgeNash: Pull this `DataBuffer.Type.DOUBLE` out into a public static?

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -463,6 +470,7 @@ public DoubleTensor atan2InPlace(double y) {
     @Override
     public DoubleTensor atan2InPlace(DoubleTensor y) {
         if (y.isScalar()) {
+            Nd4j.setDataType(DataBuffer.Type.DOUBLE);
```
GeorgeNash: What's this doing here?

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -638,7 +646,8 @@ public DoubleTensor setWithMaskInPlace(DoubleTensor mask, double value) {
         INDArray maskDup = unsafeGetNd4J(mask).dup();
 
         if (value == 0.0) {
-            INDArray swapOnesForZeros = Nd4j.ones(tensor.shape()).subi(maskDup);
+            INDArray swapOnesForZeros = maskDup.rsub(1.0);
+            ;
```
GeorgeNash: Rogue `;`

keanu-project/src/main/java/io/improbable/keanu/tensor/intgr/Nd4jIntegerTensor.java
gordoncaleb: Yeah that has to be added before every call to create a new INDArray due to ND4J defaulting to FLOAT. I've created a TypedINDArrayFactory that requires you specify the data buffer type before creation instead of letting ND4J default.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
gordoncaleb: Yup good call.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
gordoncaleb: The ones that are created need to be a double buffer or things will break. I've factored out the creation of all INDArrays to a class that forces you to provide a buffer type instead of defaulting to what ND4J thinks. Also there's `tensor.rsubi(double x)` which when x is 1 is the same as saying `ones(..).subi(tensor)`. Using rsubi removes the need to do a new buffer creation.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
gordoncaleb: Good eye.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
GeorgeNash: cool

Issue #91, opened 2018-06-07
Improve error message when trying to use variational methods on a non-continuous network


dmculley:
**Is your feature request related to a problem? Please describe.**
Obviously trying to use GradientOptimizer on a graph with Poissons in is silly but it would be good to have reliable error messaging as a safety net.

christophernorth:
An UnsupportedOperationException will now be reliably thrown if you try and use a gradient optimiser on a graph containing discrete Vertices.

PR #158, opened 2018-08-10
Feature/no gradient on discrete

christophernorth:
Fix for issue #91 

Simply throw an exception when you try and construct the GradientOptimiser when the graph contains latent Integer Vertices.  Simply looking for *any* Discrete Latent doesn't work as we need to be able to run the Gradient Optimiser over 'If' Vertices (and there are tests that rely on this).

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/gradient/GradientOptimizer.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/gradient/GradientOptimizer.java
@@ -47,7 +47,7 @@ public class GradientOptimizer implements Optimizer {
 
         if (numIntegerVertices != 0) {
             throw new UnsupportedOperationException("Gradient Optimization unsupported on Networks containing " +
-                "Discrete Latents");
+                "Discrete Latents (" + numIntegerVertices + " found)");
         }
 
         return GradientOptimizer.builder()
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/gradient/GradientOptimizer.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/gradient/GradientOptimizer.java
@@ -8,6 +8,7 @@ import java.util.Collection;
 import java.util.List;
 import java.util.function.BiConsumer;
 
+import io.improbable.keanu.vertices.intgr.IntegerVertex;
 import org.apache.commons.math3.optim.InitialGuess;
 import org.apache.commons.math3.optim.MaxEval;
 import org.apache.commons.math3.optim.PointValuePair;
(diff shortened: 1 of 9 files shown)


keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/gradient/GradientOptimizer.java
```diff
@@ -40,6 +41,15 @@
     }
 
     public static GradientOptimizer of(BayesianNetwork bayesNet) {
+        long numIntegerVertices = bayesNet.getDiscreteLatentVertices().stream()
+            .filter(v -> v instanceof IntegerVertex)
```
rory-improbable: is Integer vertex the only type of discretes we have? What about something like a categorical vertex?

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/gradient/GradientOptimizer.java
```diff
@@ -40,6 +41,15 @@
...
+            .count();
+
+        if (numIntegerVertices != 0) {
+            throw new UnsupportedOperationException("Gradient Optimization unsupported on Networks containing " +
+                "Discrete Latents");
```
rory-improbable: is it worth mentioning in this error message how many offending discrete vertices there are?

christophernorth: I had thought about that - but given we don't really give any indication of *which* vertices are causing the issue I wasn't sure it was that useful (ie you have to look through your code to figure out what you're doing wrong anyway).  I'll add though as a bit of extra information can't hurt!  When we have Vertex labelling, we'll be able to improve this sort of error further.

rory-improbable: ye good point, vertex lablels would be the only useful information maybe

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/gradient/GradientOptimizer.java
rory-improbable: also could extract this condition to a function called "isDiscreteVertex", which would make this more explicit, and also allow us to change the criteria for how we identify discrete vertices

gordoncaleb: If bayesNet.getDiscreteLatentVertices() returns anything other than an empty list then we should throw. Not need to filter for just integers.

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/gradient/GradientOptimizer.java
```diff
@@ -40,6 +41,15 @@
     }
 
     public static GradientOptimizer of(BayesianNetwork bayesNet) {
+        long numIntegerVertices = bayesNet.getDiscreteLatentVertices().stream()
+            .filter(v -> v instanceof IntegerVertex)
+            .count();
+
+        if (numIntegerVertices != 0) {
```
gordoncaleb:
```
if(!bayesNet.getDiscreteLatentVertices().isEmpty())
```

PR #166, opened 2018-08-14
Feature/vertex labelling

christophernorth:
Add the ability to Label Vertices and retrieve them from a BayesNet via the label.  Will detect repeated labels on Bayesnet creation and throw an exception.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
@@ -17,13 +17,22 @@ public class BayesianNetwork {
     private final List<? extends Vertex> vertices;
     private final HashMap<String, Vertex> labelSet;
 
+    private void addLabels(Set<? extends Vertex>vertices) {
+        for (Vertex v : vertices) {
+            String label = v.getLabel();
+            if (label != null && labelSet.containsKey(label)) {
+                throw new IllegalArgumentException("Vertex Label Repeated: " + label);
+            } else {
+                labelSet.put(label, v);
+            }
+        }
+    }
+
     public BayesianNetwork(Set<? extends Vertex> vertices) {
         this.vertices = ImmutableList.copyOf(vertices);
         this.labelSet = new HashMap<>();
 
-        vertices.stream()
-            .filter(v -> v.getLabel() != null)
-            .forEach(v -> this.labelSet.put(v.getLabel(), v));
+        addLabels(vertices);
     }
 
     public BayesianNetwork(Collection<? extends Vertex> vertices) {
--- a/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
@@ -1,6 +1,11 @@
 package io.improbable.keanu.network;
 
-import java.util.*;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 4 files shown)


keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -18,15 +15,34 @@
 public class BayesianNetwork {
 
     private final List<? extends Vertex> vertices;
+    private final HashMap<String, Vertex> labelSet;
+
+    private void addLabels(Set<? extends Vertex>vertices) {
```
gordoncaleb: the usual convention is that methods go after the constructor

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -18,15 +15,34 @@
 public class BayesianNetwork {
 
     private final List<? extends Vertex> vertices;
+    private final HashMap<String, Vertex> labelSet;
```
gordoncaleb: Since you don't often need to know the exact implementation of the Map, we define it to the lowest class that still satisfies our needs. In this case it's just `private final Map<String, Vertex> labelSet;`

gordoncaleb: `Set` is actually a type of collection in Java.  Can this be vertexLabels? Or labelToVertexMap?

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -18,15 +15,34 @@
...
+                throw new IllegalArgumentException("Vertex Label Repeated: " + label);
+            } else {
+                labelSet.put(label, v);
+            }
+        }
+    }
 
     public BayesianNetwork(Set<? extends Vertex> vertices) {
         this.vertices = ImmutableList.copyOf(vertices);
+        this.labelSet = new HashMap<>();
```
gordoncaleb: If you want to be more functional you could change your `addLabels(...)` to not have side-effects and be a pure static function that returns the label to Vertex map. `this.labelSet = getLabels(vertices)`. I tend to favour that approach but in this case it's a matter of preferred style.

keanu-project/src/test/java/io/improbable/keanu/network/BayesianNetworkTest.java
```diff
@@ -47,4 +51,38 @@ public void youCanObserveANonProbabilisticVariableAfterCreatingTheNetwork() {
         output.unobserve();
         assertThat(network.getObservedVertices(), is(empty()));
     }
+
+    @Test
+    public void youCanLabelVertices() {
+        BoolVertex a = new BernoulliVertex(0.5);
+        BoolVertex b = new BernoulliVertex(0.5);
+        BoolVertex ored = a.or(b);
+        BayesianNetwork net;
```
gordoncaleb: Is there a reason you don't declare net and assign it in the same line?

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -1,9 +1,6 @@
 package io.improbable.keanu.network;
 
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
icantsayrural: Are we ok with wild card imports?

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -18,15 +15,35 @@
...
+    }
+
+    private Map<String, Vertex> getLabels(Set<? extends Vertex>vertices) {
+        Map<String, Vertex> labelMap = new HashMap<>();
+        for (Vertex v : vertices) {
+            String label = v.getLabel();
+            if (label != null && labelMap.containsKey(label)) {
+                throw new IllegalArgumentException("Vertex Label Repeated: " + label);
+            } else {
+                labelMap.put(label, v);
```
icantsayrural: do we want to map a null label?

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -18,15 +15,35 @@
...
 
     public BayesianNetwork(Collection<? extends Vertex> vertices) {
         this(new HashSet<>(vertices));
     }
 
+    public Vertex getVertexByLabel(String label) {
+        return vertexLabels.get(label);
+    }
+
+    private Map<String, Vertex> getLabels(Set<? extends Vertex>vertices) {
```
icantsayrural: nit: space before parameter name

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
migwellian: See https://improbableio.atlassian.net/wiki/spaces/~616259109/pages/283574476/IntelliJ+settings for disabling wild card imports

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -18,15 +15,35 @@
 public class BayesianNetwork {
 
     private final List<? extends Vertex> vertices;
+    private final Map<String, Vertex> vertexLabels;
```
migwellian: Using`String` for the key will lead to issues around typos, capitalisation, white space, and type safety. By encapsulating that in a class we can avoid all those problems by overriding `equals` and `hashCode`.

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
migwellian: can be static

migwellian: A better name would be `buildLabelMap` since this method is not a getter

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -18,15 +15,35 @@
...
         this(new HashSet<>(vertices));
     }
 
+    public Vertex getVertexByLabel(String label) {
+        return vertexLabels.get(label);
+    }
+
+    private Map<String, Vertex> getLabels(Set<? extends Vertex>vertices) {
+        Map<String, Vertex> labelMap = new HashMap<>();
+        for (Vertex v : vertices) {
```
migwellian: You can do this more concisely with streams. This doesn't include the check for duplicates but you could do that easily enough by checking the size of the map.

        Map<String, ? extends Vertex> labelMap = vertices.stream()
            .filter(v -> v.getLabel() != null)
            .collect(Collectors.toMap(Vertex::getLabel, v -> v));

PR #168, opened 2018-08-14
Some work on `withParameters` API for distributions

icantsayrural:
I realise this is a controversial change. I'm open to feedback. Also, I didn't finish some tasks, because I wanted to first continue our slack discussion in a PR since we can all see the method signatures here.

- [x] Rename most parameters to [location](https://en.wikipedia.org/wiki/Location_parameter), [scale](https://en.wikipedia.org/wiki/Scale_parameter) or [shape](https://en.wikipedia.org/wiki/Shape_parameter) (alpha, beta, ...) where appropriate
- [x] Add a JavaDoc for `withParameters` method explaining the parameters in
detail, alternate names of the distribution and references to books/wiki
pages
- [x] Add a JavaDoc for sample methods that throw exceptions
- [ ] Clean up `Diff` parameter names
- [ ] Rename local variables (stopped making changes in case we don't go through with my suggestion)

Note: the shape parameter is not called `shape` because (1) the name
clashes with `shape` for tensors (2) there may be more than one shape
parameters (e.g. see Beta Distribution)

Best way to test this would be to enable JavaDocs on Intellji: go to File/Settings, Editor, click on General. Scroll down, then ✔ Show quick documentation on mouse move

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/ChiSquared.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/ChiSquared.java
@@ -21,6 +21,7 @@ public class ChiSquared implements ContinuousDistribution {
 
     /**
      * @param alpha shape parameter (not to be confused with tensor shape); number of degrees of freedom
+     * @return an instance of {@link ContinuousDistribution}
      */
     public static ContinuousDistribution withParameters(IntegerTensor alpha) {
         return new ChiSquared(alpha);
--- a/keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/ChiSquared.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/ChiSquared.java
@@ -8,19 +8,19 @@ import io.improbable.keanu.tensor.dbl.DoubleTensor;
 import io.improbable.keanu.tensor.intgr.IntegerTensor;
 import io.improbable.keanu.vertices.dbl.KeanuRandom;
 
+/**
+ * @see "Computer Generation of Statistical Distributions
+ * by Richard Saucier,
+ * ARL-TR-2168 March 2000,
+ * 5.1.4 page 16"
+ */
 public class ChiSquared implements ContinuousDistribution {
 
     private static final double LOG_TWO = Math.log(2);
     private final IntegerTensor alpha;
 
     /**
-     * <h3>Chi-Square Distribution</h3>
-     *
      * @param alpha shape parameter (not to be confused with tensor shape); number of degrees of freedom
-     * @see "Computer Generation of Statistical Distributions
-     * by Richard Saucier,
-     * ARL-TR-2168 March 2000,
-     * 5.1.4 page 16"
      */
     public static ContinuousDistribution withParameters(IntegerTensor alpha) {
         return new ChiSquared(alpha);
--- a/keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/ChiSquared.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/ChiSquared.java
@@ -17,30 +17,30 @@ import io.improbable.keanu.vertices.dbl.KeanuRandom;
 public class ChiSquared implements ContinuousDistribution {
 
     private static final double LOG_TWO = Math.log(2);
-    private final IntegerTensor alpha;
+    private final IntegerTensor degreesOfFreedom;
 
     /**
-     * @param alpha shape parameter (not to be confused with tensor shape); number of degrees of freedom
+     * @param degreesOfFreedom number of degrees of freedom
      * @return an instance of {@link ContinuousDistribution}
      */
-    public static ContinuousDistribution withParameters(IntegerTensor alpha) {
-        return new ChiSquared(alpha);
+    public static ContinuousDistribution withParameters(IntegerTensor degreesOfFreedom) {
+        return new ChiSquared(degreesOfFreedom);
     }
 
-    private ChiSquared(IntegerTensor alpha) {
-        this.alpha = alpha;
+    private ChiSquared(IntegerTensor degreesOfFreedom) {
+        this.degreesOfFreedom = degreesOfFreedom;
     }
 
     @Override
     public DoubleTensor sample(int[] shape, KeanuRandom random) {
-        return random.nextGamma(shape, DoubleTensor.TWO_SCALAR, alpha.toDouble().div(2));
+        return random.nextGamma(shape, DoubleTensor.TWO_SCALAR, degreesOfFreedom.toDouble().div(2));
     }
 
     @Override
     public DoubleTensor logProb(DoubleTensor x) {
-        DoubleTensor halfDof = alpha.toDouble().div(2);
-        DoubleTensor numerator = halfDof.minus(1).timesInPlace(x.log()).minusInPlace(x.div(2));
-        DoubleTensor denominator = halfDof.times(LOG_TWO).plusInPlace(halfDof.apply(Gamma::gamma).logInPlace());
+        DoubleTensor halfK = degreesOfFreedom.toDouble().div(2);
+        DoubleTensor numerator = halfK.minus(1).timesInPlace(x.log()).minusInPlace(x.div(2));
+        DoubleTensor denominator = halfK.times(LOG_TWO).plusInPlace(halfK.apply(Gamma::gamma).logInPlace());
         return numerator.minusInPlace(denominator);
     }
 
--- a/keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/ChiSquared.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/ChiSquared.java
@@ -11,26 +11,35 @@ import io.improbable.keanu.vertices.dbl.KeanuRandom;
 public class ChiSquared implements ContinuousDistribution {
 
     private static final double LOG_TWO = Math.log(2);
-    private final IntegerTensor k;
+    private final IntegerTensor alpha;
 
-    public static ContinuousDistribution withParameters(IntegerTensor k) {
-        return new ChiSquared(k);
+    /**
+     * <h3>Chi-Square Distribution</h3>
+     *
+     * @param alpha shape parameter (not to be confused with tensor shape); number of degrees of freedom
+     * @see "Computer Generation of Statistical Distributions
+     * by Richard Saucier,
+     * ARL-TR-2168 March 2000,
+     * 5.1.4 page 16"
+     */
+    public static ContinuousDistribution withParameters(IntegerTensor alpha) {
+        return new ChiSquared(alpha);
     }
 
-    private ChiSquared(IntegerTensor k) {
-        this.k = k;
+    private ChiSquared(IntegerTensor alpha) {
+        this.alpha = alpha;
     }
 
     @Override
     public DoubleTensor sample(int[] shape, KeanuRandom random) {
-        return random.nextGamma(shape, DoubleTensor.TWO_SCALAR, k.toDouble().div(2));
+        return random.nextGamma(shape, DoubleTensor.TWO_SCALAR, alpha.toDouble().div(2));
     }
 
     @Override
     public DoubleTensor logProb(DoubleTensor x) {
-        DoubleTensor halfK = k.toDouble().div(2);
-        DoubleTensor numerator = halfK.minus(1).timesInPlace(x.log()).minusInPlace(x.div(2));
-        DoubleTensor denominator = halfK.times(LOG_TWO).plusInPlace(halfK.apply(Gamma::gamma).logInPlace());
+        DoubleTensor halfDof = alpha.toDouble().div(2);
+        DoubleTensor numerator = halfDof.minus(1).timesInPlace(x.log()).minusInPlace(x.div(2));
+        DoubleTensor denominator = halfDof.times(LOG_TWO).plusInPlace(halfDof.apply(Gamma::gamma).logInPlace());
         return numerator.minusInPlace(denominator);
     }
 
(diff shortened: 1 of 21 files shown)


icantsayrural:
Realised that we should put the `@see` at the top so it shows up first in JavaDoc.

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/ChiSquared.java
```diff
@@ -8,34 +8,44 @@
...
+ * @see "Computer Generation of Statistical Distributions
+ * by Richard Saucier,
+ * ARL-TR-2168 March 2000,
+ * 5.1.4 page 16"
+ */
 public class ChiSquared implements ContinuousDistribution {
 
     private static final double LOG_TWO = Math.log(2);
-    private final IntegerTensor k;
+    private final IntegerTensor alpha;
```
migwellian: Wikipedia calls it `k`
Wolfram calls it `r`
Who calls it `alpha`?
We could call it `degreesOfFreedom` to be more descriptive

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/Exponential.java
```diff
@@ -8,36 +8,46 @@
...
+/**
+ * @see "Computer Generation of Statistical Distributions
+ * by Richard Saucier,
+ * ARL-TR-2168 March 2000,
+ * 5.1.8 page 20"
+ */
 public class Exponential implements ContinuousDistribution {
 
-    private final DoubleTensor lambda;
+    private final DoubleTensor rate;
```
migwellian: sgtm.

From wikipedia: 

> Here λ > 0 is the parameter of the distribution, often called the rate parameter

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/Gamma.java
```diff
@@ -18,123 +18,124 @@
...
+ * 5.1.11 page 23"
+ */
 public class Gamma implements ContinuousDistribution {
 
     private static final double M_E = 0.577215664901532860606512090082;
     private final DoubleTensor location;
-    private final DoubleTensor theta;
-    private final DoubleTensor k;
+    private final DoubleTensor scale;
+    private final DoubleTensor alpha;
```
migwellian: wikipedia calls `alpha`/`k` the "`shape`"

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/Gamma.java
```diff
@@ -18,123 +18,124 @@
...
+/**
+ * @see "Computer Generation of Statistical Distributions
+ * by Richard Saucier,
+ * ARL-TR-2168 March 2000,
+ * 5.1.11 page 23"
+ */
 public class Gamma implements ContinuousDistribution {
 
     private static final double M_E = 0.577215664901532860606512090082;
     private final DoubleTensor location;
```
migwellian: @gordoncaleb removed `location` from the `Exponential` distribution - do we want to remove it from here too?

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/Gamma.java
```diff
@@ -18,123 +18,124 @@
...
+        final DoubleTensor dLogPdscale = scale.times(alpha).plus(locationMinusX).divInPlace(scale.pow(2.)).unaryMinusInPlace();
+        final DoubleTensor dLogPdalpha = xMinusLocation.logInPlace().minusInPlace(scale.log()).minusInPlace(alpha.apply(org.apache.commons.math3.special.Gamma::digamma));
+        final DoubleTensor dLogPdx = alphaMinus1.div(xMinusLocation).minusInPlace(oneOverScale);
 
         return new Diffs()
         .put(A, dLogPdlocation)
-        .put(THETA, dLogPdtheta)
-        .put(K, dLogPdk)
+        .put(THETA, dLogPdscale)
+        .put(K, dLogPdalpha)
```
migwellian: if we go with `alpha` then you should also change `K` to `ALPHA`

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/Gaussian.java
```diff
@@ -9,51 +9,61 @@
...
 public class Gaussian implements ContinuousDistribution {
 
-    public static final double SQRT_2PI = Math.sqrt(Math.PI * 2);
-    public static final double LN_SQRT_2PI = Math.log(SQRT_2PI);
-    private final DoubleTensor mu;
-    private final DoubleTensor sigma;
+    private static final double SQRT_2PI = Math.sqrt(Math.PI * 2);
+    static final double LN_SQRT_2PI = Math.log(SQRT_2PI);
+    private final DoubleTensor location;
+    private final DoubleTensor scale;
```
migwellian: I don't like `scale` here because its ambiguous if that means the variance or the standard deviation.

I'd vote for keeping it as `mu` and `sigma` since they are a universal convention, but if we do change it then it ought to be to `mean` and `variance`.

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/InverseGamma.java
```diff
@@ -11,45 +11,55 @@
...
+ * @see "Computer Generation of Statistical Distributions
+ * by Richard Saucier,
+ * ARL-TR-2168 March 2000,
+ * 5.1.4 page 16"
+ */
 public class InverseGamma implements ContinuousDistribution {
 
     private final DoubleTensor alpha;
-    private final DoubleTensor beta;
+    private final DoubleTensor scale;
```
migwellian: I think `alpha` and `beta` are universally accepted, and it would be weird to change one and not the other.

Defer to a data scientist though.

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/LogNormal.java
```diff
@@ -10,52 +10,58 @@
...
+ * by Richard Saucier,
+ * ARL-TR-2168 March 2000,
+ * 5.1.15 page 28"
+ */
 public class LogNormal implements ContinuousDistribution {
 
-    private final DoubleTensor mu;
-    private final DoubleTensor sigma;
+    private final DoubleTensor location;
+    private final DoubleTensor scale;
```
migwellian: same comment as for the Gaussian (`scale` is ambiguous)

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/Logistic.java
```diff
@@ -9,34 +9,39 @@
...
+ * by Richard Saucier,
+ * ARL-TR-2168 March 2000,
+ * 5.1.14 page 27"
+ */
 public class Logistic implements ContinuousDistribution {
 
-    private final DoubleTensor mu;
-    private final DoubleTensor s;
+    private final DoubleTensor location;
+    private final DoubleTensor scale;
```
migwellian: makes sense since wikipedia and wolfram don't agree (`mu` and `s` vs `m` and `b`)

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/MultivariateGaussian.java
```diff
@@ -6,41 +6,48 @@
...
+     */
+    public static ContinuousDistribution withParameters(DoubleTensor location, DoubleTensor scale) {
+        return new MultivariateGaussian(location, scale);
     }
-    private MultivariateGaussian(DoubleTensor mu, DoubleTensor covariance) {
-        this.mu = mu;
-        this.covariance = covariance;
+    private MultivariateGaussian(DoubleTensor location, DoubleTensor scale) {
+        this.location = location;
+        this.scale = scale;
```
migwellian: `covariance` is actually more descriptive than `scale`

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/SmoothUniform.java
```diff
@@ -7,50 +7,16 @@
...
- * |  0    3Sw^2    2Sw   |   | h |   | 0 |
- * | -1    Sw^3     Sw^2  | * | A | = | 0 |
- * | Bw    Sw/4     2Sw/3 |   | B |   | 1 |
- * <p>
- * therefore:
- * <p>
- * h = 1 / (Sw + Bw)
- * A = -2 / (Sw^3 * (Sw + Bw))
- * B = 3 / (Sw^3 * Sw^2*Bw)
- */
```
migwellian: why are you deleting the comments from this file? They seem useful.

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/StudentT.java
```diff
@@ -14,68 +14,65 @@
...
+ * @see "Computer Generation of Statistical Distributions
+ * by Richard Saucier
+ * ARL-TR-2168 March 2000
+ * 5.1.23 page 36"
  */
 public class StudentT implements ContinuousDistribution {
 
     private static final double HALF_LOG_PI = log(PI) / 2;
-    private final IntegerTensor v;
+    private final IntegerTensor alpha;
```
migwellian: As with my comment for the ChiSquared, I'd prefer `degreesOfFreedom`

Fun fact: it's not actually meant to be "v", it's the Greek letter "nu" which looks like a "v"

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/Triangular.java
```diff
@@ -5,20 +5,31 @@
...
+ */
 public class Triangular implements ContinuousDistribution {
 
     private final DoubleTensor xMin;
     private final DoubleTensor xMax;
-    private final DoubleTensor c;
-
-    public static ContinuousDistribution withParameters(DoubleTensor xMin, DoubleTensor xMax, DoubleTensor c) {
-        return new Triangular(xMin, xMax, c);
+    private final DoubleTensor mode;
```
migwellian: sgtm

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Binomial.java
```diff
@@ -8,24 +8,34 @@
...
+ * by Richard Saucier
+ * ARL-TR-2168 March 2000
+ * 5.2.2 page 43"
+ */
 public class Binomial implements DiscreteDistribution {
 
-    private final DoubleTensor p;
-    private final IntegerTensor n;
+    private final DoubleTensor successProbability;
+    private final IntegerTensor numberOfTrials;
```
migwellian: sgtm

I'd also expect to see them in the other order (it's always `B(n,p)` as far as I know)

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Poisson.java
```diff
@@ -11,43 +11,46 @@
...
+ * @see "Computer Generation of Statistical Distributions
  * by Richard Saucier
  * ARL-TR-2168 March 2000
- * 5.2.8 page 49
+ * 5.2.8 page 49"
  */
 public class Poisson implements DiscreteDistribution {
 
-    private final DoubleTensor mu;
+    private final DoubleTensor rate;
```
migwellian: sgtm - especially since wikipedia calls it lambda and wolfram calls it nu!

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/SmoothUniform.java
icantsayrural: I removed it to add LaTeX doclet in another PR, but I don't think there is a clean way to do it so I'll add it back in.

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/ChiSquared.java
icantsayrural: The idea was to name shape parameters with greek alphabets to avoid clashing with tensor's `shape`.

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/Gamma.java
icantsayrural: I'm going to do this work after we decide on the parameter names.

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/Gaussian.java
icantsayrural: Good point! Maybe we will have to have special cases for Gaussian ...

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/InverseGamma.java
icantsayrural: Beta is a scale parameter, I wanted to make the distinction from the shape parameter alpha.

icantsayrural:
OK @migwellian, based on your comments I agree that fleshing out mean to location and variance/covariance/standard deviation to scale is overkill.

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/ChiSquared.java
migwellian: OK - it sounds like a bad idea to choose arbitrary greek letters.

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/Gamma.java
icantsayrural: #173

icantsayrural:
Beta, Gamma, Inverse Gamma and Log Normal have arbitrary shape parameters (that are not equivalent to special values like mean/standard deviation for Gaussian, degrees of freedom for StudentT and Chi-squared). Should we just call these `shape` or something else because it collides with `shape` for tensors.

PR #169, opened 2018-08-14
Increment gradle wrapper version

mhutti1:
Increment gradle wrapper so it is compatible with Java 10


mhutti1:
Having other issues with Java 10 that I will try solve first.

PR #171, opened 2018-08-15
Feature/multinomial

migwellian:
New distribution and vertex type "Multinomial": the count of sampling a categorical variable multiple times (think dice rolls).

I'm in the process of making it more performant (i.e. don't process every tensor entry individually) but will post that later in a separate PR. We need to do this for Binomial too so I'll tackle both together.


keanu-project/src/test/java/io/improbable/keanu/vertices/intgr/probabilistic/MultinomialVertexTest.java
```diff
@@ -0,0 +1,302 @@
...
+
+        for (int i = 0; i < samples.getShape()[0]; i++) {
+            System.out.println(i);
+            IntegerTensor sample = samples.slice(0, i);
+            Double probability = p.slice(0, i).scalar();
+            double mean = sample.toDouble().average();
+            double std = sample.toDouble().standardDeviation();
+
+            double epsilonForMean = 0.5;
+            double epsilonForVariance = 5.;
```
migwellian: NB my error tolerances are pretty large - is this OK?

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
```diff
@@ -0,0 +1,134 @@
...
+    private final IntegerTensor n;
+    private final DoubleTensor p;
+    private final int numCategories;
+
+    public static Multinomial withParameters(IntegerTensor n, DoubleTensor p) {
+        return new Multinomial(n, p);
+    }
+
+    /**
+     * https://en.wikipedia.org/wiki/Multinomial_distribution
```
icantsayrural: We can use the JavaDoc tag `@see` for references. (e.g. `@see <a href="https://en.wikipedia.org/wiki/Multinomial_distribution">Multinomial Distribution</a>` or if you don't `a` tag, just put the link in quotes)

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
```diff
@@ -0,0 +1,134 @@
...
+        }
+        this.n = n;
+        this.p = p;
+    }
+
+    @Override
+    public IntegerTensor sample(int[] shape, KeanuRandom random) {
+        TensorShapeValidation.checkTensorsMatchNonScalarShapeOrAreScalar(shape, n.getShape());
+
+        p.sum(0);
```
icantsayrural: I don't think this sum is in place

migwellian: yep sorry, that line just shouldn't be there at all

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
```diff
@@ -0,0 +1,134 @@
...
+            Double currentP = p[index++];
+            if (currentP == 0.) {
+                continue;
+            }
+            pCumulative += currentP;
+            if (pCumulative >= value) {
+                break;
+            }
+        }
+        return index-1;
```
icantsayrural: Is sampling -1 an expected behaviour?

migwellian: That can only happen when `p.length` is zero. I think we can reasonably assume that will never happen and it doesn't require a unit test.

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShapeValidation.java
```diff
@@ -71,20 +72,28 @@ public static void checkTensorsMatchNonScalarShapeOrAreScalar(int[] proposalShap
...
+        return checkAllShapesMatch(shapes.stream(), Optional.of(errorMessage));
     }
 
     public static int[] checkAllShapesMatch(Collection<int[]> shapes) {
-        return checkAllShapesMatch(shapes.stream());
+        return checkAllShapesMatch(shapes.stream(), Optional.empty());
     }
 
-    private static int[] checkAllShapesMatch(Stream<int[]> shapesStream) {
+    private static int[] checkAllShapesMatch(Stream<int[]> shapesStream, Optional<String> errorMessage) {
```
gordoncaleb: Do we need this Optional method arguments? Can we remove the logic branch upstream? (e.g. two methods where one takes a String and the other doesnt and the one that doesn't calls the ones that does with a default message)

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
```diff
@@ -51,7 +51,7 @@ static DoubleTensor eye(int n) {
         }
     }
 
-    static DoubleTensor zeros(int[] shape) {
+    static DoubleTensor zeros(int... shape) {
```
gordoncaleb: Nice! We should make sure this is consistent with other Tensor classes.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/ScalarDoubleTensor.java
```diff
@@ -779,4 +780,11 @@ public int hashCode() {
         result = 31 * result + Arrays.hashCode(shape);
         return result;
     }
+
+    @Override
+    public String toString() {
```
gordoncaleb: Got this coming in via a few PRs now.

keanu-project/src/main/java/io/improbable/keanu/tensor/intgr/IntegerTensor.java
```diff
@@ -52,7 +52,7 @@ static IntegerTensor eye(int n) {
         }
     }
 
-    static IntegerTensor zeros(int[] shape) {
+    static IntegerTensor zeros(int... shape) {
```
gordoncaleb: Consistent. Nice!

keanu-project/src/main/java/io/improbable/keanu/vertices/intgr/probabilistic/MultinomialVertex.java
```diff
@@ -0,0 +1,55 @@
...
+
+    private final DoubleVertex p;
+    private final IntegerVertex n;
+
+    public MultinomialVertex(int[] tensorShape, IntegerVertex n, DoubleVertex p) {
+        super(new ProbabilisticValueUpdater<>());
+
+        int[] tensorShapePrefixedByOne = ArrayUtils.insert(0, tensorShape, 1);
+        checkTensorsMatchNonScalarShapeOrAreScalar(tensorShape, n.getShape());
+        checkTensorsMatchNonScalarShapeOrAreScalar(tensorShapePrefixedByOne, p.slice(0, 0).getShape());
```
gordoncaleb: Is this just checking the length of p in dimension 1?

keanu-project/src/test/java/io/improbable/keanu/tensor/TensorMatchers.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeDiagnosingMatcher;
+
+
+public class TensorMatchers {
```
gordoncaleb: Very nice!

keanu-project/src/test/java/io/improbable/keanu/vertices/intgr/probabilistic/MultinomialVertexTest.java
```diff
@@ -0,0 +1,302 @@
...
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.operators.multiple.ConcatenationVertex;
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.operators.unary.ReshapeVertex;
+import io.improbable.keanu.vertices.intgr.IntegerVertex;
+
+public class MultinomialVertexTest {
+
+    @Test(expected = IllegalArgumentException.class)
+    public void itThrowsIfTheProbabilitiesDontSumToOne() {
+        IntegerTensor n = IntegerTensor.scalar(100);
+        DoubleTensor p = DoubleTensor.create(0., 0., 0.99, 0.).transpose();
```
gordoncaleb: Is the transpose just to avoid having to specify a column shape?

keanu-project/src/test/java/io/improbable/keanu/vertices/intgr/probabilistic/MultinomialVertexTest.java
```diff
@@ -0,0 +1,302 @@
...
+                .50, .3,
+
+                .8, .1,
+                .25, .5
+            },
+            3, 2, 2));
+        //
+        MultinomialVertex multinomial = new MultinomialVertex(n, p);
+        IntegerTensor sample = multinomial.sample(KeanuRandom.getDefaultRandom());
+        assertThat(sample, hasShape(3, 2, 2));
```
gordoncaleb: Nice matcher!

keanu-project/src/test/java/io/improbable/keanu/vertices/intgr/probabilistic/MultinomialVertexTest.java
gordoncaleb: Does it approach 0 as you take more samples? If so then it's fine.

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
```diff
@@ -0,0 +1,133 @@
...
+        this.n = n;
+        this.p = p;
+    }
+
+    @Override
+    public IntegerTensor sample(int[] shape, KeanuRandom random) {
+        TensorShapeValidation.checkTensorsMatchNonScalarShapeOrAreScalar(shape, n.getShape());
+
+        Tensor.FlattenedView<Integer> nFlattened = n.getFlattenedView();
+        List<Tensor.FlattenedView<Double>> pFlattened = Lists.newArrayList();
```
GeorgeNash: I like what you're doing here and I think it's something people will want to do a lot. Can we pull this into a method on `DoubleTensor`:

```
    @Override
    public List<DoubleTensor> sliceAlongDimension(int dimension, int indexStart, int indexEnd) {
        List<DoubleTensor> slicedTensors = new ArrayList<>();

        for (int i = indexStart; i < indexEnd; i++) {
            slicedTensors.add(slice(dimension, i));
        }

        return slicedTensors;
    }
```

Then you can do:

```
 Tensor.FlattenedView<Integer> nFlattened = n.getFlattenedView();
        List<DoubleTensor> sliced = p.sliceAlongDimension(0, 0, numCategories);

        int length = ArrayUtil.prod(shape);
        int[] samples = new int[0];

        for (int i = 0; i < length; i++) {
            final int j = i;
            List<Double> ps = sliced.stream().map(p -> p.getFlattenedView().getOrScalar(j)).collect(Collectors.toList());
            int[] sample = sample(nFlattened.getOrScalar(i), random, ps.toArray(new Double[0]));
            samples = ArrayUtils.addAll(samples, sample);
        }
```

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
```diff
@@ -0,0 +1,133 @@
...
+     */
+    private Multinomial(IntegerTensor n, DoubleTensor p) {
+        Preconditions.checkArgument(
+            p.sum(0).elementwiseEquals(DoubleTensor.ones(n.getShape())).allTrue(),
+            "Probabilities must sum to one"
+        );
+
+        numCategories = p.getShape()[0];
+        for (int category = 0; category < numCategories; category++) {
+            TensorShapeValidation.checkAllShapesMatch(n.getShape(), p.slice(0, category).getShape());
```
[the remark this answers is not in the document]
migwellian: Yes. So if I understand you correctly, you're saying I should get rid of the loop over `numCategories` and just do this one-liner:

`TensorShapeValidation.checkAllShapesMatch(n.getShape(), p.slice(0, 0).getShape());`

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    @Override
+    public IntegerTensor sample(int[] shape, KeanuRandom random) {
+        TensorShapeValidation.checkTensorsMatchNonScalarShapeOrAreScalar(shape, n.getShape());
+
+        Tensor.FlattenedView<Integer> nFlattened = n.getFlattenedView();
+        List<Tensor.FlattenedView<Double>> pFlattened = Lists.newArrayList();
+
+        for (int category = 0; category < numCategories; category++) {
```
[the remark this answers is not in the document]
migwellian: Agreed - but I found it easier to reason about a `List` of flattened views, not just one big flattened view.

In any case I have plans to replace this with a totally tensorified version (in a future PR).

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
```diff
@@ -0,0 +1,133 @@
...
+            DoubleTensor p_i = p.slice(0, category);
+            pFlattened.add(p_i.getFlattenedView());
+        }
+
+        int length = ArrayUtil.prod(shape);
+        int[] samples = new int[0];
+
+        for (int i = 0; i < length; i++) {
+            final int j = i;
+            List<Double> ps = pFlattened.stream().map(p -> p.getOrScalar(j)).collect(Collectors.toList());
```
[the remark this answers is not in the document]
migwellian: I don't have strong feelings either way, however: suppose we later switch to using `parallelStream`?

Again it's kind of moot since I have plans to replace this code with a tensorified version.

keanu-project/src/test/java/io/improbable/keanu/vertices/intgr/probabilistic/MultinomialVertexTest.java
migwellian: Yes.

I ran it for `N` = 1 million. It took 18 minutes but it gave me an accuracy of about 0.01 for mean and 0.4 for variance.

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
```diff
@@ -0,0 +1,133 @@
...
+     * @see <a href="https://en.wikipedia.org/wiki/Multinomial_distribution">Multinomial Distribution</a>
+     * Generalisation of the Binomial distribution to variables with more than 2 possible values
+     *
+     * @param n The number of draws from the variable
+     * @param p The probability of observing each of the k values (which sum to 1)
+     *          p is a Tensor whose first dimension must be of size k
+     */
+    private Multinomial(IntegerTensor n, DoubleTensor p) {
+        Preconditions.checkArgument(
+            p.sum(0).elementwiseEquals(DoubleTensor.ones(n.getShape())).allTrue(),
```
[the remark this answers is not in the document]
migwellian: Good idea. I've changed the condition to:
```
    Math.abs(p.sum() - ArrayUtil.prod(n.getShape())) < 1e-8
```

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
```diff
@@ -0,0 +1,133 @@
...
+            List<Double> ps = pFlattened.stream().map(p -> p.getOrScalar(j)).collect(Collectors.toList());
+            int[] sample = sample(nFlattened.getOrScalar(i), random, ps.toArray(new Double[0]));
+            samples = ArrayUtils.addAll(samples, sample);
+        }
+        int[] outputShape = shape;
+        if (shape[0] == 1) {
+            outputShape = ArrayUtils.remove(outputShape, 0);
+        }
+        outputShape = ArrayUtils.add(outputShape, numCategories);
+        return IntegerTensor.create(samples, outputShape).transpose();
```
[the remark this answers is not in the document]
migwellian: I realise now that there was a bug here, which I've just fixed.

Suppose there are k categories. The transpose was necessary because I'm constructing the flat array by repeatedly concatenating k values (which sum to n). That yields me a tensor of shape e.g. [a, b, c, d, k].

The bug is that by taking the transpose I get a tensor of shape [k, d, c, b, a] whereas what I actually need is shape [k, a, b, c, d]. So instead of taking a transpose, I actually need to slice in the last dimension, then concatenate again.

I've written a new method to do this:

    private IntegerTensor constructSampleTensor(int[] shape, int[] samples) {
        int[] outputShape = shape;
        if (shape[0] == 1) {
            outputShape = ArrayUtils.remove(outputShape, 0);
        }
        IntegerTensor abkTensor = IntegerTensor.create(samples, ArrayUtils.add(outputShape, numCategories));
        int[] kabArray = new int[] {};
        for (int category = 0; category < numCategories; category++) {
            IntegerTensor abTensor = abkTensor.slice(outputShape.length, category);
            kabArray = ArrayUtils.addAll(kabArray, abTensor.asFlatIntegerArray());
        }

        return IntegerTensor.create(kabArray, ArrayUtils.insert(0, outputShape, numCategories));
    }

keanu-project/src/test/java/io/improbable/keanu/vertices/intgr/probabilistic/MultinomialVertexTest.java
migwellian: Yes, it's the neatest way I know to declare a column vector. Neater than this:
```
DoubleTensor.create(new double[] {0., 0., 0.99, 0.}, new int[] {4, 1})
```

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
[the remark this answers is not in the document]
migwellian: Yes but we'll need https://github.com/improbable-research/keanu/pull/139 for that

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
```diff
@@ -0,0 +1,133 @@
...
+
+    }
+
+    private static int[] sample(int n, KeanuRandom random, Double... p) {
```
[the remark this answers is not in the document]
migwellian: Good shout, although I think `drawNTimes` is a more accurate name than `takeDrawsFromCategory`

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
migwellian: Done - made `sliceAlongDimension` a default method on the `DoubleTensor` interface.

keanu-project/src/main/java/io/improbable/keanu/vertices/intgr/probabilistic/MultinomialVertex.java
migwellian: No, it's checking the length of p in all the dimensions except 0.

I've changed it to this, which is clearer:

        int[] pShapeExcludingFirstDimension = ArrayUtils.remove(p.getShape(), 0);
        checkTensorsMatchNonScalarShapeOrAreScalar(tensorShape, pShapeExcludingFirstDimension);

migwellian:
I've pushed changes in response to all your comments - hopefully you've got something that can be merged into develop in my absence :-)

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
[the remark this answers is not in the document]
GeorgeNash: sweet - works for me

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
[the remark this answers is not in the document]
gordoncaleb:
```
0.3 0.3
0.3 0.3
0.3 0.5
```
Would sum to 2 but not the required 3 right?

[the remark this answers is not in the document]
migwellian: No, it is expected to sum to 2 but the stronger condition (which this check fails to catch) is that each column must sum to 1.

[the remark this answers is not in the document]
gordoncaleb: Oh right. I see your check for each column now.

Issue #163, opened 2018-08-13
Allow MetropolisHastings samples to be dropped as they are generated


mhutti1:
**Is your feature request related to a problem? Please describe.**
I am sampling each layer of a tree consecutively and then downsampling to a every complete iteration of the tree. The issue I am facing is that to take a decent number of samples I first need to store each vertex and every sample for it before downsampling. This uses a lot of memory and once I hit about 80GB memory pressure rises and the system becomes unusable.

**Describe the solution you'd like**
I would like to be able to downsample as we go. E.g Take around 140 samples (one sample for each tree layer), then drop all but the last one and repeat for the desired sample count. One option would be to add downsampling as an optional parameter to the getPosteriorSamples method.

**Describe alternatives you've considered**
Other possible solutions would be to allow users to create their own drop selector and pass it in, or to use streams to filter in real time (although this might be more complex).

christophernorth:
Fixed by PR #172

Issue #164, opened 2018-08-13
Set default values in the MetropolisHastings builder


mhutti1:
**Is your feature request related to a problem? Please describe.**
When creating a new instance of MetropolisHastings via ```MetropolisHastings.builder()```. All of the options must be set by the user or else they will be null. This means that for example you must call ```.random(new KeanuRandom())``` or else crash with a null pointer when attempting to sample.

```
Exception in thread "main" java.lang.NullPointerException
	at io.improbable.keanu.distributions.continuous.Triangular.sample(Triangular.java:26)
	at io.improbable.keanu.distributions.continuous.Triangular.sample(Triangular.java:8)
	at io.improbable.keanu.vertices.dbl.probabilistic.TriangularVertex.sample(TriangularVertex.java:117)
	at io.improbable.keanu.vertices.dbl.probabilistic.TriangularVertex.sample(TriangularVertex.java:14)
	at io.improbable.irsim.TreeProposal.setFor(TreeProposal.java:26)
	at io.improbable.irsim.TreeProposal.getProposal(TreeProposal.java:15)
	at io.improbable.keanu.algorithms.mcmc.MetropolisHastingsStep.step(MetropolisHastingsStep.java:74)
	at io.improbable.keanu.algorithms.mcmc.MetropolisHastingsStep.step(MetropolisHastingsStep.java:53)
	at io.improbable.keanu.algorithms.mcmc.MetropolisHastings.getPosteriorSamples(MetropolisHastings.java:71)
	at io.improbable.irsim.RoadBinaryParser.complete(RoadBinaryParser.java:137)
	at org.openstreetmap.osmosis.osmbinary.file.BlockInputStream.process(BlockInputStream.java:37)
	at io.improbable.irsim.IRSim.main(IRSim.java:29)
```

**Describe the solution you'd like**
Default values should be used so things such as a KeanuRandom() do not have to be passed in if not required.

**Describe alternatives you've considered**
An alternative solution might be to make the default configs settings mutable.

christophernorth:
Fixed by #172

PR #172, opened 2018-08-15
Configurable sampling and efficient dropping of samples from MH

gordoncaleb:
This PR addresses some of the issues we have seen internally:

1. We shouldn't store samples from Metropolis Hastings that we know will be dropped.
2. All posterior sampling algorithms should have sensible defaults
3. The settings on those algorithms should be settable post construction


keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
```diff
@@ -64,19 +94,63 @@ public NetworkSamples getPosteriorSamples(final BayesianNetwork bayesianNetwork,
...
+        @Override
+        public NetworkState sample() {
+            return new SimpleNetworkState(takeSample(verticesToSampleFrom));
+        }
+    }
+
+    private static Map<Long, ?> takeSample(List<? extends Vertex> fromVertices) {
+        Map<Long, Object> sample = new HashMap<>();
+        for (Vertex v : fromVertices) {
+            sample.put(v.getId(), v.getValue());
```
christophernorth: Could we use Collectors.toMap() here?

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/NetworkSamplesGenerator.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Map;
+import java.util.stream.Stream;
+
+import io.improbable.keanu.algorithms.NetworkSamples;
+import io.improbable.keanu.network.NetworkState;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.experimental.Accessors;
+
+@Accessors(fluent = true)
```
christophernorth: Do we definitely want to use "experimental" features from Lombok?

christophernorth:
Fixes #163 #164

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
```diff
@@ -22,36 +26,53 @@
...
+
+    public static Hamiltonian withDefaultConfig() {
+        return withDefaultConfig(KeanuRandom.getDefaultRandom());
+    }
+
+    public static Hamiltonian withDefaultConfig(KeanuRandom random) {
+        return Hamiltonian.builder()
+            .random(random)
+            .leapFrogCount(DEFAULT_LEAP_FROG_COUNT)
+            .stepSize(DEFAULT_STEP_SIZE)
```
christophernorth: Given we have a default value specified for these - do we need to specify them as part of the builder here?  Seems redundant (repeated comment as my previous one seems to have disappeared for some reason)

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
```diff
@@ -22,36 +26,50 @@
...
-    private Hamiltonian() {
+    public static Hamiltonian withDefaultConfig() {
+        return withDefaultConfig(KeanuRandom.getDefaultRandom());
     }
 
+    public static Hamiltonian withDefaultConfig(KeanuRandom random) {
+        return Hamiltonian.builder()
+            .random(random)
+            .leapFrogCount(20)
+            .stepSize(0.1)
```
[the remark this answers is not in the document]
gordoncaleb: Not needed but I figured if someone decompiled the class they'd get the lombok generated stuff and it might not be clear what the defaults are unless this method specified them. It should at least be the same defaults as the builder uses, which I've updated it to make sure happens.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
gordoncaleb: Just for clarity if someone looks a the decompiled class. I've moved them to static finals so at least it's ensured they're the same.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/NetworkSamplesGenerator.java
gordoncaleb: The only risk is that they remove the feature or if the feature is fundamentally broken. Our test ensure it's not broken and it looks like they think it'll be moved to the "stable" category soon. `Current status: positive - Currently we feel this feature may move out of experimental status with no or minor changes soon.`.

christophernorth: Yeah - I'd spotted that.  Just wanted to make sure we were making the decision with our eyes open though.  Thanks for thinking about it.

PR #173, opened 2018-08-15
Remove location parameter from Gamma

icantsayrural:
Similar to Exponential, just assume location is 0.


keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/GammaVertex.java
```diff
@@ -22,98 +22,74 @@
 
 public class GammaVertex extends DoubleVertex implements ProbabilisticDouble {
 
-    private final DoubleVertex location;
     private final DoubleVertex theta;
     private final DoubleVertex k;
 
     /**
-     * One location, theta or k or all three driving an arbitrarily shaped tensor of Gamma
+     * Theta or k or all two driving an arbitrarily shaped tensor of Gamma
```
gordoncaleb: All two = both?

icantsayrural: 😆

PR #174, opened 2018-08-16
Add a temporary fix for [plus][minus][times][div]InPlace from a scala…

icantsayrural:
…r Nd4jDoubleTensor and a higher dimension Nd4jDoubleTensor

Context: some `dLogProb` tests in `GammaVertexTest` were failing because in place operations between a scalar and tensor weren't behaving as we expected (i.e. returning a scalar instead of a tensor).

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
@@ -510,6 +510,9 @@ public class Nd4jDoubleTensor implements DoubleTensor {
     public DoubleTensor minusInPlace(DoubleTensor that) {
         if (that.isScalar()) {
             tensor.subi(that.scalar());
+        } else if (this.isScalar()) {
+            // Temporary fix: Scalar can't be subtracted by tensor inplace
+            return this.minus(that);
         } else {
             INDArrayShim.subi(tensor, unsafeGetNd4J(that), tensor);
         }
@@ -520,6 +523,9 @@ public class Nd4jDoubleTensor implements DoubleTensor {
     public DoubleTensor plusInPlace(DoubleTensor that) {
         if (that.isScalar()) {
             tensor.addi(that.scalar());
+        } else if (this.isScalar()) {
+             // Temporary fix: Scalar can't be added by tensor inplace
+            return this.plus(that);
         } else {
             INDArrayShim.addi(tensor, unsafeGetNd4J(that), tensor);
         }
@@ -530,6 +536,9 @@ public class Nd4jDoubleTensor implements DoubleTensor {
     public DoubleTensor timesInPlace(DoubleTensor that) {
         if (that.isScalar()) {
             tensor.muli(that.scalar());
+        } else if (this.isScalar()) {
+             // Temporary fix: Scalar can't be multiplied by tensor inplace
+            return this.times(that);
         } else {
             INDArrayShim.muli(tensor, unsafeGetNd4J(that), tensor);
         }
@@ -540,6 +549,9 @@ public class Nd4jDoubleTensor implements DoubleTensor {
     public DoubleTensor divInPlace(DoubleTensor that) {
         if (that.isScalar()) {
             tensor.divi(that.scalar());
+        } else if (this.isScalar()) {
+             // Temporary fix: Scalar can't be divided by tensor inplace
+            return this.div(that);
         } else {
             INDArrayShim.divi(tensor, unsafeGetNd4J(that), tensor);
         }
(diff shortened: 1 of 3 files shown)


keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -510,6 +510,9 @@ public DoubleTensor expInPlace() {
     public DoubleTensor minusInPlace(DoubleTensor that) {
         if (that.isScalar()) {
             tensor.subi(that.scalar());
+        } else if (this.isScalar()) {
```
christophernorth: Is this just a problem with Scalars?  Or a wider problem with any tensor operation between Tensors of order N, M where N < M?

keanu-project/src/test/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensorTest.java
```diff
@@ -312,4 +312,35 @@ public void canCalculateProductOfVector() {
         assertTrue(vectorA.isVector() && vectorB.isVector());
     }
 
+    @Test
+    public void scalarMinusInPlaceTensorBehavesSameAsMinus() {
```
christophernorth: Similar comment to that above really - what if we had say

{1 , 2, 3} - {{1, 2, 3},
                   {4, 5, 6}}

Wouldn't this need reshaping too?

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
christophernorth: Scratch that - I chatted to Caleb quickly and apparently ND4J will handle this for us under the covers.  Would be great to add a test case to ensure that works just fine though!

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -510,6 +510,9 @@ public DoubleTensor expInPlace() {
...
+            // Temporary fix: Scalar can't be subtracted by tensor inplace
```
gordoncaleb: Can we drop the comment about it being a "Temporary fix"? Can we document the method as having this behaviour and that is the fix. Later when we change the behaviour we'll update the method docs appropriately.

keanu-project/src/test/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensorTest.java
```diff
@@ -312,4 +312,35 @@ public void canCalculateProductOfVector() {
...
+
+        assertArrayEquals(scalar.times(tensor).asFlatDoubleArray(), scalar.timesInPlace(tensor).asFlatDoubleArray(), 1e-6);
+    }
+
+    @Test
+    public void scalarDivInPlaceTensorBehavesSameAsDiv() {
+        DoubleTensor scalar = DoubleTensor.scalar(1);
+        DoubleTensor tensor = DoubleTensor.create(2, new int[] {1, 4});
+
+        assertArrayEquals(scalar.div(tensor).asFlatDoubleArray(), scalar.divInPlace(tensor).asFlatDoubleArray(), 1e-6);
```
christophernorth: Given the numbers in use here, we should be fine not using a delta - it would be quite bad if a single op on a small whole number caused 1e-6 error - especially if we're using Double precision values.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -510,6 +510,9 @@ public DoubleTensor expInPlace() {
     public DoubleTensor minusInPlace(DoubleTensor that) {
```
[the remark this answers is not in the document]
icantsayrural: a) Done
b) 👍

keanu-project/src/test/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensorTest.java
icantsayrural: The delta is required for comparisons on doubles.

christophernorth: Aha - thanks, I see this in the API now (the version without delta is deprecated, but does exist).  I'm still a bit nervous about such a "large" delta.  I guess it's not the end of the world though!

PR #175, opened 2018-08-16
 Added HalfCauchyVertex: a subclass of CauchyVertex

robertperrotta:
Food for thought, subclassing CauchyVertex might be the easiest way to implement a HalfCauchyVertex.

I may be oversimplifying things, but I believe a half-Cauchy distribution is just a Cauchy distribution with location zero with its left half dropped and the rest scaled up by two. (I think logPdf doesn't care about scaling factors (biases in log space) so we don't even need to do that step.) And the gradient is twice as steep as it would be for the CauchyVertex. This gradient calculation implementation doesn't care if the value is less than zero, so it is not really an accurate gradient for those values, but since they have probability density zero anyway does it matter?

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/HalfCauchyVertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/HalfCauchyVertex.java
@@ -10,7 +10,7 @@ public class HalfCauchyVertex extends CauchyVertex {
     private static final double LOG_TWO = Math.log(2);
 
     /**
-     * One scale that matches a proposed tensor shape of HalfCauchy (Cauchy with location = 0 and x >= 0)
+     * One scale that matches a proposed tensor shape of HalfCauchy (Cauchy with location = 0 and non-negative x)
      * <p>
      * If provided parameter is scalar then the proposed shape determines the shape
      *
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/HalfCauchyVertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/HalfCauchyVertex.java
@@ -1,0 +1,51 @@
+package io.improbable.keanu.vertices.dbl.probabilistic;
+
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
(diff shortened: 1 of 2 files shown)


chris-major-improbable:
The code looks good to me, and while I follow the maths, I wouldn't claim to spot any errors (though it aligns with what I've read).

robertperrotta:
Definitely! I wanted to get a green light on the concept before putting in the time to write tests but I would gladly adopt the CauchyVertex tests for this. I'll take a crack at that next week when I'm in London.

robertperrotta:
Three corrections to my approach:

1. For non-scalar values, the bias of log(2) needs to be added *per element* in value.
2. The gradient doesn't need to change at all. It is the gradient of the *log* of the pdf, which has only changed by a constant offset.
3. The sample method needs to be overloaded, too (by taking the absolute value of the result of super.sample).

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/HalfCauchyVertex.java
```diff
@@ -0,0 +1,48 @@
...
+    @Override
+    public double logProb(DoubleTensor value) {
+        if (value.greaterThanOrEqual(0.0).allTrue()) {
+            return super.logProb(value) + LOG_TWO * value.getLength();
+        }
+        return Double.NEGATIVE_INFINITY;
+    }
+
+    @Override
+    public DoubleTensor sample(KeanuRandom random) {
```
icantsayrural: Looks great!
nit: should put `sample` above `logProb` for consistency.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/HalfCauchyVertex.java
```diff
@@ -0,0 +1,48 @@
...
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+import io.improbable.keanu.vertices.dbl.DoubleVertex;
+import io.improbable.keanu.vertices.dbl.KeanuRandom;
+
+public class HalfCauchyVertex extends CauchyVertex {
+
+    private static final double LOG_TWO = Math.log(2);
+
+    /**
+     * One scale that matches a proposed tensor shape of Cauchy
```
GeorgeNash: `HalfCauchy`

PR #178, opened 2018-08-17
Concatenation forward vertex fix

GeorgeNash:
Fixes forward mode auto-diff for the concat vertex

- Fills zero's where appropriate so that the auto diff of the concat vertex is correct


keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/DualNumber.java
```diff
@@ -32,6 +31,70 @@ public static DualNumber ifThenElse(BooleanTensor predicate, DualNumber thn, Dua
         }
     }
 
+    public static DualNumber concat(Map<Vertex, DualNumber> dualNumbers,
```
gordoncaleb: Looks good for now. We should add a task in the backlog to revisit this, but we should fix this now.

PR #187, opened 2018-08-22
Fix broadcast ops

GeorgeNash:
- Fixes a bug in the broadcast operations when operating on two tensors. Eg, multiplying a 2x2 by a 2x2x2x2 produced incorrect numbers and was different if you were doing A x B or B x A. The behaviour is now representative of numpy.
- Fixes a bug in TensorShape prepend with 1's
- Due to the behaviour of ND4J it's impossible to perform a broadcast in place on a tensor that has to increase in size. You could do an `inPlace` if multiplying a 2x2x2x2 by a 2x2, but not a 2x2 by a 2x2x2x2. To avoid confusing behaviour I've made it so that `inPlace` on a broadcast returns a duplicate. `There's no in place unless the tensors match shape`.
- Fixed `DoubleIfVertex` to use the new broadcast. It was using the previously broken broadcast to apply a mask in a pretty cheeky way 

It's impossible in `ND4J` to currently do a broadcast op on a small tensor against a larger tensor. To get around this I've done some shenanigans with swapping to ensure the larger tensor is always the primary

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
@@ -1,6 +1,7 @@
 package io.improbable.keanu.tensor;
 
 import com.google.common.primitives.Ints;
+import org.nd4j.linalg.api.buffer.DataBuffer;
 import org.nd4j.linalg.api.ndarray.INDArray;
 import org.nd4j.linalg.api.ops.BroadcastOp;
 import org.nd4j.linalg.api.ops.impl.broadcast.BroadcastAddOp;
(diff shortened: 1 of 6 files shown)


keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -24,76 +24,116 @@
  */
 public class INDArrayShim {
 
-    public static void muli(INDArray left, INDArray right, INDArray result) {
```
christophernorth: Good change - this looks like it brings us more in line with the Base API anyway.

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -24,76 +24,116 @@
...
+            return broadcastMultiply(left, right);
         }
     }
 
-    private static void broadcastMultiply(INDArray a, INDArray b, INDArray result) {
-        int[] broadcastDimensions = Shape.getBroadcastDimensions(a.shape(), b.shape());
-        execBroadcast(a, b,
-            new BroadcastMulOp(a, b, result, broadcastDimensions)
+    private static INDArray broadcastMultiply(INDArray a, INDArray b) {
+        int[] broadcastDimensions = getBroadcastDimensions(a.shape(), b.shape());
```
christophernorth: Should we do the 'if' first?  This seems like dead code if we're just going to make the recursive call.

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -24,76 +24,116 @@
...
     }
 
-    private static void broadcastDivide(INDArray a, INDArray b, INDArray result) {
-        int[] broadcastDimensions = Shape.getBroadcastDimensions(a.shape(), b.shape());
-        execBroadcast(a, b,
-            new BroadcastDivOp(a, b, result, broadcastDimensions)
+    private static INDArray broadcastDivide(INDArray a, INDArray b) {
+        int[] broadcastDimensions = getBroadcastDimensions(a.shape(), b.shape());
+        if (a.shape().length < b.shape().length) {
+            return broadcastMultiply(b, a.rdiv(1.0));
```
christophernorth: Think you mean divide here?

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -24,76 +24,116 @@
...
     }
 
-    private static void broadcastMinus(INDArray a, INDArray b, INDArray result) {
+    private static INDArray broadcastMinus(INDArray a, INDArray b) {
         int[] broadcastDimensions = Shape.getBroadcastDimensions(a.shape(), b.shape());
-        execBroadcast(a, b,
-            new BroadcastSubOp(a, b, result, broadcastDimensions)
-        );
+        if (a.shape().length < b.shape().length) {
+            return broadcastPlus(a.neg(), b);
```
christophernorth: broadcastMinus here?

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -538,12 +538,12 @@ public DoubleTensor minusInPlace(DoubleTensor that) {
         } else if (this.isScalar()) {
             return this.minus(that);
         } else {
-            INDArrayShim.subi(tensor, unsafeGetNd4J(that), tensor);
+            return new Nd4jDoubleTensor(INDArrayShim.subi(tensor, unsafeGetNd4J(that)));
```
christophernorth: Do we *always* want to return the new Tensor?  This seems to undo the point of the inPlace stuff - can we do a check that the value we get back is the same one as before and return a new Tensor only if the array objects differ?

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
christophernorth: Actually, sorry - I completely mis-read this - please ignore.

christophernorth: although perhaps we should just call the Divide function anyway to make it easier to read?

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
christophernorth: As for the previous comment.

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
```diff
@@ -1,6 +1,8 @@
 package io.improbable.keanu.tensor;
 
+import java.util.ArrayList;
```
icantsayrural: These imports look like they aren't being used?

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -24,76 +24,116 @@
...
+        } else {
+            shapeB = TensorShape.shapeToDesiredRankByPrependingOnes(shapeB, shapeA.length);
+        }
+
         List<Integer> along = new ArrayList<>();
-        for (int i = 0; i < minRank; i++) {
-            if (shapeA[i] == shapeB[i]) {
+
+        for (int i = maxRank - 1; i >= 0; i--) {
+            if (shapeA[i] == shapeB[i]){
```
christophernorth: Two functions that are very similar apart from a single operation - could we collapse the code down a bit?

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
GeorgeNash: yes great shout

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
GeorgeNash: Nice, I've implemented a check to see if the tensor returned equals the tensor we're operating on. If they're the same then we just return `this`

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
GeorgeNash: nice

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -24,76 +24,116 @@
...
+            return left.subi(right);
         } else {
-            broadcastMinus(left, right, result);
+            return broadcastMinus(left, right);
         }
     }
 
-    private static void broadcastMinus(INDArray a, INDArray b, INDArray result) {
+    private static INDArray broadcastMinus(INDArray a, INDArray b) {
         int[] broadcastDimensions = Shape.getBroadcastDimensions(a.shape(), b.shape());
```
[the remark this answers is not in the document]
GeorgeNash: Yep good spot, thanks

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
GeorgeNash: I'd normally agree - but the small difference between the functions is so crucial I'd rather it was overly explicit what was going on. I tried a boolean flag whether it's `along` or `matching` but it just makes something very simple a bit confusing

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -21,79 +21,128 @@
  * Until this is fixed in the ND4J codebase, these methods can be
  * used to work around the issue. The need for this should be
  * reevaluated each time the ND4J dependency is updated.
+ *
+ * To work around another issue in ND4J where you cannot broadcast
+ * a higher rank tensor onto a lower rank tensor, the shim broadcast operations
+ * ensure the higher rank tensor is always being operated on. In the case of
+ * subtract and minus, this requires a small change in the logic, as A - B != B - A and
+ * A / B != B / A.
```
christophernorth: Awesome stuff :)

keanu-project/src/main/java/io/improbable/keanu/tensor/intgr/Nd4jIntegerTensor.java
```diff
@@ -368,7 +377,8 @@ public IntegerTensor divInPlace(IntegerTensor that) {
         if (that.isScalar()) {
             tensor.divi(that.scalar());
         } else {
-            INDArrayShim.divi(tensor, unsafeGetNd4J(that), tensor);
+            INDArray result = INDArrayShim.divi(tensor, unsafeGetNd4J(that));
+            return new Nd4jIntegerTensor(INDArrayExtensions.castToInteger(result, false));
```
christophernorth: Do we need the "if result != tensor" again here?

keanu-project/src/test/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensorTest.java
```diff
@@ -21,6 +22,7 @@
     public void setup() {
         matrixA = Nd4jDoubleTensor.create(new double[]{1, 2, 3, 4}, new int[]{2, 2});
         matrixB = Nd4jDoubleTensor.create(new double[]{1, 2, 3, 4}, new int[]{2, 2});
+        matrixC = Nd4jDoubleTensor.create(new double[]{1, 2, 3, 4, 1, 2, 3, 4, 1, 2, 3, 4, 1, 2, 3, 4}, new int[]{4, 4});
```
christophernorth: Not sure if you're using matrixC?  Perhaps I'm just missing it?

keanu-project/src/test/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensorTest.java
```diff
@@ -456,4 +598,135 @@ public void doesCompareGreaterThanOrEqualScalarTensor() {
...
+            4, 3, 2, 1, 4, 3, 2, 1,
+            4, 3, 2, 1, 4, 3, 2, 1,
+            4, 3, 2, 1, 4, 3, 2, 1,
+            4, 3, 2, 1, 4, 3, 2, 1
+        }, new int[]{4, 2, 2, 2});
+
+        assertMinusOperationEquals(matrix, rank4, expected);
+        assertMinusInPlaceOperationEquals(matrix, rank4, expected);
+    }
+
```
christophernorth: Should we have a negative test case that checks we now don't allow things that *Shouldn't* be supported?

keanu-project/src/main/java/io/improbable/keanu/tensor/intgr/Nd4jIntegerTensor.java
GeorgeNash: good catch!!!

Issue #190, opened 2018-08-22
NetworkSamples streaming never updates network state


robertperrotta:
`networkSamples.generate(N)` does not produce the same output as `networkSamples.stream().limit(N).toList()` For `MetropolisHastings`.

I believe the problem is that in `MetropolisHastings.java` lines 317 - 146:

```java
        @Override
        public void sample(Map<Long, List<?>> samplesByVertex) {
            step();
            takeSamples(samplesByVertex, verticesToSampleFrom);
        }

        @Override
        public NetworkState sample() {
            return new SimpleNetworkState(takeSample(verticesToSampleFrom));
        }
```

the first method (called by `generate(N)`) calls `step()` before `takeSamples` but the second doesn't.

christophernorth:
Good spot - we'll work on getting this sorted.

christophernorth:
Will make it out as part of the next release - look out for 0.0.13!

PR #199, opened 2018-08-23
Fix for Issue 190

christophernorth:
Fix Issue with Metropolis Hastings algorithm not generating samples correctly (we always returned the same value) if Downsampling not used when streaming.

Also includes a small fix to the Metropolis Hastings tests to make it more reliable (it was randomly failing for me).

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
@@ -142,6 +142,7 @@ public class MetropolisHastings implements PosteriorSamplingAlgorithm {
 
         @Override
         public NetworkState sample() {
+            step();
             return new SimpleNetworkState(takeSample(verticesToSampleFrom));
         }
     }
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/SamplingAlgorithm.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/SamplingAlgorithm.java
@@ -8,21 +8,20 @@ import io.improbable.keanu.network.NetworkState;
 public interface SamplingAlgorithm {
 
     /**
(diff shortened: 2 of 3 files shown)


christophernorth:
Fixes #190

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
```diff
@@ -142,6 +142,7 @@ public void sample(Map<Long, List<?>> samplesByVertex) {
 
         @Override
         public NetworkState sample() {
+            step();
```
icantsayrural: The doc in `SamplingAlgorithm#sample` is unclear to me.
```
    /**
     * @return a network state that represents the value of vertices at the
     * end of the algorithm step
     */
```
Whether or not it's being used to sample at the end of the algorithm step is outside of its scope, even more so if we add this in. Maybe it'll be better to change `NetworkSamplesGenerator#stream` to take one more step?

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/SamplingAlgorithm.java
```diff
@@ -8,20 +8,23 @@
...
 
     /**
-     * Same effect as a sample but the result isn't saved or returned.
+     * Move forward the state of the Sampling Algorithm by a single step.
      */
     void step();
 
     /**
-     * Takes a sample with the algorithm and saves it in the supplied map
+     * Takes a sample with the algorithm and saves it in the supplied map.  Repeated calls to this function will return
```
icantsayrural: > Repeated calls to this function will return the same values without an intermediary call to `step()`

I don't follow. This method doesn't explicitly return anything? And repeated calls will update the sample list for each vertex.

christophernorth: Good catch - I was documenting the wrong fix.  Made changes.

PR #204, opened 2018-08-24
Add reshape to vertices

GeorgeNash:
Adds reshape to `Double`, `Integer` and `Bool` vertices.

Also adds an `(int...)` constructor to `ones` in `IntegerTensor` to keep it consistent.
Also adds a `trues` and `falses` constructor for `BooleanTensor`

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/tensor/intgr/IntegerTensor.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/tensor/intgr/IntegerTensor.java
@@ -36,7 +36,7 @@ public interface IntegerTensor extends NumberTensor<Integer, IntegerTensor>, Int
         return create(values, 1, values.length);
     }
 
-    static IntegerTensor ones(int[] shape) {
+    static IntegerTensor ones(int... shape) {
         if (Arrays.equals(shape, Tensor.SCALAR_SHAPE)) {
             return new ScalarIntegerTensor(1);
         } else {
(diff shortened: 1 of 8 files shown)


keanu-project/src/main/java/io/improbable/keanu/tensor/intgr/IntegerTensor.java
```diff
@@ -36,7 +36,7 @@ static IntegerTensor create(int[] values) {
         return create(values, 1, values.length);
     }
 
-    static IntegerTensor ones(int[] shape) {
+    static IntegerTensor ones(int... shape) {
```
icantsayrural: 🎉

icantsayrural:
Shouldn't the user just pass the vertex to a reshape vertex?

GeorgeNash:
They can do, but people shouldn't really be creating new vertices. For example with multiplication we do `vertex.times(anotherVertex)` rather than `new MultiplicationVertex(vertex, anotherVertex)`. Both boil down to the same thing

icantsayrural:
Ah, I didn't see we support that for other operation vertices. Makes sense.

keanu-project/src/test/java/io/improbable/keanu/vertices/bool/BoolVertexTest.java
```diff
@@ -11,6 +11,9 @@
 
 import com.sun.org.apache.xpath.internal.operations.Bool;
 import io.improbable.keanu.tensor.bool.BooleanTensor;
+import io.improbable.keanu.tensor.intgr.IntegerTensor;
```
icantsayrural: Some unused imports

PR #207, opened 2018-08-28
Make Vertex ID an object instead of long.

christophernorth:
A long, but relatively simple PR that moves away from using a long value to represent a vertexID and instead uses a new VertexId object.  All files + tests should be migrated and all tests pass.

I've done this to make the graphception stuff a bit more elegant (more on that soon) and thought that given the amount of changes, best to get that reviewed separately.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
@@ -3,9 +3,9 @@ package io.improbable.keanu.vertices;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.atomic.AtomicLong;
 
 import com.google.common.collect.ImmutableSet;
 
@@ -16,9 +16,7 @@ import io.improbable.keanu.vertices.dbl.KeanuRandom;
 
 public abstract class Vertex<T> implements Observable<T> {
 
-    public static final AtomicLong ID_GENERATOR = new AtomicLong(0L);
-
-    private long uuid = ID_GENERATOR.getAndIncrement();
+    private VertexId id = new VertexId();
     private Set<Vertex> children = Collections.emptySet();
     private Set<Vertex> parents = Collections.emptySet();
     private T value;
@@ -172,8 +170,8 @@ public abstract class Vertex<T> implements Observable<T> {
         return observation.getObservedValue();
     }
 
-    public long getId() {
-        return uuid;
+    public VertexId getId() {
+        return id;
     }
 
     public Set<Vertex> getChildren() {
@@ -213,12 +211,12 @@ public abstract class Vertex<T> implements Observable<T> {
 
         Vertex<?> vertex = (Vertex<?>) o;
 
-        return uuid == vertex.uuid;
+        return Objects.equals(this.id, vertex.id);
     }
 
     @Override
     public int hashCode() {
-        return (int) (uuid ^ (uuid >>> 32));
+        return (int)Objects.hash(id);
     }
 
     public Set<Vertex> getConnectedGraph() {
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
@@ -1,0 +1,63 @@
+package io.improbable.keanu.vertices;
+
+import java.util.LinkedList;
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicLong;
+
+import com.google.common.primitives.Ints;
+
+/**
+ * An object representing the ID value of a vertex.  IDs are assigned in such a way that a Lexicographic ordering of
+ * all nodes based on their ID value is a valid topological ordering of the graph made up by those vertices.
+ *
+ * Ids also encapsulate the notion of "Depth".  When we have graphs within graphs, the depth tells us at what level the
+ * graph exists - ie depth 1 is the outermost graph, depth 2 is a graph within a graph etc.
+ */
+public class VertexId implements Comparable<VertexId> {
+
+    public static final AtomicLong ID_GENERATOR = new AtomicLong(0L);
+
+    LinkedList<Long> idValues = new LinkedList<>();
+
+    public VertexId() {
+        long newId = ID_GENERATOR.getAndIncrement();
+        idValues.push(newId);
+    }
+
+    @Override
+    public int compareTo(VertexId that) {
+        long comparisonValue = 0;
+
+        for (int i = 0; i < Math.min(this.idValues.size(), that.idValues.size()) && comparisonValue == 0; i++) {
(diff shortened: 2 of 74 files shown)


keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
```diff
@@ -0,0 +1,63 @@
...
+ * Ids also encapsulate the notion of "Depth".  When we have graphs within graphs, the depth tells us at what level the
+ * graph exists - ie depth 1 is the outermost graph, depth 2 is a graph within a graph etc.
+ */
+public class VertexId implements Comparable<VertexId> {
+
+    public static final AtomicLong ID_GENERATOR = new AtomicLong(0L);
+
+    LinkedList<Long> idValues = new LinkedList<>();
+
+    public VertexId() {
```
icantsayrural: If I understand this correctly, each vertex needs to keep track of its parents' id values? Should we add a non-default constructor that takes `VertexId`?

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
```diff
@@ -0,0 +1,63 @@
...
+        long newId = ID_GENERATOR.getAndIncrement();
+        idValues.push(newId);
+    }
+
+    @Override
+    public int compareTo(VertexId that) {
+        long comparisonValue = 0;
+
+        for (int i = 0; i < Math.min(this.idValues.size(), that.idValues.size()) && comparisonValue == 0; i++) {
```
icantsayrural: nit: might be worth doing `min` once outside the loop

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
```diff
@@ -213,12 +211,12 @@ public boolean equals(Object o) {
 
         Vertex<?> vertex = (Vertex<?>) o;
 
-        return uuid == vertex.uuid;
+        return Objects.equals(this.id, vertex.id);
```
icantsayrural: `Objects.equals` returns true if both ids are null. Is this what we want?

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
christophernorth: It doesn't keep track of all of its parents values - the ID just becomes a sequence of longs dependent upon its position in the graphception hierarchy (ie if Vertex 10 is stored in BayesVertex ID 5, then the ID would be [5, 10] - completely unrelated to the parent vertices.

The stuff to actually construct these properly will come in as part of the full graphception PR - this is just the code to move things out such that everything works as it currently does.

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
christophernorth: Yeah, I was just trying to be as explicit as possible, but I'll update.

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
christophernorth: Someone would have to explicitly write null to the ID to make that happen - for now - I'll just make the ID final so that this definitely can't happen - ok?

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
```diff
@@ -0,0 +1,64 @@
...
+        idValues.push(newId);
+    }
+
+    @Override
+    public int compareTo(VertexId that) {
+        long comparisonValue = 0;
+        int  minOfListSizes = Math.min(this.idValues.size(), that.idValues.size());
+
+        for (int i = 0; i < minOfListSizes && comparisonValue == 0; i++) {
+            comparisonValue = this.idValues.get(i) - that.idValues.get(i);
```
gordoncaleb: Using `.get(i)` on a LinkedList is going to be potentially very inefficient. For the compareTo contract I *think* you only need to return a -1, 0 or 1 for the comparison. In which case you should be able to return on id size differences or in the case where the size is the same, the comparison results of the first different id.

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
```diff
@@ -0,0 +1,64 @@
...
+
+        return Ints.saturatedCast(comparisonValue);
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        VertexId vertexId = (VertexId) o;
+        return compareTo(vertexId) == 0;
```
gordoncaleb: If you use an array, you should be able to just use `Arrays.equals(...)`

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public int hashCode() {
```
gordoncaleb: You can use IntelliJ to generate good `hashCode` and `equals` methods. These look like you may have manually written them.

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
gordoncaleb: Or use Lombok `@EqualsAndHashCode` https://projectlombok.org/features/EqualsAndHashCode

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
```diff
@@ -213,12 +211,12 @@ public boolean equals(Object o) {
...
     }
 
     @Override
     public int hashCode() {
-        return (int) (uuid ^ (uuid >>> 32));
+        return (int)Objects.hash(id);
```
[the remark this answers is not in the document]
christophernorth: Strangely this is what Intellij generated - I'll fix it up.

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
christophernorth: Yeah - I knew about the .get() being inefficient - as we discussed the LinkedList was a temporary placeholder whilst I went through the standard library to make sure there was nothing else more appropriate.  In my latest code I use an array as we discussed.

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
christophernorth: Yep - I think I'll use this in future - the IntelliJ generated code does't seem to pass our internal standards :)

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
christophernorth: I actually did generate with IntelliJ.

PR #211, opened 2018-08-30
Add logProb to NetworkSamples

icantsayrural:


keanu-project/src/main/java/io/improbable/keanu/algorithms/NetworkSamples.java
```diff
@@ -78,8 +82,9 @@ public NetworkSamples downSample(final int downSampleInterval) {
                 e -> downSample(e.getValue(), downSampleInterval)
                 )
             );
+        final List<Double> withLogProbsDownSampled = (List<Double>) downSample(logOfMasterP, downSampleInterval);
 
-        return new NetworkSamples(withSamplesDownSampled, size / downSampleInterval);
+        return new NetworkSamples(withSamplesDownSampled, withLogProbsDownSampled, size / downSampleInterval);
```
christophernorth: This just seems to be the pattern this code uses at the moment but obviously for large network samples recreating all this stuff is going to have pretty high transitory memory usage and is not gonna be great performance wise.  Should we be considering "inPlace" kind of operations?

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
```diff
@@ -92,6 +93,8 @@ public NetworkSamples getPosteriorSamples(final BayesianNetwork bayesNet,
         final Map<VertexId, DoubleTensor> momentumBeforeLeapfrog = new HashMap<>();
 
         double logOfMasterPBeforeLeapfrog = bayesNet.getLogOfMasterP();
+        final List<Double> logProbs = new LinkedList<>();
```
christophernorth: Is LinkedList the best choice here?  Seems like the overheads to walk and get etc will be pretty high - Perhaps we should consider an ArrayList?

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
```diff
@@ -136,9 +136,10 @@ public void step() {
...
 
         @Override
-        public void sample(Map<VertexId, List<?>> samplesByVertex) {
+        public void sample(Map<VertexId, List<?>> samplesByVertex, List<Double> logProbs) {
             step();
             takeSamples(samplesByVertex, verticesToSampleFrom);
+            logProbs.add(logProbabilityBeforeStep);
         }
 
         @Override
```
christophernorth: Should we be adding support for streaming the samples with an associated log prob too?  It seems for our use case, the streaming stuff might be a much more efficient way of running KL divergence, rather than building up a large set of samples?

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/SamplingAlgorithm.java
```diff
@@ -17,9 +17,10 @@
      * Takes a sample with the algorithm and saves it in the supplied map (creating a new entry in the list if the
      * Vertex already exists).
      *
-     * @param samples map to store sampled vertex values
+     * @param samples  map to store sampled vertex values
+     * @param logProbs list of log probabilities for each sample
```
christophernorth: Might be worth documenting what the "LogProb" is here and how it lines up with the samples Map (ie presumably length logProbs == length of each of the values from the samples Map

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/Prior.java
```diff
@@ -44,14 +46,17 @@ public static NetworkSamples sample(BayesianNetwork bayesNet,
...
+        List<Double> logProbs = new LinkedList<>();
 
         for (int sampleNum = 0; sampleNum < sampleCount; sampleNum++) {
             nextSample(topologicallySorted, random);
             takeSamples(samplesByVertex, fromVertices);
+            logProbs.add(bayesNet.getLogOfMasterP());
         }
 
-        return new NetworkSamples(samplesByVertex, sampleCount);
+        return new NetworkSamples(samplesByVertex, logProbs, sampleCount);
```
christophernorth: Looks like Caleb is deleting this file - you might want to check with him if it's worth keeping it around?

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/RejectionSampler.java
```diff
@@ -74,17 +75,20 @@ public static NetworkSamples getPosteriorSamples(BayesianNetwork bayesNet,
...
             sampleLatents(bayesNet.getLatentVertices(), random);
             if (matchesObservation(bayesNet.getObservedVertices())) {
                 takeSamples(samples, fromVertices);
+                logProbs.add(bayesNet.getLogOfMasterP());
                 acceptedCount++;
             }
         }
 
-        return new NetworkSamples(samples, sampleCount);
+        return new NetworkSamples(samples, logProbs, sampleCount);
```
christophernorth: As above.

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/Prior.java
icantsayrural: Yup that should be OK, I just made the change to satisfy the contracts so it doesn't break the tests.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
icantsayrural: Agreed. Talked about this with Caleb and that's something we'll add after. Right now individual samples don't have a fitting representation (except the network state, but that isn't specific to samples), so we can't track probabilities.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
icantsayrural: Agreed!

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/Prior.java
gordoncaleb: Sorry guys! This algo and the RejectionSampler aren't really useful. Prior is replaced by MetropolisHastings and the RejectionSampler only works for a very specific type of problem and even in that case it's effectively a bad version of MetropolisHastings.

keanu-project/src/main/java/io/improbable/keanu/algorithms/NetworkSamples.java
```diff
@@ -78,8 +81,9 @@ public NetworkSamples downSample(final int downSampleInterval) {
                 e -> downSample(e.getValue(), downSampleInterval)
                 )
             );
+        final List<Double> withLogProbsDownSampled = (List<Double>) downSample(logOfMasterPForEachSample, downSampleInterval);
```
gordoncaleb: Looks like you've repurposed the `downSample(...)` method to down sample probabilities as well. Good work! Can we make the method generic? It might require a cast on `getDoubleTensorSamples(...)` and `getIntegerTensorSamples(...)` but that's where I would expect the cast instead of here with `withLogProbsDownSampled `.

keanu-project/src/main/java/io/improbable/keanu/algorithms/NetworkSamples.java
```diff
@@ -106,6 +110,18 @@ public double probability(Function<NetworkState, Boolean> predicate) {
         return (double) trueCount / networkStates.size();
     }
 
+    public List<Object> getSample(int sample) {
```
gordoncaleb: Are there any test for these two methods?

gordoncaleb: Should this return a NetworkState at a specified sample number? Otherwise how can the user know with object belongs to which RV/operation result?

icantsayrural: They are just getters, so I didn't feel the need to

PR #213, opened 2018-08-31
Rocket test streaming and remove old algos

gordoncaleb:
This PR converts the rocket launch example to use streams which is faster and shows how to use streaming effectively. It also cleans up the rocket launch graph to use a conditional probability table instead of a few extra RVs and boolean ops.

The Prior algo and the RejectionSampler aren't really useful. Prior is replaced by MetropolisHastings and the RejectionSampler only works for a very specific type of problem and even in that case it's effectively a bad version of MetropolisHastings. So I've removed where they are used and then removed them entirely.


keanu-project/src/test/java/io/improbable/keanu/e2e/rocket/RocketTest.java
```diff
@@ -95,4 +77,27 @@ public void shouldYouLaunch() {
...
+            alarm1FalsePositiveCount = alarm1FalsePositive ? 1 : 0;
+        }
+
+        public EventCounts combine(EventCounts that) {
+            EventCounts counts = new EventCounts();
+            counts.oRingFailureCount = this.oRingFailureCount + that.oRingFailureCount;
+            counts.residualFuelCount = this.residualFuelCount + that.residualFuelCount;
+            counts.alarm1FalsePositiveCount = this.alarm1FalsePositiveCount + that.alarm1FalsePositiveCount;
+            return counts;
+        }
```
christophernorth: I'd just add the values to the current object rather than just creating a new object every time, but I'm pretty old fashioned :)

PR #214, opened 2018-08-31
Basic Graphception

christophernorth:
These diffs provide basic Graphception support (renamed to ModelComposition in the actual code)

It allows a user to pass a BayesNet to a function which will then encapsulate the model using the new "depth" mechanism in the ID object, hook up real input vertices to output vertices (using the new Proxy Vertices) and will ensure that everything is consistent in the global "computational graph" (ie ID ordering implies Topological Ordering, and that labels don't "leak" between BayesNets).  Also added convenience functions to the Bayesnet object to allow a user to only retrieve things at a certain depth (mainly useful to allow a user to see a simplified version of the top level graph).

There's more we can do with this like potentially changing how the BayesNet functions work publicly (ie only return vertices from the correct depth rather than the whole computational graph) - add methods to only return certain subgraphs (ie provide accessors on something more fine grained than depth) but I didn't want to add too much before we thought on what would be useful.  Feel free to provide comments on APIs you think would be useful and use cases and I can either add here if essential or raise more PRs as we go forward.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
@@ -22,6 +22,7 @@ public class BayesianNetwork {
 
     private final List<? extends Vertex> vertices;
     private final Map<VertexLabel, Vertex> vertexLabels;
+    private int depth = 1;
 
     public BayesianNetwork(Set<? extends Vertex> vertices) {
         this.vertices = ImmutableList.copyOf(vertices);
@@ -159,6 +160,14 @@ public class BayesianNetwork {
         return getLatentVertices().stream()
             .filter(v -> !(v.getValue() instanceof DoubleTensor))
             .collect(Collectors.toList());
+    }
+
+    public int getDepth() {
+        return depth;
+    }
+
+    public void incrementDepth() {
+        depth++;
     }
 
 }
--- a/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
@@ -37,11 +37,11 @@ public class BayesianNetwork {
         return vertexLabels.get(label);
     }
 
-    private static Map<VertexLabel, Vertex> buildLabelMap(Set<? extends Vertex> vertices) {
+    private Map<VertexLabel, Vertex> buildLabelMap(Set<? extends Vertex> vertices) {
         Map<VertexLabel, Vertex> labelMap = new HashMap<>();
         for (Vertex v : vertices) {
             VertexLabel label = v.getLabel();
-            if (label != null) {
+            if (v.getId().getDepth() == this.depth && label != null) {
                 if (labelMap.containsKey(label)) {
                     throw new IllegalArgumentException("Vertex Label Repeated: " + label);
                 } else {
--- a/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
@@ -76,7 +76,7 @@ public class BayesianNetwork {
 
     public List<Vertex> getLatentAndObservedVerticesAtDepth(int maxDepth) {
         return getFilteredVertexList((isProbabilistic, isObserved, depth)
-            -> (isProbabilistic || isObserved) && maxDepth >= this.depth);
+            -> (isProbabilistic || isObserved) && maxDepth >= depth);
     }
 
     /**
@@ -88,7 +88,7 @@ public class BayesianNetwork {
 
     public List<Vertex> getLatentVerticesAtDepth(int maxDepth) {
         return getFilteredVertexList((isProbabilistic, isObserved, depth)
-            -> (isProbabilistic && !isObserved) && maxDepth >= this.depth);
+            -> (isProbabilistic && !isObserved) && maxDepth >= depth);
     }
 
     /**
@@ -100,7 +100,7 @@ public class BayesianNetwork {
 
     public List<Vertex> getObservedVerticesAtDepth(int maxDepth) {
         return getFilteredVertexList((isProbabilistic, isObserved, depth) ->
-            isObserved && maxDepth >= this.depth);
+            isObserved && maxDepth >= depth);
     }
 
     public double getLogOfMasterP() {
--- a/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
@@ -57,13 +57,13 @@ public class BayesianNetwork {
         return vertices;
     }
 
-    private interface VertexProbabilisticObservedFilter {
-        boolean filter(boolean isObserved, boolean b);
+    private interface VertexFilter {
+        boolean filter(boolean isProbabilistic, boolean isObserved, int depth);
     }
 
-    private List<Vertex> getFilteredVertexList(VertexProbabilisticObservedFilter filter) {
+    private List<Vertex> getFilteredVertexList(VertexFilter filter) {
         return vertices.stream()
-            .filter(v -> filter.filter(v.isProbabilistic(), v.isObserved()))
+            .filter(v -> filter.filter(v.isProbabilistic(), v.isObserved(), v.getId().getDepth()))
             .collect(Collectors.toList());
     }
 
@@ -71,23 +71,36 @@ public class BayesianNetwork {
      * @return The union of getLatentVertices and getObservedVertices
      */
     public List<Vertex> getLatentAndObservedVertices() {
-        return getFilteredVertexList((isProbabilistic, isObserved) -> isProbabilistic || isObserved);
+        return getLatentAndObservedVerticesAtDepth(Integer.MAX_VALUE);
     }
 
-    //public List<Vertex> getLatent
+    public List<Vertex> getLatentAndObservedVerticesAtDepth(int maxDepth) {
+        return getFilteredVertexList((isProbabilistic, isObserved, depth)
+            -> (isProbabilistic || isObserved) && maxDepth >= this.depth);
+    }
 
     /**
      * @return All vertices that are latent (i.e. probabilistic non-observed)
      */
     public List<Vertex> getLatentVertices() {
-        return getFilteredVertexList((isProbabilistic, isObserved) -> isProbabilistic && !isObserved);
+        return getLatentVerticesAtDepth(Integer.MAX_VALUE);
+    }
+
+    public List<Vertex> getLatentVerticesAtDepth(int maxDepth) {
+        return getFilteredVertexList((isProbabilistic, isObserved, depth)
+            -> (isProbabilistic && !isObserved) && maxDepth >= this.depth);
     }
 
     /**
      * @return All vertices that are observed - which may be probabilistic or non-probabilistic
      */
     public List<Vertex> getObservedVertices() {
-        return getFilteredVertexList((isProbabilistic, isObserved) -> isObserved);
+        return getObservedVerticesAtDepth(Integer.MAX_VALUE);
+    }
+
+    public List<Vertex> getObservedVerticesAtDepth(int maxDepth) {
+        return getFilteredVertexList((isProbabilistic, isObserved, depth) ->
+            isObserved && maxDepth >= this.depth);
     }
 
     public double getLogOfMasterP() {
--- a/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
@@ -53,7 +53,7 @@ public class BayesianNetwork {
         return labelMap;
     }
 
-    public List<? extends Vertex> getAllVertices() {
+    List<? extends Vertex> getAllVertices() {
         return vertices;
     }
 
--- a/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
@@ -57,31 +57,37 @@ public class BayesianNetwork {
         return vertices;
     }
 
+    private interface VertexProbabilisticObservedFilter {
+        boolean filter(boolean isObserved, boolean b);
+    }
+
+    private List<Vertex> getFilteredVertexList(VertexProbabilisticObservedFilter filter) {
+        return vertices.stream()
+            .filter(v -> filter.filter(v.isProbabilistic(), v.isObserved()))
+            .collect(Collectors.toList());
+    }
+
     /**
      * @return The union of getLatentVertices and getObservedVertices
      */
     public List<Vertex> getLatentAndObservedVertices() {
-        return vertices.stream()
-            .filter(v -> v.isProbabilistic() || v.isObserved())
-            .collect(Collectors.toList());
+        return getFilteredVertexList((isProbabilistic, isObserved) -> isProbabilistic || isObserved);
     }
+
+    //public List<Vertex> getLatent
 
     /**
      * @return All vertices that are latent (i.e. probabilistic non-observed)
      */
     public List<Vertex> getLatentVertices() {
-        return vertices.stream()
-            .filter(v -> v.isProbabilistic() && !v.isObserved())
-            .collect(Collectors.toList());
+        return getFilteredVertexList((isProbabilistic, isObserved) -> isProbabilistic && !isObserved);
     }
 
     /**
      * @return All vertices that are observed - which may be probabilistic or non-probabilistic
      */
     public List<Vertex> getObservedVertices() {
-        return vertices.stream()
-            .filter(Vertex::isObserved)
-            .collect(Collectors.toList());
+        return getFilteredVertexList((isProbabilistic, isObserved) -> isObserved);
     }
 
     public double getLogOfMasterP() {
--- a/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
@@ -57,6 +57,9 @@ public class BayesianNetwork {
         return vertices;
     }
 
+    /**
+     * @return The union of getLatentVertices and getObservedVertices
+     */
     public List<Vertex> getLatentAndObservedVertices() {
         return vertices.stream()
             .filter(v -> v.isProbabilistic() || v.isObserved())
--- a/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
@@ -52,6 +52,10 @@ public class BayesianNetwork {
         return labelMap;
     }
 
+    public List<? extends Vertex> getAllVertices() {
+        return vertices;
+    }
+
     public List<Vertex> getLatentAndObservedVertices() {
         return vertices.stream()
             .filter(v -> v.isProbabilistic() || v.isObserved())
--- a/keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
@@ -60,4 +60,5 @@ public class ModelComposition {
             }
         }
     }
+
 }
--- a/keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
@@ -14,7 +14,7 @@ public class ModelComposition {
                                                                Map<VertexLabel, Vertex> inputVertices,
                                                                List<VertexLabel> desiredOutputs) {
         Map<VertexLabel, Vertex> outputMap = extractOutputs(bayesianNetwork, desiredOutputs);
-        increaseDepthForAllNodes(bayesianNetwork);
+        increaseDepth(bayesianNetwork);
         checkAndLinkInputs(bayesianNetwork, inputVertices);
 
         return outputMap;
@@ -39,7 +39,8 @@ public class ModelComposition {
         return outputMap;
     }
 
-    private static void increaseDepthForAllNodes(BayesianNetwork bayesianNetwork) {
+    private static void increaseDepth(BayesianNetwork bayesianNetwork) {
+        bayesianNetwork.incrementDepth();
         for (Vertex v : bayesianNetwork.getAllVertices()) {
             v.getId().increaseDepth();
         }
--- a/keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
@@ -6,6 +6,7 @@ import java.util.Map;
 
 import io.improbable.keanu.vertices.ProxyVertex;
 import io.improbable.keanu.vertices.Vertex;
+import io.improbable.keanu.vertices.VertexId;
 import io.improbable.keanu.vertices.VertexLabel;
 
 public class ModelComposition {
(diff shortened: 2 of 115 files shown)


keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
```diff
@@ -17,13 +17,26 @@
...
 
+    public void increaseDepth(VertexId prefix) {
+        long[] newIdValues = new long[idValues.length + prefix.idValues.length];
+        System.arraycopy(prefix.idValues, 0, newIdValues, 0, prefix.idValues.length);
+        System.arraycopy(idValues, 0, newIdValues, prefix.idValues.length, idValues.length);
+        idValues = newIdValues;
+    }
+
+    public void resetID() {
+        idValues = new long[DEPTH_ONE_ARRAY_SIZE];
```
icantsayrural: Can a model be nested inside a nested model? Should the output vertices always get reset to depth 1 in that case?

keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
```diff
@@ -0,0 +1,98 @@
...
+                } else {
+                    proxyVertex.setParent(entry.getValue());
+                }
+            } else {
+                throw new IllegalArgumentException("Input node \"" + entry.getKey() + "\" is not a Proxy Vertex");
+            }
+        }
+    }
+
+    private static void cleanOutputLabels(Map<VertexLabel, Vertex> outputMap) {
```
icantsayrural: Why don't we clean input and proxy vertices?

keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
```diff
@@ -0,0 +1,98 @@
...
+            outputMap.put(label, v);
+        }
+
+        return outputMap;
+    }
+
+    private static void increaseDepth(BayesianNetwork bayesianNetwork, Map<VertexLabel, Vertex> outputVertices) {
+        VertexId newPrefix = new VertexId();
+        bayesianNetwork.incrementDepth();
+        bayesianNetwork.getAllVertices().stream()
```
icantsayrural: Could we pass a filter (and maybe also `increaseDepth`) to `BayesianNetwork` instead of asking for all the vertices?

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
christophernorth: Yep - a model can be nested within another - but with how everything is set up at the moment you can only really operate on the "outermost" graph at any particular point (so when you call ModelComposition methods you're operating by necessity on the outermost graph).

keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
christophernorth: This is a little subtle and relates to how I've ensured that we create a "namespace" for labels.
What I'm trying to avoid is a situation, for instance, where we use two instances of a subgraph - both of these will have Vertices with the same labels - meaning we have a clash of labels in the outer graph.

The output vertices themselves "pop out" of the inner graph and exist at the same level as the parent graph whereas all the other vertices get pushed down to a lower depth.  When creating the label map in the BayesNet I only look at vertices at the top level - effectively creating separate "namespaces" for Vertex Labels.  The exception is obviously the output vertices which will be picked up, so I clean the labels explicitly for this.

This is all sort of temporary stuff until we have the full functionality to "instantiate" graphs from models at which point a lot of this stuff can be cleaned up a bit.

keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
christophernorth: Probably easiest to chat on this - it would be possible to do this, but I'm not sure it would necessarily be clearer (and we're just returning a list reference, so there's no perf concern here).

I wanted to keep all this logic together for now as really this process is more of something we'll do to a "model" rather than an "instantiation" (ie a BayesNet) so in future this code won't likely interact with a BayesNet at all anyway...

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/nonprobabilistic/BoolProxyVertex.java
```diff
@@ -0,0 +1,51 @@
...
+import io.improbable.keanu.tensor.Tensor;
+import io.improbable.keanu.tensor.bool.BooleanTensor;
+import io.improbable.keanu.vertices.NonProbabilistic;
+import io.improbable.keanu.vertices.ProxyVertex;
+import io.improbable.keanu.vertices.bool.BoolVertex;
+import io.improbable.keanu.vertices.dbl.KeanuRandom;
+
+public class BoolProxyVertex extends BoolVertex implements ProxyVertex<BoolVertex>, NonProbabilistic<BooleanTensor> {
+
+    private BoolVertex parentVertex;
```
migwellian: you shouldn't need to store `parentVertex` or `shape` because the information already exists in the parent class `Vertex`. For example:

    public class BoolProxyVertex extends BoolVertex implements ProxyVertex<BoolVertex>, NonProbabilistic<BooleanTensor> {
    
        /**
         * This vertex acts as a "Proxy" to allow a BayesNet to be built up before parents are explicitly known (ie for
         * model in model scenarios) but allows linking at a later point in time.
         */
        public BoolProxyVertex() {
            this(Tensor.SCALAR_SHAPE);
        }
    
        public BoolProxyVertex(int[] shape) {
            this.setValue(BooleanTensor.placeHolder(shape));
        }
    
        @Override
        public BooleanTensor calculate() {
            return getParent().getValue();
        }
    
        @Override
        public BooleanTensor sample(KeanuRandom random) {
            return getParent().sample();
        }
    
        @Override
        public void setParent(BoolVertex newParent) {
            checkTensorsMatchNonScalarShapeOrAreScalar(getShape(), newParent.getShape());
            setParents(newParent);
        }
    
        @Override
        public BoolVertex getParent() {
            return (BoolVertex) Iterables.getOnlyElement(getParents());
        }
    
        @Override
        public boolean hasParent() {
            return getParents().size() > 0;
        }
    }

keanu-project/src/main/java/io/improbable/keanu/vertices/ProxyVertex.java
```diff
@@ -0,0 +1,11 @@
+package io.improbable.keanu.vertices;
+
+/**
+ * A Proxy vertex is one which has no parents at creation time but can at a later point "hook up" a parent node.  All
+ * Vertex operations are then passed through to the parent (as if the they'd been called on the parent directly)
```
migwellian: This description is true of your classes that implement `ProxyVertex`, but it's not a true description of the `ProxyVertex` interface itself. I think something along these lines is more accurate:

     * A Proxy vertex is one with a single parent. It is expected that all vertex operations will be delegated to the parent.
     * It typically has no parents at creation time - the parent is node is "hooked up" at a later point

keanu-project/src/main/java/io/improbable/keanu/vertices/generic/DoubleBayesNetVertex.java
```diff
@@ -0,0 +1,4 @@
+package io.improbable.keanu.vertices.generic;
+
+public class DoubleBayesNetVertex {
```
migwellian: This class does nothing and isn't used?

keanu-project/src/test/java/io/improbable/keanu/vertices/bool/BoolVertexTest.java
```diff
@@ -1,12 +1,11 @@
 package io.improbable.keanu.vertices.bool;
 
+import static io.improbable.keanu.vertices.bool.BoolVertex.not;
```
migwellian: looks like your IntelliJ settings are different to [mine](https://improbableio.atlassian.net/wiki/spaces/~616259109/pages/283574476/IntelliJ+settings)

keanu-project/src/test/java/io/improbable/keanu/network/ModelCompositionTest.java
```diff
@@ -0,0 +1,126 @@
...
+    private BayesianNetwork outerNet;
+    private DoubleVertex location;
+    private DoubleVertex size;
+    private DoubleVertex gaussian;
+    private DoubleVertex pareto;
+    private DoubleVertex gaussOutputVertex;
+    private DoubleVertex paretoOutputVertex;
+
+    @Before
+    public void setup() {
```
migwellian: some comments here would be helpful, to describe the structure of the graph.

keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
```diff
@@ -0,0 +1,98 @@
...
+        if (desiredOutputs.size() == 0) {
+            throw new IllegalArgumentException("At least one output must be specified");
+        }
+
+        Map<VertexLabel, Vertex> outputMap = new HashMap<>();
+
+        for (VertexLabel label: desiredOutputs) {
+            Vertex v = bayesianNetwork.getVertexByLabel(label);
+            if (v == null) {
+                throw new IllegalArgumentException("Unable to find Output Vertex: " + label);
```
migwellian: could you add a test for this?

keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
```diff
@@ -0,0 +1,98 @@
...
+            Vertex v = bayesianNetwork.getVertexByLabel(entry.getKey());
+
+            if (v == null) {
+                throw new IllegalArgumentException("No node labelled \"" + entry.getKey() + "\" found");
+            }
+
+            if (v instanceof ProxyVertex) {
+                ProxyVertex proxyVertex = (ProxyVertex)v;
+                if (proxyVertex.hasParent()) {
+                    throw new IllegalArgumentException("Proxy Vertex for \"" + v.getLabel() + "\" already has Parent");
```
migwellian: could you add a test for this?

keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
```diff
@@ -0,0 +1,98 @@
...
+            .filter(v -> outputVertices.containsKey(v.getLabel()))
+            .forEach(v -> v.getId().resetID());
+    }
+
+    private static void checkAndLinkInputs(BayesianNetwork bayesianNetwork, Map<VertexLabel, Vertex> inputs) {
+        for (Map.Entry<VertexLabel, Vertex> entry: inputs.entrySet()) {
+            Vertex v = bayesianNetwork.getVertexByLabel(entry.getKey());
+
+            if (v == null) {
+                throw new IllegalArgumentException("No node labelled \"" + entry.getKey() + "\" found");
```
migwellian: could you add a test for this?

keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
```diff
@@ -0,0 +1,98 @@
...
+
+            if (v instanceof ProxyVertex) {
+                ProxyVertex proxyVertex = (ProxyVertex)v;
+                if (proxyVertex.hasParent()) {
+                    throw new IllegalArgumentException("Proxy Vertex for \"" + v.getLabel() + "\" already has Parent");
+                } else {
+                    proxyVertex.setParent(entry.getValue());
+                }
+            } else {
+                throw new IllegalArgumentException("Input node \"" + entry.getKey() + "\" is not a Proxy Vertex");
```
migwellian: could you add a test for this?

keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
migwellian: Instead of having the output vertices "pop out" of the inner graph, would it be possible (and not too difficult) to leave them (labelled) in the inner graph but link them to (new) vertices in the outer graph? That way it's consistent with how you link the input vertices, and probably easier to understand how all this works.

christophernorth: I'll chat to you Mike - it's not ideal to not pop them out (ie I can't use the same proxy vertex trick I've used for inputs) and this is the most elegant solution I could think of.

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -22,6 +22,7 @@
 
     private final List<? extends Vertex> vertices;
     private final Map<VertexLabel, Vertex> vertexLabels;
+    private int depth = 1;
```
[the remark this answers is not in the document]
christophernorth: I'll have a think

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/nonprobabilistic/BoolProxyVertex.java
christophernorth: Cool - good suggestion.

keanu-project/src/main/java/io/improbable/keanu/vertices/generic/DoubleBayesNetVertex.java
christophernorth: Aha - sorry about this.  This was left around from one of my original prototypes for this feature -I'll get it deleted.

keanu-project/src/test/java/io/improbable/keanu/vertices/bool/BoolVertexTest.java
christophernorth: I've just gone through these and the import stuff is as defined on StackOverflow?

keanu-project/src/test/java/io/improbable/keanu/network/ModelCompositionTest.java
christophernorth: I'll add a short comment.

keanu-project/src/test/java/io/improbable/keanu/network/ModelCompositionTest.java
```diff
@@ -33,23 +36,32 @@
...
     private DoubleVertex gaussOutputVertex;
-    private DoubleVertex paretoOutputVertex;
 
+    /*
+     * These tests run over a single graph within a graph.  The "inner" graph has a single Proxy input for the location
+     * a probabilistic value for size and two potential outputs:
+     *
+     * | Proxy "Location" |              | Uniform - Size |
+     *                     \            /
+     *     | Gaussian ("Output1" & Pareto ("Output2") |
```
migwellian: missed a `)`

keanu-project/src/test/java/io/improbable/keanu/network/ModelCompositionTest.java
migwellian: Looks great, thanks!

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -52,28 +53,54 @@ public Vertex getVertexByLabel(VertexLabel label) {
         return labelMap;
     }
 
-    public List<Vertex> getLatentAndObservedVertices() {
+    List<? extends Vertex> getAllVertices() {
```
GeorgeNash: `getVertices`

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -52,28 +53,54 @@ public Vertex getVertexByLabel(VertexLabel label) {
...
+        return getLatentAndObservedVerticesAtDepth(Integer.MAX_VALUE);
+    }
+
+    public List<Vertex> getLatentAndObservedVerticesAtDepth(int maxDepth) {
+        return getFilteredVertexList((isProbabilistic, isObserved, depth)
+            -> (isProbabilistic || isObserved) && maxDepth >= depth);
+    }
+
     /**
      * @return All vertices that are latent (i.e. probabilistic non-observed)
```
GeorgeNash: This is a bit confusing, let's match the style of the other comments in this class to keep it consistent: `All vertices are that are latent or observed`. Call the method `getLatentOrObservedVertices()`? And `getLatentOrObservedVerticesAtDepth`

keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
```diff
@@ -0,0 +1,98 @@
...
+import io.improbable.keanu.vertices.Vertex;
+import io.improbable.keanu.vertices.VertexId;
+import io.improbable.keanu.vertices.VertexLabel;
+
+public final class ModelComposition {
+
+    private ModelComposition() { }
+
+    /**
+     * Compose one model within a wider model.
```
GeorgeNash: Might be clearer to say connects one model to another, as the `compose within` makes it sound like it's compacted into one vertex

keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
```diff
@@ -0,0 +1,98 @@
...
+
+    private static Map<VertexLabel, Vertex> extractOutputs(BayesianNetwork bayesianNetwork,
+                                                           List<VertexLabel> desiredOutputs) {
+        if (desiredOutputs.size() == 0) {
+            throw new IllegalArgumentException("At least one output must be specified");
+        }
+
+        Map<VertexLabel, Vertex> outputMap = new HashMap<>();
+
+        for (VertexLabel label: desiredOutputs) {
```
GeorgeNash: nit: `for (VertexLabel label : desiredOutputs)`

keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
```diff
@@ -0,0 +1,98 @@
...
+        increaseDepth(bayesianNetwork, outputMap);
+        checkAndLinkInputs(bayesianNetwork, inputVertices);
+        cleanOutputLabels(outputMap);
+
+        return outputMap;
+    }
+
+    private static Map<VertexLabel, Vertex> extractOutputs(BayesianNetwork bayesianNetwork,
+                                                           List<VertexLabel> desiredOutputs) {
+        if (desiredOutputs.size() == 0) {
```
GeorgeNash: Swap this for `isEmpty()`

keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
```diff
@@ -0,0 +1,98 @@
...
+    private static void checkAndLinkInputs(BayesianNetwork bayesianNetwork, Map<VertexLabel, Vertex> inputs) {
+        for (Map.Entry<VertexLabel, Vertex> entry: inputs.entrySet()) {
+            Vertex v = bayesianNetwork.getVertexByLabel(entry.getKey());
+
+            if (v == null) {
+                throw new IllegalArgumentException("No node labelled \"" + entry.getKey() + "\" found");
+            }
+
+            if (v instanceof ProxyVertex) {
+                ProxyVertex proxyVertex = (ProxyVertex)v;
```
GeorgeNash: nit: `(ProxyVertex) v`

keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
```diff
@@ -0,0 +1,98 @@
...
+        bayesianNetwork.getAllVertices().stream()
+            .filter(v -> !outputVertices.containsKey(v.getLabel()))
+            .forEach(v -> v.getId().increaseDepth(newPrefix));
+        bayesianNetwork.getAllVertices().stream()
+            .filter(v -> outputVertices.containsKey(v.getLabel()))
+            .forEach(v -> v.getId().resetID());
+    }
+
+    private static void checkAndLinkInputs(BayesianNetwork bayesianNetwork, Map<VertexLabel, Vertex> inputs) {
+        for (Map.Entry<VertexLabel, Vertex> entry: inputs.entrySet()) {
```
GeorgeNash: nit: `entry :`

keanu-project/src/test/java/io/improbable/keanu/network/ModelCompositionTest.java
```diff
@@ -0,0 +1,206 @@
...
+     *                     \            /
+     *     | Gaussian ("Output1") & Pareto ("Output2") |
+     *
+     * The Outer graph hooks up a Uniform to the "Location" input of the inner graph and takes the Output "Output1" as
+     * the single output it cares about.
+     */
+    @Before
+    public void setup() {
+
+        createInnerNet();
```
GeorgeNash: This looks a bit odd - can you change `createInnerNet` to return a `BayesianNetwork` and do `innerNet = createInnerNet();`

keanu-project/src/main/java/io/improbable/keanu/network/ModelComposition.java
```diff
@@ -0,0 +1,98 @@
...
+     *     - Pass back any specified output nodes from the Bayesnet (keeping them at the outer depth)
+     *
+     *     Output nodes will be returned unlabelled
+     * </p>
+     * @param bayesianNetwork The Bayesian Network to compose in to the current model
+     * @param inputVertices The mapping from Proxy label to actual input vertex
+     * @param desiredOutputs The set of labels we wish to output from the supplied BayesNet
+     * @return A map of Labels to Output Vertices
+     */
+    public static Map<VertexLabel, Vertex> createModelVertices(BayesianNetwork bayesianNetwork,
```
GeorgeNash: Can we change the name to reflect what it's actually doing, something like `composeModel` or `connectModels` or `connectVertices`

keanu-project/src/test/java/io/improbable/keanu/network/ModelCompositionTest.java
```diff
@@ -0,0 +1,206 @@
...
+        assertEquals(trueLocation.getId().getDepth(), 1);
+        assertEquals(location.getId().getDepth(), 2);
+        assertEquals(size.getId().getDepth(), 2);
+        assertEquals(gaussian.getId().getDepth(), 1);
+        assertEquals(pareto.getId().getDepth(), 2);
+    }
+
+    @Test
+    public void netsAtCorrectDepth() {
+        assertEquals(innerNet.getIndentation(), 2);
```
GeorgeNash: why is the method on the bayes net called `getIndentation` and the one on the vertex called `getDepth`? Can we just call them the same thing?

keanu-project/src/test/java/io/improbable/keanu/network/ModelCompositionTest.java
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    @Test
+    public void outputVerticesCorrectlyReturned() {
+        assertEquals(gaussOutputVertex, gaussian);
+    }
+
+    @Test
+    public void verticesAreAtCorrectDepth() {
+        assertEquals(trueLocation.getId().getDepth(), 1);
```
GeorgeNash: This looks a bit odd and will be quite annoying for users. Can we chuck in a shortcut on the vertex class so we can do `trueLocation.getDepth()`? 

```
    public int getDepth() {
        return getId().idValues.length;
    }
```

keanu-project/src/test/java/io/improbable/keanu/network/ModelCompositionTest.java
```diff
@@ -0,0 +1,206 @@
...
+        assertEquals(innerNet.getIndentation(), 2);
+        assertEquals(outerNet.getIndentation(), 1);
+    }
+
+    @Test
+    public void idOrderingStillImpliesTopologicalOrdering() {
+        for (Vertex v: outerNet.getAllVertices()) {
+            Set<Vertex> parentSet = v.getParents();
+            for (Vertex parent : parentSet) {
+                assertTrue(v.getId().compareTo(parent.getId()) > 0);
```
GeorgeNash: Nice!

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/nonprobabilistic/BoolProxyVertex.java
```diff
@@ -0,0 +1,53 @@
...
+import io.improbable.keanu.vertices.bool.BoolVertex;
+import io.improbable.keanu.vertices.dbl.KeanuRandom;
+
+public class BoolProxyVertex extends BoolVertex implements ProxyVertex<BoolVertex>, NonProbabilistic<BooleanTensor> {
+
+    /**
+     * This vertex acts as a "Proxy" to allow a BayesNet to be built up before parents are explicitly known (ie for
+     * model in model scenarios) but allows linking at a later point in time.
+     */
+    public BoolProxyVertex() {
```
migwellian: Should we enforce (in the constructor) that a proxy vertex always has a label? Does it make sense to have an unlabelled proxy vertex?

christophernorth: Good call - no, Proxy Vertices really do need a label as otherwise they're useless - and making sure people create this as part of the constructor would be the nicest API.

PR #215, opened 2018-08-31
Add KL divergence

icantsayrural:
Equation:
<a href="https://www.codecogs.com/eqnedit.php?latex=D_{KL}(P||Q)&space;=&space;\sum_{i}&space;P(i)\log&space;\left(\frac{P(i)}{Q(i)}&space;\right&space;)" target="_blank"></a>

Requirements:
- Q can't have smaller support than P
- When P is 0, contribution to divergence is 0

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/KLDivergence.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/KLDivergence.java
@@ -3,8 +3,10 @@ package io.improbable.keanu.algorithms.variational;
 import com.google.common.collect.Iterables;
 import io.improbable.keanu.algorithms.NetworkSamples;
 import io.improbable.keanu.network.NetworkState;
+import io.improbable.keanu.vertices.VertexId;
 import io.improbable.keanu.vertices.dbl.probabilistic.ProbabilisticDouble;
 
+import java.util.Set;
 import java.util.function.Function;
 
 public class KLDivergence {
@@ -14,7 +16,14 @@ public class KLDivergence {
     }
 
     public static double compute(ProbabilisticDouble q, NetworkSamples p) {
-        return compute(p, networkState -> q.logProb(networkState.get(Iterables.getOnlyElement(networkState.getVertexIds()))));
+        return compute(p, networkState -> {
+            Set<VertexId> vertexIds = networkState.getVertexIds();
+            if (vertexIds.size() != 1) {
+                throw new IllegalArgumentException("A NetworkState does not contain exactly 1 vertex and ProbabilisticDouble can only compute the log probability of one value. Try computing KL divergence against a QDistribution instead.");
+            }
+
+            return q.logProb(networkState.get(Iterables.getOnlyElement(vertexIds)));
+        });
     }
 
     private static double compute(NetworkSamples samples, Function<NetworkState, Double> qLogProbCalculator) {
@@ -28,7 +37,7 @@ public class KLDivergence {
 
             if (pLogProb != Double.NEGATIVE_INFINITY) {
                 if (qLogProb == Double.NEGATIVE_INFINITY) {
-                    throw new IllegalArgumentException("Q cannot have smaller support than P");
+                    throw new IllegalArgumentException("Q cannot have smaller support than P.");
                 }
                 divergence += (pLogProb - qLogProb) * Math.exp(pLogProb);
             }
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/KLDivergence.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/KLDivergence.java
@@ -1,0 +1,39 @@
+package io.improbable.keanu.algorithms.variational;
+
+import com.google.common.collect.Iterables;
+import io.improbable.keanu.algorithms.NetworkSamples;
+import io.improbable.keanu.network.NetworkState;
+import io.improbable.keanu.vertices.dbl.probabilistic.ProbabilisticDouble;
+
+import java.util.function.Function;
+
+public class KLDivergence {
+
+    public static double compute(QDistribution q, NetworkSamples p) {
+        return compute(p, q::getLogOfMasterP);
+    }
+
+    public static double compute(ProbabilisticDouble q, NetworkSamples p) {
+        return compute(p, networkState -> q.logProb(networkState.get(Iterables.getOnlyElement(networkState.getVertexIds()))));
+    }
+
+    private static double compute(NetworkSamples samples, Function<NetworkState, Double> qLogProbCalculator) {
+        double divergence = 0.;
+
+        for (int i = 0; i < samples.size(); i++) {
+            double pLogProb = samples.getLogOfMasterP(i);
+
+            NetworkState state = samples.getNetworkState(i);
+            double qLogProb = qLogProbCalculator.apply(state);
+
+            if (pLogProb != Double.NEGATIVE_INFINITY) {
+                if (qLogProb == Double.NEGATIVE_INFINITY) {
+                    throw new IllegalArgumentException("Q cannot have smaller support than P");
+                }
+                divergence += (pLogProb - qLogProb) * Math.exp(pLogProb);
+            }
+        }
+
+        return divergence;
+    }
+}
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/QDistribution.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/QDistribution.java
@@ -1,0 +1,11 @@
+package io.improbable.keanu.algorithms.variational;
+
+import io.improbable.keanu.network.NetworkState;
+
+/**
+ * {@link QDistribution} represents Q in D(P‖Q) = sum_i P(i) log(P(i)/Q(i)), which is the
+ * {@link KLDivergence} (Kullback–Leibler divergence) from probability distributions P to Q.
+ */
+public interface QDistribution {
+    double getLogOfMasterP(NetworkState state);
+}
--- a/keanu-project/src/test/java/io/improbable/keanu/algorithms/variational/KLDivergenceTest.java
+++ b/keanu-project/src/test/java/io/improbable/keanu/algorithms/variational/KLDivergenceTest.java
@@ -14,10 +14,13 @@ import io.improbable.keanu.vertices.dbl.probabilistic.ProbabilisticDouble;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.lessThan;
 import static org.junit.Assert.assertEquals;
+import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
+import org.junit.rules.ExpectedException;
 
 import java.util.Arrays;
 import java.util.Collections;
@@ -27,42 +30,51 @@ import java.util.Map;
 
 public class KLDivergenceTest {
 
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    private QDistribution qDist;
+
+    @Before
+    public void setup() {
+        qDist = mock(QDistribution.class);
+    }
+
     @Test
     public void klDivergenceIsZeroIfPAndQAreIdentical() {
         double identicalProb = Math.log(0.5);
 
-        QDistribution qDist = mock(QDistribution.class);
         when(qDist.getLogOfMasterP(any(NetworkState.class))).thenReturn(identicalProb);
         NetworkSamples samples = createNetworkSamplesWithOneVertexAndOneSample(identicalProb);
 
         assertEquals(0., KLDivergence.compute(qDist, samples), 1e-6);
     }
 
-    @Test(expected = IllegalArgumentException.class)
+    @Test
     public void throwsExceptionIfPProbIsNotZeroButQIs() {
-        QDistribution q = mock(QDistribution.class);
-        when(q.getLogOfMasterP(any(NetworkState.class))).thenReturn(Double.NEGATIVE_INFINITY);
+        when(qDist.getLogOfMasterP(any(NetworkState.class))).thenReturn(Double.NEGATIVE_INFINITY);
         NetworkSamples samples = createNetworkSamplesWithOneVertexAndOneSample(Math.log(0.5));
 
-        KLDivergence.compute(q, samples);
+        thrown.expect(IllegalArgumentException.class);
+        thrown.expectMessage("Q cannot have smaller support than P");
+
+        KLDivergence.compute(qDist, samples);
     }
 
     @Test
     public void klDivergenceReturnsZeroIfPIsZeroButQIsNot() {
-        QDistribution q = mock(QDistribution.class);
-        when(q.getLogOfMasterP(any(NetworkState.class))).thenReturn(Math.log(0.5));
+        when(qDist.getLogOfMasterP(any(NetworkState.class))).thenReturn(Math.log(0.5));
         NetworkSamples samples = createNetworkSamplesWithOneVertexAndOneSample(Double.NEGATIVE_INFINITY);
 
-        assertEquals(0., KLDivergence.compute(q, samples), 1e-10);
+        assertEquals(0., KLDivergence.compute(qDist, samples), 1e-10);
     }
 
     @Test
     public void klDivergenceReturnsZeroIfPAndQAreZero() {
-        QDistribution q = mock(QDistribution.class);
-        when(q.getLogOfMasterP(any(NetworkState.class))).thenReturn(Double.NEGATIVE_INFINITY);
+        when(qDist.getLogOfMasterP(any(NetworkState.class))).thenReturn(Double.NEGATIVE_INFINITY);
         NetworkSamples samples = createNetworkSamplesWithOneVertexAndOneSample(Double.NEGATIVE_INFINITY);
 
-        assertEquals(0., KLDivergence.compute(q, samples), 1e-10);
+        assertEquals(0., KLDivergence.compute(qDist, samples), 1e-10);
     }
 
     @Test
@@ -99,8 +111,8 @@ public class KLDivergenceTest {
         assertThat(KLDivergence.compute(q1, samples), lessThan(KLDivergence.compute(q2, samples)));
     }
 
-    @Test(expected = IllegalArgumentException.class)
-    public void throwsExceptionIfNetworkStateHasMoreThanOneVertexAndQIsProbabilisticDouble(){
+    @Test
+    public void throwsExceptionIfNetworkStateHasMoreThanOneVertexAndQIsProbabilisticDouble() {
         GaussianVertex v1 = new GaussianVertex(0., 1.);
         ConstantDoubleVertex v2 = new ConstantDoubleVertex(0.1);
         DoubleVertex v3 = v1.plus(v2);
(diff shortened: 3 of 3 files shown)


keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/KLDivergence.java
```diff
@@ -0,0 +1,24 @@
...
+import io.improbable.keanu.algorithms.NetworkSamples;
+import io.improbable.keanu.vertices.ProposalVertex;
+
+public class KLDivergence {
+    public static double compute(ProposalVertex q, NetworkSamples samples) {
+        double divergence = 0;
+
+        for (int i = 0; i < samples.size(); i++) {
+            double pLogProb = samples.getLogOfMasterP(i);
+            double qLogProb = q.getLogOfMasterP(samples.getSample(i));
```
gordoncaleb: I think NetworkSamples should return something more structured than just a list of objects. Maybe a NetworkState that still has the ability to figure out which object maps to which vertex?

keanu-project/src/test/java/io/improbable/keanu/algorithms/variational/KLDivergenceTest.java
```diff
@@ -0,0 +1,31 @@
...
+import org.junit.Test;
+
+import java.util.Collections;
+
+public class KLDivergenceTest {
+    @Test
+    public void testKLDivergence() {
+        GaussianVertex v1 = new GaussianVertex(0., 1.);
+        ConstantDoubleVertex v2 = new ConstantDoubleVertex(0.1);
+        AdditionVertex v3 = new AdditionVertex(v1, v2);
```
gordoncaleb: `v3 = v1.plus(v2);`

keanu-project/src/test/java/io/improbable/keanu/algorithms/variational/KLDivergenceTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+        BayesianNetwork network = new BayesianNetwork(v3.getConnectedGraph());
+        NetworkSamples samples = MetropolisHastings
+            .withDefaultConfig()
+            .getPosteriorSamples(network, Collections.singletonList(v1), 1000);
+
+        ProposalVertex q = new GaussianVertex(0.1, 1.);
+
+        assertTrue(KLDivergence.compute(q, samples) < 3.0);
```
gordoncaleb: This assertion is a bit arbitrary. Do we know what the KLDiv for these two distributions should be?

gordoncaleb: Maybe you can calculate the KLDiv for Gaussian(0.1, 1.0) and Gaussian(3.0, 1.0) and assert that the first is less than the second?

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/KLDivergence.java
```diff
@@ -0,0 +1,24 @@
...
+
+            if (pLogProb != Double.NEGATIVE_INFINITY) {
+                if (qLogProb == Double.NEGATIVE_INFINITY) {
+                    throw new IllegalArgumentException("Q cannot have smaller support than P");
```
migwellian: It might be worth including a link in the error message to a page that explains what "support" means. Unless we think that this is the kind of knowledge that all our users will have?

migwellian: You should capture this in a test case. Also the converse, i.e. it's OK for P to have a smaller support than Q.

keanu-project/src/test/java/io/improbable/keanu/algorithms/variational/KLDivergenceTest.java
```diff
@@ -0,0 +1,31 @@
...
+import io.improbable.keanu.vertices.ProposalVertex;
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.ConstantDoubleVertex;
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.operators.binary.AdditionVertex;
+import io.improbable.keanu.vertices.dbl.probabilistic.GaussianVertex;
+import static org.junit.Assert.assertTrue;
+import org.junit.Test;
+
+import java.util.Collections;
+
+public class KLDivergenceTest {
```
migwellian: Can you add a test that the divergence is zero when P == Q?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/GaussianVertex.java
```diff
@@ -21,7 +23,7 @@
 import io.improbable.keanu.vertices.dbl.nonprobabilistic.diff.DualNumber;
 import io.improbable.keanu.vertices.dbl.nonprobabilistic.diff.PartialDerivatives;
 
-public class GaussianVertex extends DoubleVertex implements ProbabilisticDouble {
+public class GaussianVertex extends DoubleVertex implements ProbabilisticDouble, ProposalVertex {
```
gordoncaleb: I like that the GaussianVertex can be used easily with KLDivergence but ideally this feature would be decoupled from the GaussianVertex itself. Can we add a `compute(ProbabilisticDouble q, NetworkSamples samples)` that does the wrapping inside the KLDivergence class?

keanu-project/src/main/java/io/improbable/keanu/vertices/ProposalVertex.java
```diff
@@ -0,0 +1,7 @@
+package io.improbable.keanu.vertices;
+
+import java.util.List;
+
+public interface ProposalVertex {
```
migwellian: Unless I've misunderstood something, it feels like it would be better to parametrise this class by `<T extends Tensor>`, and expect `List<T>` instead of `List<?>` (where the `T` will be the same type that the concrete Vertex class is parametrised on). 

That avoids the need for a yucky cast to `DoubleTensor` in your implementation of the method in `GaussianVertex`.

gordoncaleb: The proposal might not be a vertex. It might be a distribution represented by a bayesian network OR just a function that represents a distribution OR a single vertex that represents a distribution. Considering this interface is for KLDivergence and within the context of KL is very commonly called Q, maybe we could call it QDistribution and keep it in the same package as the KLDivergence class?

keanu-project/src/test/java/io/improbable/keanu/algorithms/variational/KLDivergenceTest.java
```diff
@@ -0,0 +1,31 @@
...
+    @Test
+    public void testKLDivergence() {
```
gordoncaleb: JUnit isn't great at helping developers describe the intention of their test but ideally the method name is the test name and the test name should describe what the assertion in the test is. This also encourages single assertions in tests which is good practice.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/GaussianVertex.java
```diff
@@ -130,4 +132,12 @@ public DualNumber calculateDualNumber(Map<Vertex, DualNumber> dualNumbers) {
             return DualNumber.createWithRespectToSelf(getId(), getValue());
         }
     }
+
+    @Override
+    public double getLogOfMasterP(List<?> values) {
+        if (values.size() != 1) {
+            throw new IllegalArgumentException("Univariate distributions takes exactly 1 value and you supplied: " + values.size());
+        }
+        return logProb((DoubleTensor) values.get(0));
```
migwellian: You could replace this with a one-liner:

    return logProb((DoubleTensor) Iterables.getOnlyElement(values));

Which gives you a decent exception:

    java.lang.IllegalArgumentException: expected one element but was: <1.0, 1.0>

	at com.google.common.collect.Iterators.getOnlyElement(Iterators.java:322)
	at com.google.common.collect.Iterables.getOnlyElement(Iterables.java:294)
	at io.improbable.keanu.vertices.dbl.probabilistic.GaussianVertex.getLogOfMasterP(GaussianVertex.java:140)

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/GaussianVertex.java
```diff
@@ -130,4 +132,12 @@ public DualNumber calculateDualNumber(Map<Vertex, DualNumber> dualNumbers) {
             return DualNumber.createWithRespectToSelf(getId(), getValue());
         }
     }
+
+    @Override
+    public double getLogOfMasterP(List<?> values) {
```
migwellian: Could you add test coverage for the case when the list has size > 1?

migwellian:
> "When P is 0, contribution to divergence is 0"

 - can you capture this in a test?

keanu-project/src/test/java/io/improbable/keanu/algorithms/variational/KLDivergenceTest.java
icantsayrural: Theoretically, the divergence should be 0. But we are sampling p here, so it'll just be arbitrarily low.

I like the idea of comparing the divergences between two different Gaussians!

icantsayrural: Although, I see potential for flaky tests ... But I think we will be OK if the two qs are _very_ different.

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/KLDivergence.java
icantsayrural: Wikipedia has a section on support for all distributions, so I think we are OK

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/GaussianVertex.java
icantsayrural: Nice!

keanu-project/src/test/java/io/improbable/keanu/algorithms/variational/KLDivergenceTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+        BayesianNetwork network = new BayesianNetwork(v3.getConnectedGraph());
+        NetworkSamples samples = MetropolisHastings
+            .withDefaultConfig()
+            .getPosteriorSamples(network, Collections.singletonList(v1), 1000);
+
+        ProbabilisticDouble q1 = new GaussianVertex(0.1, 1.);
+        ProbabilisticDouble q2 = new GaussianVertex(10.0, 1.);
+
+        assertTrue(KLDivergence.compute(q1, samples) < KLDivergence.compute(q2, samples));
```
migwellian: It's a good idea to use Hamcrest matchers for asserts. The code reads more fluently and the error messages are much more informative.

For example:

    assertThat(KLDivergence.compute(q1, samples), lessThan(KLDivergence.compute(q2, samples)));

yields messages like:

    java.lang.AssertionError: 
    Expected: a value less than <14325.522530625412>
         but: <1435844.110279129> was greater than <14325.522530625412>

Whereas the `assertTrue` statement only tells you that there's an `java.lang.AssertionError`, with no message.

keanu-project/src/test/java/io/improbable/keanu/algorithms/variational/KLDivergenceTest.java
```diff
@@ -0,0 +1,112 @@
...
+import static org.junit.Assert.assertTrue;
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class KLDivergenceTest {
```
migwellian: Nice tests! They have good names that tell you exactly what they're testing for.

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/QDistribution.java
```diff
@@ -0,0 +1,7 @@
+package io.improbable.keanu.algorithms.variational;
+
+import io.improbable.keanu.network.NetworkState;
+
+public interface QDistribution {
```
migwellian: There are no instantiations of this interface in the code base, except for your TestQDistribution which only returns a constant. Can we add some more real-world examples, and tests to accompany them?

migwellian: Can you add some JavaDoc that explains what a Q Distribution is?

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/KLDivergence.java
```diff
@@ -0,0 +1,43 @@
+package io.improbable.keanu.algorithms.variational;
+
+import com.google.common.collect.Iterables;
+import io.improbable.keanu.algorithms.NetworkSamples;
+import io.improbable.keanu.network.NetworkState;
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+import io.improbable.keanu.vertices.dbl.probabilistic.ProbabilisticDouble;
+
+public class KLDivergence {
```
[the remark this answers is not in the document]
gordoncaleb: 👍 It's a very nice use of a lambda actually. It's a single lambda used effectively as a reduce function. Ace.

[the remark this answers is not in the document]
GeorgeNash: I actually like the idea of using the `QDistribution` interface here so we begin to see it in the code. What're your thoughts on something like:

```
    public static double compute(ProbabilisticDouble q, NetworkSamples samples) {
        return compute(samples, i -> new ProbabilisticDoubleAsQDistribution(q).getLogOfMasterP(samples.getNetworkState(i)));
    }
```

```
    private static class ProbabilisticDoubleAsQDistribution implements QDistribution {

        final ProbabilisticDouble q;

        ProbabilisticDoubleAsQDistribution(ProbabilisticDouble q) {
            this.q = q;
        }

        @Override
        public double getLogOfMasterP(NetworkState state) {
            DoubleTensor vertexValue = state.get(Iterables.getOnlyElement(state.getVertexIds()));
            return q.logProb(vertexValue);
        }
    }
```

keanu-project/src/test/java/io/improbable/keanu/algorithms/variational/KLDivergenceTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+    private NetworkSamples createNetworkSamplesWithOneVertexAndOneSample(double p) {
+        Map<VertexId, List<DoubleTensor>> samplesByVertex = new HashMap<>();
+        samplesByVertex.put(new VertexId(), Collections.singletonList(DoubleTensor.scalar(1.)));
+        List<Double> logOfMasterPForEachSample = Collections.singletonList(p);
+
+        return new NetworkSamples(samplesByVertex, logOfMasterPForEachSample, 1);
+    }
+
+    private class TestGaussianQDistribution extends GaussianVertex implements QDistribution {
```
migwellian: Nice

keanu-project/src/test/java/io/improbable/keanu/algorithms/variational/KLDivergenceTest.java
```diff
@@ -0,0 +1,137 @@
...
+        QDistribution qDist = mock(QDistribution.class);
+        when(qDist.getLogOfMasterP(any(NetworkState.class))).thenReturn(identicalProb);
+        NetworkSamples samples = createNetworkSamplesWithOneVertexAndOneSample(identicalProb);
+
+        assertEquals(0., KLDivergence.compute(qDist, samples), 1e-6);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void throwsExceptionIfPProbIsNotZeroButQIs() {
+        QDistribution q = mock(QDistribution.class);
```
migwellian: you can pull out `q` as a field of the class

keanu-project/src/test/java/io/improbable/keanu/algorithms/variational/KLDivergenceTest.java
```diff
@@ -0,0 +1,137 @@
...
+        when(q.getLogOfMasterP(any(NetworkState.class))).thenReturn(Double.NEGATIVE_INFINITY);
+        NetworkSamples samples = createNetworkSamplesWithOneVertexAndOneSample(Math.log(0.5));
+
+        KLDivergence.compute(q, samples);
+    }
+
+    @Test
+    public void klDivergenceReturnsZeroIfPIsZeroButQIsNot() {
+        QDistribution q = mock(QDistribution.class);
+        when(q.getLogOfMasterP(any(NetworkState.class))).thenReturn(Math.log(0.5));
```
migwellian: you could pull this out into a method:

    private QDistribution mockThatReturnsConstantValue(double identicalProb) {
        QDistribution q = mock(QDistribution.class);
        when(q.getLogOfMasterP(any(NetworkState.class))).thenReturn(identicalProb);
        return q;
    }

keanu-project/src/test/java/io/improbable/keanu/algorithms/variational/KLDivergenceTest.java
```diff
@@ -0,0 +1,137 @@
...
+            .getPosteriorSamples(network, Collections.singletonList(v1), 1000);
+
+        QDistribution q1 = new TestGaussianQDistribution(0.1, 1.);
+        QDistribution q2 = new TestGaussianQDistribution(10.0, 1.);
+
+        assertThat(KLDivergence.compute(q1, samples), lessThan(KLDivergence.compute(q2, samples)));
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void throwsExceptionIfNetworkStateHasMoreThanOneVertexAndQIsProbabilisticDouble(){
```
migwellian: I think this could do with some explanation in a javadoc comment. Why is this a bad thing to do? Is this just something that's not yet supported, or is it a logical mistake on the user's part? If the latter - what should the user do instead?

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/KLDivergence.java
```diff
@@ -0,0 +1,39 @@
...
+
+        for (int i = 0; i < samples.size(); i++) {
+            double pLogProb = samples.getLogOfMasterP(i);
+
+            NetworkState state = samples.getNetworkState(i);
+            double qLogProb = qLogProbCalculator.apply(state);
+
+            if (pLogProb != Double.NEGATIVE_INFINITY) {
+                if (qLogProb == Double.NEGATIVE_INFINITY) {
+                    throw new IllegalArgumentException("Q cannot have smaller support than P");
```
migwellian: not tested

keanu-project/src/test/java/io/improbable/keanu/algorithms/variational/KLDivergenceTest.java
migwellian: A neat library you can use here is https://junit.org/junit4/javadoc/latest/org/junit/rules/ExpectedException.html#expectMessage(java.lang.String)

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/KLDivergence.java
icantsayrural: It is, but as discussed I'll use the library to expect a message to be more clear in my tests.

PR #218, opened 2018-09-05
Add split for double tensors

gordoncaleb:
This PR adds a split that is nice and will be required by reverse mode autodiff for concat. 

From the JavaDoc:

```
    /**
     * @param dimension      the dimension to split on
     * @param splitAtIndices the indices that the dimension to split on should be split on
     * @return pieces of the tensor split in the order specified by splitAtIndices. To get
     * pieces that encompasses the entire tensor, the last index in the splitAtIndices must
     * be the length of the dimension being split on.
     * <p>
     * e.g A =
     * [
     * 1, 2, 3, 4, 5, 6
     * 7, 8, 9, 1, 2, 3
     * ]
     * <p>
     * A.split(0, [1]) gives List([1, 2, 3, 4, 5, 6])
     * A.split(0, [1, 2] gives List([1, 2, 3, 4, 5, 6], [7, 8, 9, 1, 2, 3]
     * <p>
     * A.split(1, [1, 3, 6]) gives
     * List(
     * [1, [2, 3  , [4, 5, 6,
     * 7]   8, 9]    1, 2, 3]
     * )
     */
```

This is only for DoubleTensor and not the other tensors because it's currently only needed for autodiff, which is double only. We may add one for Integer and Boolean in future.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
@@ -145,7 +145,7 @@ public class TensorShape {
         return dims;
     }
 
-    public static int[] moveAxis(int from, int to, int[] shape) {
+    public static int[] slideDimension(int from, int to, int[] shape) {
         List<Integer> shapeList = new ArrayList<>(Ints.asList(shape));
         Integer dimLength = shapeList.remove(from);
         shapeList.add(to, dimLength);
--- a/keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
@@ -850,7 +850,7 @@ public class Nd4jDoubleTensor implements DoubleTensor {
      * ]
      * <p>
      * A.split(0, [1]) gives List([1, 2, 3, 4, 5, 6])
-     * A.split(0, [1, 2] gives List([1, 2, 3, 4, 5, 6], [7, 8, 9, 1, 2, 3]
+     * A.split(0, [1, 2]) gives List([1, 2, 3, 4, 5, 6], [7, 8, 9, 1, 2, 3]
      * <p>
      * A.split(1, [1, 3, 6]) gives
      * List(
@@ -864,8 +864,8 @@ public class Nd4jDoubleTensor implements DoubleTensor {
         DoubleTensor duplicateTensor = this.duplicate();
         int[] tensorShape = duplicateTensor.getShape();
         int[] dimensionRange = TensorShape.dimensionRange(0, tensorShape.length);
-        int[] moveDimToZero = TensorShape.moveAxis(dimension, 0, dimensionRange);
-        int[] moveZeroToDim = TensorShape.moveAxis(0, dimension, dimensionRange);
+        int[] moveDimToZero = TensorShape.slideDimension(dimension, 0, dimensionRange);
+        int[] moveZeroToDim = TensorShape.slideDimension(0, dimension, dimensionRange);
 
         DoubleTensor permutedTensor = duplicateTensor.permute(moveDimToZero).reshape(1, (int) duplicateTensor.getLength());
 
@@ -873,33 +873,29 @@ public class Nd4jDoubleTensor implements DoubleTensor {
 
         List<DoubleTensor> splitTensor = new ArrayList<>();
 
-        int previousIndex = 0;
+        int previousSplitAtIndex = 0;
         int rawBufferPosition = 0;
-        for (int index : splitAtIndices) {
+        for (int splitAtIndex : splitAtIndices) {
 
             int[] subTensorShape = Arrays.copyOf(tensorShape, tensorShape.length);
-            int subTensorLengthInDimension = index - previousIndex;
+            int subTensorLengthInDimension = splitAtIndex - previousSplitAtIndex;
 
-            if (subTensorLengthInDimension > tensorShape[dimension] || subTensorLengthInDimension < 0) {
-                throw new IllegalArgumentException("Invalid index to split on " + index + " at " + dimension + " for tensor of shape " + Arrays.toString(tensorShape));
+            if (subTensorLengthInDimension > tensorShape[dimension] || subTensorLengthInDimension <= 0) {
+                throw new IllegalArgumentException("Invalid index to split on " + splitAtIndex + " at " + dimension + " for tensor of shape " + Arrays.toString(tensorShape));
             }
 
             subTensorShape[dimension] = subTensorLengthInDimension;
-            previousIndex = index;
-
             int subTensorLength = (int) TensorShape.getLength(subTensorShape);
 
             double[] buffer = new double[subTensorLength];
             System.arraycopy(rawBuffer, rawBufferPosition, buffer, 0, buffer.length);
 
-            int[] subTensorPermutedShape = TensorShape.moveAxis(dimension, 0, subTensorShape);
-
-            DoubleTensor subTensor = DoubleTensor.create(buffer, new int[]{1, subTensorLength}).reshape(subTensorPermutedShape).permute(moveZeroToDim);
-
+            int[] subTensorPermutedShape = TensorShape.slideDimension(dimension, 0, subTensorShape);
+            DoubleTensor subTensor = DoubleTensor.create(buffer, subTensorPermutedShape).permute(moveZeroToDim);
             splitTensor.add(subTensor);
 
+            previousSplitAtIndex = splitAtIndex;
             rawBufferPosition += buffer.length;
-
         }
 
         return splitTensor;
--- a/keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
@@ -21,6 +21,8 @@ import org.nd4j.linalg.api.ops.impl.transforms.comparison.OldGreaterThanOrEqual;
 import org.nd4j.linalg.api.ops.impl.transforms.comparison.OldLessThan;
 import org.nd4j.linalg.api.ops.impl.transforms.comparison.OldLessThanOrEqual;
 import org.nd4j.linalg.factory.Nd4j;
+import org.nd4j.linalg.indexing.INDArrayIndex;
+import org.nd4j.linalg.indexing.NDArrayIndex;
 import org.nd4j.linalg.indexing.conditions.Conditions;
 import org.nd4j.linalg.inverse.InvertMatrix;
 import org.nd4j.linalg.ops.transforms.Transforms;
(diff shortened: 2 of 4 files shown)


keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -833,6 +836,75 @@ public DoubleTensor concat(int dimension, DoubleTensor... those) {
...
+     * be the length of the dimension being split on.
+     * <p>
+     * e.g A =
+     * [
+     * 1, 2, 3, 4, 5, 6
+     * 7, 8, 9, 1, 2, 3
+     * ]
+     * <p>
+     * A.split(0, [1]) gives List([1, 2, 3, 4, 5, 6])
+     * A.split(0, [1, 2] gives List([1, 2, 3, 4, 5, 6], [7, 8, 9, 1, 2, 3]
```
icantsayrural: nit: missing closing bracket for split and List

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
```diff
@@ -141,6 +145,13 @@ public static boolean isScalar(int[] shape) {
         return dims;
     }
 
+    public static int[] moveAxis(int from, int to, int[] shape) {
```
GeorgeNash: I know you usually use this to move axis by passing zero as one of the parameters but it could really be used for all sorts. A better name is probably: `moveDimensions`. Or changing this to only take a `from` parameter and assume `to` is always `0`

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -833,6 +836,75 @@ public DoubleTensor concat(int dimension, DoubleTensor... those) {
...
+     * <p>
+     * A.split(1, [1, 3, 6]) gives
+     * List(
```
GeorgeNash: Can you change the formatting a little:

```
     * List(
     * [1 , [2, 3  , [4, 5, 6,
     *  7]   8, 9]    1, 2, 3]
     * )
```

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -833,6 +836,75 @@ public DoubleTensor concat(int dimension, DoubleTensor... those) {
...
+
+            int[] subTensorShape = Arrays.copyOf(tensorShape, tensorShape.length);
+            int subTensorLengthInDimension = index - previousIndex;
+
+            if (subTensorLengthInDimension > tensorShape[dimension] || subTensorLengthInDimension < 0) {
+                throw new IllegalArgumentException("Invalid index to split on " + index + " at " + dimension + " for tensor of shape " + Arrays.toString(tensorShape));
+            }
+
+            subTensorShape[dimension] = subTensorLengthInDimension;
+            previousIndex = index;
```
icantsayrural: nit: group all the updates at the end of the loop (i.e. with `rawBufferPosition`)

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -833,6 +836,75 @@ public DoubleTensor concat(int dimension, DoubleTensor... those) {
...
+    @Override
+    public List<DoubleTensor> split(int dimension, int[] splitAtIndices) {
+
+        DoubleTensor duplicateTensor = this.duplicate();
+        int[] tensorShape = duplicateTensor.getShape();
+        int[] dimensionRange = TensorShape.dimensionRange(0, tensorShape.length);
+        int[] moveDimToZero = TensorShape.moveAxis(dimension, 0, dimensionRange);
+        int[] moveZeroToDim = TensorShape.moveAxis(0, dimension, dimensionRange);
+
+        DoubleTensor permutedTensor = duplicateTensor.permute(moveDimToZero).reshape(1, (int) duplicateTensor.getLength());
```
GeorgeNash: Would be cool to pull that flatten out into a function:

```
public DoubleTensor flatten() {
        return this.duplicate().reshape(1, (int) getLength());
    }
```

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -833,6 +836,75 @@ public DoubleTensor concat(int dimension, DoubleTensor... those) {
...
+            previousIndex = index;
+
+            int subTensorLength = (int) TensorShape.getLength(subTensorShape);
+
+            double[] buffer = new double[subTensorLength];
+            System.arraycopy(rawBuffer, rawBufferPosition, buffer, 0, buffer.length);
+
+            int[] subTensorPermutedShape = TensorShape.moveAxis(dimension, 0, subTensorShape);
+
+            DoubleTensor subTensor = DoubleTensor.create(buffer, new int[]{1, subTensorLength}).reshape(subTensorPermutedShape).permute(moveZeroToDim);
```
GeorgeNash: What a naughty tensor line

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/ScalarDoubleTensor.java
```diff
@@ -426,6 +428,11 @@ public DoubleTensor concat(int dimension, DoubleTensor... those) {
         return Nd4jDoubleTensor.scalar(value).concat(dimension, those);
     }
 
+    @Override
+    public List<DoubleTensor> split(int dimension, int[] indices) {
```
icantsayrural: `int[] splitAtIndices`

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
icantsayrural: If a user passes 0 as an index, then `DoubleTensor.create` will throw an exception because `subTensorLength` will be 0. I think we should catch this earlier on above?

gordoncaleb: I've updated the check

` if (subTensorLengthInDimension > tensorShape[dimension] || subTensorLengthInDimension < 0) {` to have `subTensorLengthInDimension <= 0` which should catch any split at index that causes a zero length tensor.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
gordoncaleb: Nice eye.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
gordoncaleb: Done. Renamed a few things to make this loop a bit clearer.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/ScalarDoubleTensor.java
gordoncaleb: Yup.

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
gordoncaleb: Renamed to `slideDimension`

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
gordoncaleb: It's only used here and I feel like there are some semantics to define around flatten. Like flatten to column OR row OR to a vector (rank 1). With that in mind I'd say do it in another PR if we find it useful somewhere else.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
gordoncaleb: It's a big one.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
gordoncaleb: Done.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -833,6 +830,70 @@ public DoubleTensor concat(int dimension, DoubleTensor... those) {
...
+
+        Nd4j.getCompressor().autoDecompress(tensor);
+
+        List<DoubleTensor> splits = new ArrayList<>();
+        int previousSplitIndex = 0;
+        for (int i = 0; i < splitAtIndices.length; i++) {
+
+            INDArrayIndex[] indices = new INDArrayIndex[tensor.rank()];
+
+            if(previousSplitIndex == splitAtIndices[i]){
```
icantsayrural: nit: space

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -833,6 +830,70 @@ public DoubleTensor concat(int dimension, DoubleTensor... those) {
...
+        int[] shape = getShape();
+        if (dimension < 0 || dimension >= shape.length) {
+            throw new IllegalArgumentException("Invalid dimension to split on " + dimension);
+        }
+
+        Nd4j.getCompressor().autoDecompress(tensor);
+
+        List<DoubleTensor> splits = new ArrayList<>();
+        int previousSplitIndex = 0;
+        for (int i = 0; i < splitAtIndices.length; i++) {
```
icantsayrural: We could use a foreach loop

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -833,6 +830,70 @@ public DoubleTensor concat(int dimension, DoubleTensor... those) {
...
+     */
+    @Override
+    public List<DoubleTensor> split(int dimension, int[] splitAtIndices) {
+
+        int[] shape = getShape();
+        if (dimension < 0 || dimension >= shape.length) {
+            throw new IllegalArgumentException("Invalid dimension to split on " + dimension);
+        }
+
+        Nd4j.getCompressor().autoDecompress(tensor);
```
icantsayrural: What does this do?

gordoncaleb: Nd4j sometimes compresses the tensor. This makes sure it's operable.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
gordoncaleb: I thought it was easier to think about as iterating through the number of splits rather than iterating over the places to split.

PR #221, opened 2018-09-06
Expose vertex id

GeorgeNash:
When debugging it's a massive pain to access things if I don't have access to a vertex as I'm unable to create a `VertexID` on the fly. This let's you make a `VertexID`.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
@@ -24,6 +24,10 @@ public class VertexId implements Comparable<VertexId> {
         idValues[0] = ID_GENERATOR.getAndIncrement();
     }
 
+    public VertexId(long id) {
+        idValues[0] = id;
+    }
+
     @Override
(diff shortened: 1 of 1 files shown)


keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
```diff
@@ -24,6 +24,10 @@ public VertexId() {
         idValues[0] = ID_GENERATOR.getAndIncrement();
     }
 
+    public VertexId(long id) {
```
christophernorth: Should we expose this publicly if it's just for debug purposes?

gordoncaleb: I think it's useful for cases where you need to access data from a Map keyed by a VertexID (e.g. PartialDerivatives) and didn't have access to the actual VertexId object that lives on the Vertex. That effectively forces you to have the Vertex if you want to access the Map. This would also be needed if we were deserialising a graph with ids (which we don't currently do but yeah).

PR #225, opened 2018-09-07
added the concept of a namespace (list of strings) to VertexLabel

migwellian:
This will be needed for Plates: every Vertex in the Plate will have a VertexLabel, but the names in a BayesNet need to be unique. So we will scope the VertexLabel with the ID of the Plate.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
@@ -9,7 +9,6 @@ import com.google.common.collect.Lists;
 
 public class VertexLabel {
     private static final char NAMESPACE_SEPARATOR = '.';
-    private static final char NAME_PREPENDER = ':';
     private final String name;
     private final List<String> namespace;
 
@@ -39,7 +38,7 @@ public class VertexLabel {
     @Override
     public String toString() {
         String namespaceString = Joiner.on(NAMESPACE_SEPARATOR).join(Lists.reverse(namespace));
-        return String.format("%s%c%s", namespaceString, NAME_PREPENDER, name);
+        return String.format("%s%c%s", namespaceString, NAMESPACE_SEPARATOR, name);
     }
 
     @Override
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
@@ -37,8 +37,8 @@ public class VertexLabel {
 
     @Override
     public String toString() {
-        String namespaceString = Joiner.on(NAMESPACE_SEPARATOR).join(Lists.reverse(namespace));
-        return String.format("%s%c%s", namespaceString, NAMESPACE_SEPARATOR, name);
+        ImmutableList<String> names = ImmutableList.<String>builder().add(name).addAll(namespace).build();
+        return Joiner.on(NAMESPACE_SEPARATOR).join(Lists.reverse(names));
     }
 
     @Override
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
@@ -1,17 +1,44 @@
 package io.improbable.keanu.vertices;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import java.util.Objects;
 
-public class VertexLabel {
-    private final String label;
+import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableList;
 
-    public VertexLabel(String label) {
-        this.label = label;
+public class VertexLabel {
+    private static final char NAMESPACE_SEPARATOR = '.';
+    private static final char NAME_PREPENDER = ':';
+    private final String name;
+    private final List<String> namespace;
+
+    public VertexLabel(String name, String... namespace) {
+        this.name = name;
+        this.namespace = ImmutableList.copyOf(namespace);
+    }
+
+    public VertexLabel inNamespace(String topLevelNamespace) {
+        List<String> newNamespace = ImmutableList.<String>builder().addAll(namespace).add(topLevelNamespace).build();
+        return new VertexLabel(this.name, newNamespace.toArray(new String[0]));
+    }
+
+    public VertexLabel removeOuterNamespace() throws VertexLabelException {
+        try {
+            List<String> reducedNamespace = namespace.subList(0, namespace.size()-1);
+            return new VertexLabel(this.name, reducedNamespace.toArray(new String[0]));
+        } catch (IndexOutOfBoundsException e) {
+            throw new VertexLabelException("There is no namespace to remove", e);
+        }
     }
 
     @Override
     public String toString() {
-        return label;
+        ArrayList reversedNamespace = new ArrayList(namespace);
+        Collections.reverse(reversedNamespace);
+        String namespaceString = Joiner.on(NAMESPACE_SEPARATOR).join(reversedNamespace);
+        return String.format("%s%c%s", namespaceString, NAME_PREPENDER, name);
     }
 
     @Override
@@ -19,12 +46,13 @@ public class VertexLabel {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         VertexLabel that = (VertexLabel) o;
-        return Objects.equals(label, that.label);
+        return Objects.equals(name, that.name) &&
+            Objects.equals(namespace, that.namespace);
     }
 
     @Override
     public int hashCode() {
 
-        return Objects.hash(label);
+        return Objects.hash(name, namespace);
     }
 }
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
@@ -15,19 +15,23 @@ public class VertexLabel {
     private final List<String> namespace;
 
     public VertexLabel(String name, String... namespace) {
+        this(name, ImmutableList.copyOf(namespace));
+    }
+
+    public VertexLabel(String name, List<String> namespace) {
         this.name = name;
         this.namespace = ImmutableList.copyOf(namespace);
     }
 
     public VertexLabel inNamespace(String topLevelNamespace) {
         List<String> newNamespace = ImmutableList.<String>builder().addAll(namespace).add(topLevelNamespace).build();
-        return new VertexLabel(this.name, newNamespace.toArray(new String[0]));
+        return new VertexLabel(this.name, newNamespace);
     }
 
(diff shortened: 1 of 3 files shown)


keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
```diff
@@ -1,30 +1,58 @@
...
+    private static final char NAME_PREPENDER = ':';
+    private final String name;
+    private final List<String> namespace;
+
+    public VertexLabel(String name, String... namespace) {
+        this.name = name;
+        this.namespace = ImmutableList.copyOf(namespace);
+    }
+
+    public VertexLabel inNamespace(String topLevelNamespace) {
```
christophernorth: I see what this does, but I wonder if the name is slightly confusing (when doing my first skim read through I thought this was some form of matching function rather than something that put on a new layer of name-spacing).

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
```diff
@@ -1,30 +1,58 @@
...
+        List<String> newNamespace = ImmutableList.<String>builder().addAll(namespace).add(topLevelNamespace).build();
+        return new VertexLabel(this.name, newNamespace.toArray(new String[0]));
```
christophernorth: Totally not your doing - but I'm not loving this instance of the toArray API - doesn't seem ideal to have to create an array for no reason!

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
```diff
@@ -1,30 +1,58 @@
...
+import java.util.Collections;
+import java.util.List;
 import java.util.Objects;
 
+import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableList;
+
 public class VertexLabel {
-    private final String label;
+    private static final char NAMESPACE_SEPARATOR = '.';
```
christophernorth: Should we use '/' as Charlie requested?  I don't have strong feelings tbh.

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
```diff
@@ -1,30 +1,58 @@
...
+    private static final char NAME_PREPENDER = ':';
```
christophernorth: Is there a reason you've used a different value for the two separators?  If we went with something more "file system like" then just using the same '/' char would work for both perhaps?

christophernorth: Obviously not something we care about too much at the moment, but it's definitely the case that when we come to read this stuff in we'll want to avoid our separators being part of the actual identifiers in the namespace.  It might be worth checking for these things now to ensure that any users of this functionality don't hit issues down the line if they use an invalid character?

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
```diff
@@ -1,30 +1,58 @@
...
+    }
 
-    public VertexLabel(String label) {
-        this.label = label;
+    public VertexLabel removeOuterNamespace() throws VertexLabelException {
+        try {
+            List<String> reducedNamespace = namespace.subList(0, namespace.size()-1);
+            return new VertexLabel(this.name, reducedNamespace.toArray(new String[0]));
+        } catch (IndexOutOfBoundsException e) {
+            throw new VertexLabelException("There is no namespace to remove", e);
```
christophernorth: Clean!

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
```diff
@@ -1,30 +1,58 @@
...
+        }
     }
 
     @Override
     public String toString() {
-        return label;
+        ArrayList reversedNamespace = new ArrayList(namespace);
```
christophernorth: Could we use Lists.reverse() here to avoid having to create a copy and having to reverse in place?

keanu-project/src/test/java/io/improbable/keanu/vertices/VertexLabelTest.java
```diff
@@ -0,0 +1,99 @@
...
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.not;
+
+import org.junit.Test;
+
+public class VertexLabelTest {
+    @Test
+    public void byDefaultALabelHasNoNamespace() {
+        VertexLabel foo = new VertexLabel("foo");
+        assertThat(foo.toString(), equalTo(":foo"));
```
christophernorth: Do we always want that special character in place?

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
migwellian: Got any suggestions for a better name?

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
migwellian: no problem, I'll just add a constructor that takes a List

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
migwellian: It's no big deal since it's only used in `toString`.

I chose `.` because it helps with the intuition that it's similar to a Java namespace.

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
migwellian: No major reason other than to distinguish more clearly between the namespace and the name.

I have no strong opinion, and these are only used in `toString`.

migwellian: these constants are only used in `toString`

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
migwellian: much better - thanks.

keanu-project/src/test/java/io/improbable/keanu/vertices/VertexLabelTest.java
migwellian: I have no strong opinion about that.

migwellian:
Thanks for your feedback @christophernorth - I've pushed 2 changes.

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
christophernorth: Yeah, that doesn't mean it couldn't lead to confusion though.

If, say, a Keanu user started using Vertex Labels (which they will if they use Graphception) then they could very well accidentally use a character we've decided to use as a namespace separator meaning both:
A) We could get confusing toString output (which could be important for both internal debug and potential visualisation/editor tooling).
B) If we want to save & load Vertices to disk then  either we make their existing Vertex Labels invalid, or we need to use a more complex encoding scheme for names when storing them to disk.

So, not stopping them using the separator characters now could well cause them headaches/confusion in future if we add extra functionality.  Seems worth the small amount of extra code to me for now.

Thinking more deeply on this, there are other potential characters that would cause issues for case B - (eg special characters for JSON) - so longer term we will probably want to vet any string they pass, but perhaps we just need to raise a new ticket for this.

christophernorth: Our toString() could well be useful for other stuff though - not least of which debug could be an important consumer

keanu-project/src/test/java/io/improbable/keanu/vertices/VertexLabelTest.java
christophernorth: I think from the PoV of a user - they'd expect to see the label as they put it in if they aren't using any namespace functionality?

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
christophernorth: A tricky one :)

perhaps addNamespace or plusNamespace or similar?

christophernorth: Or encapsulate()?

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
```diff
@@ -1,30 +1,59 @@
...
+    public VertexLabel inNamespace(String topLevelNamespace) {
+        List<String> newNamespace = ImmutableList.<String>builder().addAll(namespace).add(topLevelNamespace).build();
+        return new VertexLabel(this.name, newNamespace);
+    }
 
-    public VertexLabel(String label) {
-        this.label = label;
+    public VertexLabel removeOuterNamespace() throws VertexLabelException {
+        try {
+            List<String> reducedNamespace = namespace.subList(0, namespace.size()-1);
```
GeorgeNash: nit: `namespace.subList(0, namespace.size() - 1);`

keanu-project/src/test/java/io/improbable/keanu/vertices/VertexLabelTest.java
migwellian: fair enough

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
migwellian: ok, so in conclusion shall we go with:
* remove `NAME_PREPENDER` and use `NAMESPACE_SEPARATOR` in its place
* set `NAMESPACE_SEPARATOR` to be `/` instead of `.`
* throw if the user passes in a string containing `NAMESPACE_SEPARATOR `

christophernorth: That sounds perfect to me - thanks for thinking about this!

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
migwellian: In order to indicate that they are non-mutating, how about `withExtraNamespace(String)` and `withoutOuterNamespace()`?

keanu-project/src/test/java/io/improbable/keanu/vertices/VertexLabelTest.java
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
migwellian: done

PR #232, opened 2018-09-10
Fix bug in double if vertex

GeorgeNash:
- If Vertex assumed you were performing it on a tensor of rank 2, this fix enables it to work for all ranks

### DIFF
--- a/keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/DoubleIfVertexTest.java
+++ b/keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/DoubleIfVertexTest.java
@@ -171,7 +171,7 @@ public class DoubleIfVertexTest {
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/DoubleIfVertexTest.java
```diff
@@ -170,6 +170,40 @@ public void canExtractDualNumberFromMixedPredicateWithDifferentParentsAndFillInZ
         Assert.assertArrayEquals(dFdd.getShape(), dIfdD.getShape());
     }
 
+    @Test
+    public void canExtractDualNumberFromMixedPredicateWithDifferentParentsAndFillInZeroesRankThree() {
```
christophernorth: What a name!

Could we perhaps cut it down a little bit though?  It's a bit of a mouthful for a test that is just checking that the shapes are correct?

GeorgeNash: Ok - made it a little shorter!!

PR #235, opened 2018-09-13
Added a catch for the Not Strictly Positive exception that we then te…

christophernorth:
Add a temporary translation between Exception types with, hopefully, a more informative message in the case that we hit the appveyor issue moving forward.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/gradient/GradientOptimizer.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/gradient/GradientOptimizer.java
@@ -8,6 +8,7 @@ import java.util.Collection;
 import java.util.List;
 import java.util.function.BiConsumer;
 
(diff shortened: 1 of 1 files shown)


keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/gradient/GradientOptimizer.java
```diff
@@ -179,4 +187,13 @@ private static void warnIfGradientIsFlat(double[] gradient) {
             throw new IllegalStateException("The initial gradient is very flat. The largest gradient is " + maxGradient);
         }
     }
+
+    private static void translateNSPExceptionToIllegalArgException(NotStrictlyPositiveException e) {
+        long bitwiseMinValue = Double.doubleToLongBits(e.getMin().doubleValue());
```
gordoncaleb: Does the code that creates this exception actually populate these values?

PR #236, opened 2018-09-13
Fixed findbug errors about storing/returning references to arrays:

migwellian:
May expose internal representation by returning reference to mutable object
May expose internal representation by incorporating reference to mutable object


migwellian:
@gordoncaleb any concerns around performance impact?

PR #238, opened 2018-09-14
Feature/0log0

migwellian:
The issue is described in https://improbableio.atlassian.net/browse/SOL-2250

In short: 0log0 == NaN but [should equal 0](https://arcsecond.wordpress.com/2009/03/19/0log0-0-for-real/).

* New classes that (1) check values in a tensor (2) handle unexpected values either by throwing or changing the value.
* For Multinomial, I decided that the right fix was to disallow zero-valued probabilities
* For other distributions, we will need to decide whether to disallow zeros or to fix NaNs on the fly


keanu-project/src/main/java/io/improbable/keanu/tensor/validate/TensorValidator.java
```diff
@@ -0,0 +1,48 @@
...
+import io.improbable.keanu.tensor.Tensor;
+import io.improbable.keanu.tensor.bool.BooleanTensor;
+import io.improbable.keanu.tensor.validate.check.CustomTensorValueChecker;
+import io.improbable.keanu.tensor.validate.check.TensorValueChecker;
+import io.improbable.keanu.tensor.validate.check.TensorValueNotEqualsCheck;
+import io.improbable.keanu.tensor.validate.policy.TensorValidationPolicy;
+
+public class TensorValidator<U, T extends Tensor<U>> implements TensorValueChecker<T> {
+
+    public static <U, T extends Tensor<U>> TensorValidator<U, T> thatChecksFor(U v) {
```
GeorgeNash: I found this really confusing! When I first read the tests I assumed that `thatChecksFor` was checking that they existed when in face it checks its not equal. Can we rename it to `checkValueIsNotPresent` or `checksIsNotValue` or something like that?

keanu-project/src/test/java/io/improbable/keanu/tensor/intgr/Nd4jIntegerTensorTest.java
```diff
@@ -463,4 +471,38 @@ public void canRepresentAllValues() {
...
+                true, false, true},
+            3, 2);
+
+        TensorValidator<Integer, IntegerTensor> validator = TensorValidator.thatChecksFor(0);
+        assertThat(validator.check(containsZero), equalTo(expectedMask));
+    }
+
+    @Test
+    public void youCanFixAValidationIssueByReplacingTheValue() {
+        IntegerTensor containsZero = IntegerTensor.create(1, 0, -1);
```
GeorgeNash: `containsMinusOne`

keanu-project/src/test/java/io/improbable/keanu/tensor/intgr/Nd4jIntegerTensorTest.java
```diff
@@ -463,4 +471,38 @@ public void canRepresentAllValues() {
...
+        IntegerTensor expectedResult = IntegerTensor.create(1, 0, 0);
+
+        TensorValidator validator = TensorValidator.thatChecksFor(-1).withPolicy(TensorValidationPolicy.changeValueTo(0));
+        assertThat(validator.validate(containsZero), equalTo(expectedResult));
+    }
+
+    @Test
+    public void youCanFixACustomValidationIssueByReplacingTheValue() {
+        IntegerTensor containsZero = IntegerTensor.create(1, 0, -1);
```
GeorgeNash: `containsMinusOne`

keanu-project/src/test/java/io/improbable/keanu/tensor/intgr/Nd4jIntegerTensorTest.java
```diff
@@ -463,4 +471,38 @@ public void canRepresentAllValues() {
...
+        IntegerTensor expectedResult = IntegerTensor.create(1, 0, 0);
+
+        Function<Integer, Boolean> checkFunction = x -> x >= 0;
```
GeorgeNash: Very cool!!

keanu-project/src/main/java/io/improbable/keanu/tensor/validate/check/CustomTensorValueChecker.java
```diff
@@ -0,0 +1,30 @@
...
+import org.nd4j.linalg.util.ArrayUtil;
+
+import io.improbable.keanu.tensor.Tensor;
+import io.improbable.keanu.tensor.bool.BooleanTensor;
+import io.improbable.keanu.tensor.validate.policy.TensorValidationPolicy;
+
+public class CustomTensorValueChecker<U, T extends Tensor<U>> implements TensorValueChecker<T> {
+
+    private final Function<U, Boolean> checkFunction;
+    private TensorValidationPolicy policy = TensorValidationPolicy.throwMessage("Invalid value found");
```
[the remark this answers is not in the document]
migwellian: Good spot - I moved it and forgot to delete it from here.

keanu-project/src/main/java/io/improbable/keanu/tensor/validate/check/CustomTensorValueChecker.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public CustomTensorValueChecker(Function<U, Boolean> checkFunction) {
+        this.checkFunction = checkFunction;
+    }
+
+    @Override
+    public BooleanTensor check(T tensor) {
```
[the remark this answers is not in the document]
migwellian: You mean this? It's not all that similar as it uses a `DataBuffer`.

    @Override
    public DoubleTensor apply(Function<Double, Double> function) {
        DataBuffer data = tensor.data().dup();
        for (int i = 0; i < data.length(); i++) {
            data.put(i, function.apply(data.getDouble(i)));
        }
        return new Nd4jDoubleTensor(data.asDouble(), this.getShape());
    }

keanu-project/src/main/java/io/improbable/keanu/tensor/validate/policy/ChangeValueTo.java
```diff
@@ -0,0 +1,30 @@
...
+
+import io.improbable.keanu.tensor.Tensor;
+import io.improbable.keanu.tensor.bool.BooleanTensor;
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+import io.improbable.keanu.tensor.intgr.IntegerTensor;
+
+public class ChangeValueTo<U, T extends Tensor<U>> implements TensorValidationPolicy<T> {
+    private final U value;
+
+    ChangeValueTo(U value) {
```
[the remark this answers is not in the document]
migwellian: package private - because it's created by the factory method TensorValidationPolicy.throwMessage. I've added a comment.

keanu-project/src/test/java/io/improbable/keanu/tensor/TensorValidationTest.java
```diff
@@ -0,0 +1,63 @@
...
+import io.improbable.keanu.tensor.validate.TensorValidator;
+import io.improbable.keanu.tensor.validate.policy.TensorValidationPolicy;
+
+public class TensorValidationTest {
+    @Test
+    public void youCanCheckTheValueOfATensor() {
+        DoubleTensor containsZero = DoubleTensor.create(1.0, 0.0, -1.0);
+        BooleanTensor expectedMask = BooleanTensor.create(new boolean[] {true, false, true});
+
+        TensorValidator validator = TensorValidator.thatChecksFor(0.).withPolicy(TensorValidationPolicy.changeValueTo(1e-8));
```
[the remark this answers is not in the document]
migwellian: yes indeed

keanu-project/src/main/java/io/improbable/keanu/tensor/validate/TensorValidator.java
migwellian: changed to `thatExpectsNotToFind`

keanu-project/src/test/java/io/improbable/keanu/tensor/intgr/Nd4jIntegerTensorTest.java
migwellian: roger

keanu-project/src/test/java/io/improbable/keanu/tensor/intgr/Nd4jIntegerTensorTest.java
migwellian: roger roger

migwellian:
I just pushed an update in which:
* I've added examples of fixing a distribution in 2 different ways:
  * Multinomial - disallow zeros
  * Logistic - use the new operator `logTimes(DoubleTensor)` which changes `NaN` to `0`
* Changed semantics of `TensorValidationPolicy#handle` and `TensorValidator#validate`: they now operate on the tensor inline (this made it easier to work with the inPlace operators)
* fixed a bug in `setWithMask` so that it works with NaN values (tricksy NaN!)

keanu-project/src/main/java/io/improbable/keanu/tensor/validate/policy/ChangeValueTo.java
```diff
@@ -0,0 +1,30 @@
...
+        this.value = value;
+    }
+
+    @Override
+    public void handle(T tensor, BooleanTensor result) {
+        if (tensor instanceof DoubleTensor) {
+            Double value = (Double) this.value;
+            DoubleTensor mask = result.not().toDoubleMask();
+            ((DoubleTensor) tensor).setWithMaskInPlace(mask, value);
```
gordoncaleb: Will this be a problem with the `inPlace` if possible contract?

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -472,6 +479,16 @@ public DoubleTensor logInPlace() {
         return this;
     }
 
+    @Override
+    public DoubleTensor logTimesInPlace(DoubleTensor y) {
+        TensorValidator<Double, Tensor<Double>> validator = TensorValidator.thatChecksForNaN();
+        validator.check(this);
+        validator.check(y);
+        this.logInPlace().timesInPlace(y);
+        validator.withPolicy(TensorValidationPolicy.changeValueTo(0.)).validate(this);
```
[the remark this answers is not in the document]
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/tensor/KeanuValueException.java
```diff
@@ -0,0 +1,16 @@
+package io.improbable.keanu.tensor;
+
+import io.improbable.keanu.tensor.bool.BooleanTensor;
+
+public class KeanuValueException extends RuntimeException {
```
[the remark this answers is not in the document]
migwellian: Maybe not the best name... it means there's an error relating to a value in the tensor, except it might also be in a scalar so I thought it wouldn't be right to call it TensorValueException.

Suggestions for a better name welcome!

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/ScalarDoubleTensor.java
```diff
@@ -717,6 +726,11 @@ public BooleanTensor greaterThanOrEqual(double value) {
         return BooleanTensor.scalar(this.value >= value);
     }
 
+    @Override
+    public BooleanTensor isNaN() {
+        return TensorValidator.thatChecksForNaN().check(this).not();
```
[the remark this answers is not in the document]
migwellian: done - 2 seperate ones, depending on the validation policy (throws vs. changes NaN to 0)

keanu-project/src/main/java/io/improbable/keanu/tensor/validate/policy/ChangeValueTo.java
migwellian: initially I had this method return a TENSOR but found I needed to do it in-place to make it work with the existing `inPlace` methods. I don't see any problems - am I missing something?

keanu-project/src/main/java/io/improbable/keanu/tensor/validate/TensorValidator.java
GeorgeNash: nice

keanu-project/src/main/java/io/improbable/keanu/tensor/validate/check/CustomTensorValueChecker.java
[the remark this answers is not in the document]
GeorgeNash: Ah fair enough

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -472,6 +479,23 @@ public DoubleTensor logInPlace() {
...
 
+    /**
+     * This is identical to log().times(y), except that it changes NaN results to 0.
+     * This is important when calculating 0log0, which should return 0
+     * See https://arcsecond.wordpress.com/2009/03/19/0log0-0-for-real/ for some mathematical justification
+     * @param y The tensor value to multiply by
+     * @return the log of this tensor multiplied by y
+     */
+    @Override
+    public DoubleTensor logTimesInPlace(DoubleTensor y) {
```
GeorgeNash: Awesome. I like the idea of putting a word in the method name that denotes it's doing validation as otherwise it may appear to be a shortcut. Perhaps: `safeLogTimes` or `validatedLogTimes` or `nanFreeSince93LogTimes`. (Not really on the last one)

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -472,6 +479,23 @@ public DoubleTensor logInPlace() {
...
+        TensorValidator<Double, Tensor<Double>> validator = TensorValidator.thatChecksForNaN();
+        validator.check(this);
+        validator.check(y);
+        this.logInPlace().timesInPlace(y);
```
GeorgeNash: I think you want to do `this = this.logInPlace().timesInPlace(y)` as the `timesInPlace` support broadcasting (times with different ranked tensors) which don't fulfil the `inPlace` contract

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -935,6 +961,11 @@ public BooleanTensor greaterThanOrEqual(double value) {
         return fromMask(tensor.gte(value), copyOf(getShape(), getShape().length));
     }
 
+    @Override
+    public BooleanTensor isNaN() {
+        return TensorValidator.thatChecksForNaN().check(this).not();
```
GeorgeNash: I find reading this a bit strange. A method called `isNaN` and then we're calling a method called `thatChecksForNaN` and we're inverting it. Can we change this to `TensorValidator. thatExpectsNotToFindNan().check(this).not()`

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -719,19 +743,21 @@ public DoubleTensor getLessThanOrEqualToMask(DoubleTensor lessThanOrEqualToThis)
     public DoubleTensor setWithMaskInPlace(DoubleTensor mask, Double value) {
 
         INDArray maskDup = unsafeGetNd4J(mask).dup();
-
+        double trueValue = 1.0;
```
icantsayrural: How is this change related to this PR?

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/ScalarDoubleTensor.java
```diff
@@ -491,6 +496,23 @@ public DoubleTensor logInPlace() {
...
+    /**
+     * This is identical to log().times(y), except that it changes NaN results to 0.
+     * This is important when calculating 0log0, which should return 0
+     * See https://arcsecond.wordpress.com/2009/03/19/0log0-0-for-real/ for some mathematical justification
+     * @param y The tensor value to multiply by
+     * @return the log of this tensor multiplied by y
+     */
+    @Override
+    public DoubleTensor logTimesInPlace(DoubleTensor y) {
+        TensorValidator<Double, Tensor<Double>> validator = TensorValidator.thatChecksForNaN();
```
icantsayrural: Shouldn't we be checking for 0?

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
migwellian: `safeLogTimes` sounds good to me

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
migwellian: it's a bug fix, and it is related to the rest of the PR: previously `setWithMaskInPlace` didn't work for `NaN`, i.e. it didn't set the tensor in those places where the tensor value was `NaN`.

I've captured this failure case in `Nd4jDoubleTensorTest#canSetToZeroWhenNaN`

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
migwellian: This is obsolete now as I've changed (improved) how this method works. It now does some direct tensor stuff.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/ScalarDoubleTensor.java
migwellian: No - we allow 0. The logic here is
1. before we do the operation, check there are no `NaN`s in `this` or `y` as if there were it would be wrong to assume it's ok to change them to `0`
2. do the log and times operations
3. change any resulting `NaN` to `0`

migwellian:
@gordoncaleb I've changed it so that `validate` returns a tensor now, which is nicer. The contract can be that `validate` should not mutate the tensor passed in.

FWIW - the code path in `setWithMaskInPlace` that does the operation not in-place will not be reached here, because the mask will always be the same shape as the tensor being set.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -472,6 +484,21 @@ public DoubleTensor logInPlace() {
...
+     * See https://arcsecond.wordpress.com/2009/03/19/0log0-0-for-real/ for some mathematical justification
+     * @param y The tensor value to multiply by
+     * @return the log of this tensor multiplied by y
+     */
+    @Override
+    public DoubleTensor safeLogTimesInPlace(DoubleTensor y) {
+        TensorValidator.NAN_CATCHER.validate(this);
+        TensorValidator.NAN_CATCHER.validate(y);
+        DoubleTensor result = this.logInPlace().timesInPlace(y);
+        return TensorValidator.NAN_FIXER.validate(result);
```
icantsayrural: If `this` is not zero but `y` is, then we should still return `NaN`

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -472,6 +484,21 @@ public DoubleTensor logInPlace() {
...
+    /**
+     * This is identical to log().times(y), except that it changes NaN results to 0.
+     * This is important when calculating 0log0, which should return 0
+     * See https://arcsecond.wordpress.com/2009/03/19/0log0-0-for-real/ for some mathematical justification
+     * @param y The tensor value to multiply by
+     * @return the log of this tensor multiplied by y
+     */
+    @Override
+    public DoubleTensor safeLogTimesInPlace(DoubleTensor y) {
+        TensorValidator.NAN_CATCHER.validate(this);
```
icantsayrural: Isn't it a bit aggressive to throw an exception if there is NaN in the tensor?

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
icantsayrural: Oh right, I mistakenly thought `log(0)` returns `NaN`!

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
icantsayrural: Without thinking too much about it, I'd expect values that at some point evaluate to `NaN` to stay `NaN`, so that the operations are not interrupted for the rest of the values in the tensor?

migwellian: 👍

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
migwellian: 👍

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
migwellian: I could implement it so if there's any `NaN`s in the tensors before the operation, they get put back into the result afterwards. It wouldn't be too hard to do, but I question if that's what we want. Happy to put the question up for debate.

keanu-project/src/main/java/io/improbable/keanu/tensor/validate/NaNFixingTensorValidator.java
```diff
@@ -0,0 +1,23 @@
...
+
+    private final double replacementValue;
+
+    public NaNFixingTensorValidator(double replacementValue) {
+        this.replacementValue = replacementValue;
+    }
+
+    @Override
+    public BooleanTensor check(DoubleTensor tensor) {
+        return tensor.isNaN().not();
```
gordoncaleb: The ND4J tensor does a `.not()` for each `isNan()` and then we do another `.not()` here and then the ChangeValueTo policy gives it another `.not()`. Do we ever need the isNaN to return the individual value? Any way to just replace the NaNs if present?

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/Logistic.java
```diff
@@ -30,7 +30,7 @@ private Logistic(DoubleTensor mu, DoubleTensor s) {
 
     @Override
     public DoubleTensor sample(int[] shape, KeanuRandom random) {
-        return random.nextDouble(shape).reciprocalInPlace().minusInPlace(1).logInPlace().timesInPlace(mu.minus(s));
```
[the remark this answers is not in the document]
migwellian: (After discussion with Caleb) - understood, `nextDouble` does not include the upper bound 1.0, so this can't happen. I'll revert it.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
migwellian: After discussion with the team - I've now implemented this as debug-only. In other words, by default the `NAN_CATCHER` and `NAN_FIXER` do nothing, unless you enable them programatically.

keanu-project/src/main/java/io/improbable/keanu/tensor/validate/NaNFixingTensorValidator.java
migwellian: Hmmm - 3 `not`s does seem excessive, doesn't it?

I've replaced `isNaN` with its opposite, `notNaN`. That eliminates the first 2 of the 3 `not`s.

I think we'd probably still want a method named `isNaN` in the API, since it matches the naming convention of `Double#isNaN`.

    default BooleanTensor isNaN() { return notNaN().not(); }

As for the 3rd `not` (in `ChangeValueTo`) - I decided as I worked through it that the semantics made more sense if when `check()` returns `true` it means "everything is fine", however `validate` has to then invert it to create a mask.

migwellian: Also - note that this class doesn't use a `ChangeValueTo` policy, it implements `validate` directly, by calling `tensor.replaceNaN` (which for tensors does the check and replace in one operation, without needing a mask to apply).

keanu-project/src/main/java/io/improbable/keanu/tensor/KeanuValueException.java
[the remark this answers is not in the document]
migwellian: I've renamed it `TensorValueException`.

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
```diff
@@ -39,6 +40,7 @@ private Multinomial(IntegerTensor n, DoubleTensor p) {
             p.sum(0).elementwiseEquals(DoubleTensor.ones(n.getShape())).allTrue(),
             "Probabilities must sum to one"
         );
+        TensorValidator.ZERO_CATCHER.validate(p);
```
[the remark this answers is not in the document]
migwellian: No

keanu-project/src/main/java/io/improbable/keanu/tensor/validate/check/TensorValueNotEqualsCheck.java
```diff
@@ -0,0 +1,28 @@
...
+
+    private final DATATYPE value;
+
+    public TensorValueNotEqualsCheck(DATATYPE value) {
+        this.value = value;
+    }
+
+    @Override
+    public BooleanTensor check(TENSOR tensor) {
+        if (tensor instanceof DoubleTensor) {
```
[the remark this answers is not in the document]
migwellian: But how would I create `that` in a generic way?

e.g.:

            DoubleTensor testTensor = DoubleTensor.create((Double) value, tensor.getShape());

keanu-project/src/main/java/io/improbable/keanu/tensor/validate/policy/ChangeValueTo.java
```diff
@@ -0,0 +1,31 @@
...
+    private final DATATYPE value;
+
+    // package private - because it's created by the factory method TensorValidationPolicy.changeValueTo
+    ChangeValueTo(DATATYPE value) {
+        this.value = value;
+    }
+
+    @Override
+    public TENSOR handle(TENSOR tensor, BooleanTensor result) {
+        if (tensor instanceof DoubleTensor) {
```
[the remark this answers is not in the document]
migwellian: There is a `setWithMask` on `NumberTensor`, but not on `Tensor`. I might be able to make that work, but it entails other changes: the `handle` method would need to take a `TENSOR mask` instead of `BooleanTensor result`. This changes the interface so has an impact on the other classes related to `TensorValidator`.

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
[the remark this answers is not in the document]
migwellian: I have implemented this now - so it's disabled for this specific context. I don't think it would be right to disable `ZERO_CATCHER` globally (even though it's not used anywhere else currently).

    public static final DebugTensorValidator CATEGORY_PROBABILITIES_CANNOT_BE_ZERO = new DebugTensorValidator<>(TensorValidator.ZERO_CATCHER);

PR #241, opened 2018-09-14
Feature/black box

GeorgeNash:
- The concept of a black box inside a vertex
- There are currently two implementations
-- A very sweeping / generic lambda implementation that lets the user execute anything. Java code, calls to a web server etc.
-- A more specific process vertex that's geared towards executing something on the command line

I thought we'd get these two in and then we could begin adding more specific http models etc

Things I'm not happy with right now:
- What value should be on these Model vertex's? Right now they have a hidden map that `ModelResultVertex`'s know to extract values from. The value on the model vertex itself should never be looked at, so for now I've just put it as zero. This feels horrendous.
- Having to call separate methods on the `ModelVertex` for different types of value


keanu-project/src/main/java/io/improbable/keanu/network/NetworkSnapshot.java
```diff
@@ -32,10 +32,10 @@ private NetworkSnapshot(Collection<Vertex> vertices) {
      */
     public void apply() {
         for (Vertex v : values.keySet()) {
-            v.unobserve();
             if (observed.contains(v)) {
                 v.observe(values.get(v));
             } else {
+                v.unobserve();
```
christophernorth: Nice change.  Just FMI - did you actually see a problem because of this, or is this just a code cleanup thing?

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/Black.java
```diff
@@ -0,0 +1,77 @@
+//package io.improbable.remorsefulradish.model.bellhop;
```
christophernorth: Mistakenly included?

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/Thing.java
```diff
@@ -0,0 +1,51 @@
+//package io.improbable.remorsefulradish.model.vertices;
```
christophernorth: And again?

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/nonprobabilistic/operators/BoolModelResultVertex.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public BoolModelResultVertex(ModelVertex model, VertexLabel label) {
+        this.model = model;
+        this.label = label;
+        setParents((Vertex) model);
+    }
+
+    @Override
+    public BooleanTensor getValue() {
+        if (!model.hasCalculated()) {
```
christophernorth: Do we have an issue where we will only calculate once?  Can we just make "eval" on the model vertex itself do the work?

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/nonprobabilistic/operators/BoolModelResultVertex.java
```diff
@@ -0,0 +1,42 @@
...
+import io.improbable.keanu.vertices.Vertex;
+import io.improbable.keanu.vertices.VertexLabel;
+import io.improbable.keanu.vertices.bool.BoolVertex;
+import io.improbable.keanu.vertices.dbl.KeanuRandom;
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.operators.multiple.ModelVertex;
+
+/**
+ * A non-probabilistic boolean vertex whose value is extracted from an upstream model vertex.
+ */
+public class BoolModelResultVertex extends BoolVertex implements NonProbabilistic<BooleanTensor> {
```
christophernorth: Having thought more on this, I wonder if the Result Vertices would be better as abstract base classes that each different model Vertex can provides a concrete implementation of (albeit a version that can read directly from a map would likely be useful for a lot of cases).  For instance, I could imagine a model that builds up a huge database - we wouldn't necessary want to store everything in memory all the time and we might just want to retrieve from disk as and when required?  Similarly, for something like a loop Vertex we'd be able to read the value directly from the "output" of the Loop whereas this implementation would mandate we store this output in a map to allow it to be extracted.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/ModelVertex.java
```diff
@@ -0,0 +1,45 @@
...
+import io.improbable.keanu.vertices.dbl.DoubleVertex;
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.DoubleModelResultVertex;
+import io.improbable.keanu.vertices.intgr.IntegerVertex;
+import io.improbable.keanu.vertices.intgr.nonprobabilistic.operators.IntegerModelResultVertex;
+
+import java.util.Map;
+
+public interface ModelVertex<T> extends NonProbabilistic<T> {
+
+    void run(Map<VertexLabel, Vertex<? extends Tensor>> inputs);
```
christophernorth: Do we need to re-specify the inputs here?  Shouldn't we just be able to get them from the input Vertices we've previously specified?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/ModelVertex.java
```diff
@@ -0,0 +1,45 @@
...
+
+    Map<VertexLabel, Tensor> updateValues(Map<VertexLabel, Vertex<? extends Tensor>> inputs);
+
+    boolean hasCalculated();
+
+    DoubleTensor getDoubleModelOutputValue(VertexLabel label);
+
+    IntegerTensor getIntegerModelOutputValue(VertexLabel label);
+
+    BooleanTensor getBooleanModelOutputValue(VertexLabel label);
```
christophernorth: I don't think we should have these at all really - we'd be better off IMO just throwing an exception if anyone tries to get the value of a ModelVertex explicitly (as opposed to via an output Vertex).  We should also likely throw an exception if anyone tries to explicitly set the model vertex as a parent.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/LambdaModelVertex.java
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * A vertex whose operation is the execution of a lambda.
+ *
+ * It is able to execute a lambda and is able to parse the result.
+ *
+ * It stores multiple output values and a model result vertex is required to extract a specific value.
+ */
+public class LambdaModelVertex extends DoubleVertex implements ModelVertex<DoubleTensor> {
+
+    private static final DoubleTensor MODEL_RETURN_VALUE = DoubleTensor.scalar(0.);
```
christophernorth: As mentioned below, we should probably just throw an exception if someone tries to get our value - did you see a problem with that approach?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/LambdaModelVertex.java
```diff
@@ -0,0 +1,93 @@
...
+
+    private Map<VertexLabel, Vertex<? extends Tensor>> inputs;
+    private Map<VertexLabel, Tensor> outputs;
+    private Consumer<Map<VertexLabel, Vertex<? extends Tensor>>> executor;
+    private Function<Map<VertexLabel, Vertex<? extends Tensor>>, Map<VertexLabel, Tensor>> extractOutput;
+    private boolean hasCalculated;
+
+    public LambdaModelVertex(Map<VertexLabel, Vertex<? extends Tensor>> inputs,
+                             Consumer<Map<VertexLabel, Vertex<? extends Tensor>>> executor,
+                             Function<Map<VertexLabel, Vertex<? extends Tensor>>, Map<VertexLabel, Tensor>> extractOutput) {
```
christophernorth: What's the thinking behind splitting up the executor and the extractOutput functions separate?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/ProcessModelVertex.java
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * It is able to execute an external process and is able to parse the result.
+ *
+ * It stores multiple output values and a model result vertex is required to extract a specific value.
+ */
+public class ProcessModelVertex extends DoubleVertex implements ModelVertex<DoubleTensor> {
+
+    private static final DoubleTensor MODEL_RETURN_VALUE = DoubleTensor.scalar(0.);
+
+    private String command;
```
migwellian: I wonder if you can remove the code duplication between `ProcessModelVertex` and `LambdaModelVertex`? For example both have the same fields `inputs`, `outputs`, `extractOutput`, `hasCalculated` and the same method implementations `sample`, `calculate`, `updateValues`

migwellian: in fact the two classes are so similar, do we even need both? A `ProcessModelVertex` is really just a `LambdaModelVertex` with a certain type of `executor`

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/LambdaModelVertex.java
```diff
@@ -0,0 +1,93 @@
...
+    @Override
+    public DoubleTensor calculate() {
+        run(inputs);
+        updateValues(inputs);
+        return MODEL_RETURN_VALUE;
+    }
+
+    @Override
+    public DoubleTensor sample(KeanuRandom random) {
+        for (Map.Entry<VertexLabel, Vertex<? extends Tensor>> input : inputs.entrySet()) {
```
gordoncaleb: You can iterate over inputs.values() instead of entrySet() and avoid confusion with getValue().

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/ProcessModelVertex.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Override
+    public void run(Map<VertexLabel, Vertex<? extends Tensor>> inputs) {
+        String newCommand = commandFormatter.apply(inputs, command);
+        try {
+            Process cmd = Runtime.getRuntime().exec(newCommand);
+            cmd.waitFor();
+            hasCalculated = true;
+        } catch (IOException | InterruptedException e) {
+            throw new IllegalArgumentException("Failed during execution of the process. " + e);
```
gordoncaleb: Is this an IllegalArgument? You can just wrap it in `RuntimeException("Message....", e)`

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/ProcessModelVertex.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Override
+    public BooleanTensor getBooleanModelOutputValue(VertexLabel label) {
+        return (BooleanTensor) outputs.get(label);
+    }
+
+    @Override
+    public DoubleTensor sample(KeanuRandom random) {
+        for (Map.Entry<VertexLabel, Vertex<? extends Tensor>> input : inputs.entrySet()) {
```
gordoncaleb: Iterate over `inputs.values()`

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/LambdaModelVertex.java
```diff
@@ -0,0 +1,93 @@
...
+            input.getValue().sample();
+        }
+        return calculate();
+    }
+
+    @Override
+    public void run(Map<VertexLabel, Vertex<? extends Tensor>> inputs) {
+        executor.accept(inputs);
+        hasCalculated = true;
```
gordoncaleb: Is this just hasCalculated at least once? As in, my parent does have a non-null value for me?

keanu-project/src/main/java/io/improbable/keanu/network/NetworkSnapshot.java
GeorgeNash: Was causing problems with the `ModelVertex` and it ended up being this unneccessary observe!

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/Black.java
GeorgeNash: deleted

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/Thing.java
GeorgeNash: oops - deleted

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/ModelVertex.java
GeorgeNash: Nice - deleted the argument!

keanu-project/src/main/java/io/improbable/keanu/network/NetworkSnapshot.java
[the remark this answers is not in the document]
GeorgeNash: yep

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/ModelVertex.java
GeorgeNash: Yeah - the algorithms will ask it to recalculate but the value will not be looked at. Throwing an exception will cause it to halt

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/LambdaModelVertex.java
GeorgeNash: Yeah - the algorithms will ask it to recalculate but the value will not be looked at. Throwing an exception will cause it to halt

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/LambdaModelVertex.java
GeorgeNash: To make what's going on as clear to the user as possible. By splitting it up it makes it clear that they need to run some command and then extract the outputs somehow. Whereas if we required one argument that had to execute and get outputs it would be a bit of a daunting task.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/ModelVertex.java
```diff
@@ -0,0 +1,45 @@
+package io.improbable.keanu.vertices.dbl.nonprobabilistic.operators.multiple;
```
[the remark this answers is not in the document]
GeorgeNash: Good idea - moved them

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/nonprobabilistic/operators/BoolModelResultVertex.java
GeorgeNash: Interesting idea - it sounds like a separate ticket and outside the scope of this ticket though. I think it would be best to get a basic implementation completed first that used maps and then add a ticket to extend this to use a user defined concept

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/ModelVertex.java
GeorgeNash: As mentioned above - these are needed for the current `ModelResultVertex` implementation

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/ProcessModelVertex.java
GeorgeNash: Great point - have factored out the `ProcessModelVertex` to use the `LambdaModelVertex`

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/LambdaModelVertex.java
```diff
@@ -0,0 +1,93 @@
...
+
+    private static final DoubleTensor MODEL_RETURN_VALUE = DoubleTensor.scalar(0.);
+
+    private Map<VertexLabel, Vertex<? extends Tensor>> inputs;
+    private Map<VertexLabel, Tensor> outputs;
+    private Consumer<Map<VertexLabel, Vertex<? extends Tensor>>> executor;
+    private Function<Map<VertexLabel, Vertex<? extends Tensor>>, Map<VertexLabel, Tensor>> extractOutput;
+    private boolean hasCalculated;
+
+    public LambdaModelVertex(Map<VertexLabel, Vertex<? extends Tensor>> inputs,
```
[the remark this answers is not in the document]
GeorgeNash: Yep only supports tensor vertices at the moment. Do you think it's worth extending to support more?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/LambdaModelVertex.java
```diff
@@ -0,0 +1,93 @@
...
+        this.extractOutput = extractOutput;
+        this.hasCalculated = false;
+        setParents(inputs.entrySet().stream().map(r -> r.getValue()).collect(Collectors.toList()));
+    }
+
+    @Override
+    public DoubleTensor calculate() {
+        run(inputs);
+        updateValues(inputs);
+        return MODEL_RETURN_VALUE;
```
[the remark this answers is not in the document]
GeorgeNash: Changed it to be a void method - we still need to `setValue()` on this model vertex though....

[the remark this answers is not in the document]
GeorgeNash: Ok - changed `hasValue()` to be an overridden method

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/ProcessModelVertex.java
GeorgeNash: Ok - changed.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/LambdaModelVertex.java
GeorgeNash: Yep exactly, I can change the name to reflect that

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/nonprobabilistic/operators/BoolModelResultVertex.java
GeorgeNash: Added a test for this - `canRunEvalOnTheOutputsToRecalculateTheModel`

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
```diff
@@ -77,6 +78,13 @@ public NetworkSamples getPosteriorSamples(BayesianNetwork bayesianNetwork,
             .generate(sampleCount);
     }
 
+    public NetworkSamples getPosteriorSamples(BayesianNetwork bayesianNetwork,
+                                              Vertex vertexToSampleFrom,
+                                              int sampleCount) {
+        return generatePosteriorSamples(bayesianNetwork, Arrays.asList(vertexToSampleFrom))
```
christophernorth: is it worth considering Collections.singletonList() instead of asList()?

gordoncaleb: 👍

keanu-project/src/main/java/io/improbable/keanu/vertices/NonProbabilistic.java
```diff
@@ -3,8 +3,8 @@
 public interface NonProbabilistic<T> extends Observable<T> {
 
     default boolean contradictsObservation() {
-        return isObserved() && getObservedValue().map(v -> !v.equals(calculate())).orElse(false);
+        return isObserved() && getObservedValue().map(v -> !v.equals(getObservedValue())).orElse(false);
```
gordoncaleb: This will always succeed. Isn't it comparing the observed value to the observed value? It might be that calculate() can't be void....

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
GeorgeNash: cool changed to `singletonList`

keanu-project/src/main/java/io/improbable/keanu/vertices/NonProbabilistic.java
GeorgeNash: Aye....reverted the `calculate` change

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/LambdaModelVertexTest.java
```diff
@@ -0,0 +1,274 @@
...
+
+    @Before
+    public void setup() throws IOException {
+        random = new KeanuRandom(1);
+        rainReader = mock(BufferedReader.class);
+        humidityReader = mock(BufferedReader.class);
+        suggestedFactorSuncream = mock(BufferedReader.class);
+        isSunnyReader = mock(BufferedReader.class);
+
+        when(rainReader.readLine()).thenAnswer(new Answer<Object>() {
```
gordoncaleb: If you don't need the arguments to a mocked method then I think you can do just `.thenReturn()`

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
```diff
@@ -2,11 +2,7 @@
 
 import static io.improbable.keanu.algorithms.mcmc.proposal.MHStepVariableSelector.SINGLE_VARIABLE_SELECTOR;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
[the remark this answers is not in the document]
GeorgeNash: oops - reverted

keanu-project/src/main/java/io/improbable/keanu/vertices/model/LambdaModelVertex.java
```diff
@@ -0,0 +1,85 @@
...
+
+    public LambdaModelVertex(Map<VertexLabel, Vertex<? extends Tensor>> inputs,
+                             Consumer<Map<VertexLabel, Vertex<? extends Tensor>>> executor,
+                             Function<Map<VertexLabel, Vertex<? extends Tensor>>, Map<VertexLabel, Tensor>> extractOutput) {
+        this.inputs = inputs;
+        this.outputs = Collections.EMPTY_MAP;
+        this.executor = executor;
+        this.extractOutput = extractOutput;
+        this.hasValue = false;
+        setParents(inputs.entrySet().stream().map(r -> r.getValue()).collect(Collectors.toList()));
```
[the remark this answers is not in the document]
GeorgeNash: nice!

keanu-project/src/main/java/io/improbable/keanu/vertices/model/ProcessModelVertex.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.function.Function;
+
+/**
+ * A vertex whose operation is the execution of a command line process.
+ *
+ * It is able to execute this process and parse the result.
+ *
+ * It stores multiple output values and a model result vertex is required to extract a specific value.
+ */
+public class ProcessModelVertex {
```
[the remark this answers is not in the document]
GeorgeNash: Moved to `LambdaModelVertex.createFromProcess`

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/LambdaModelVertexTest.java
```diff
@@ -0,0 +1,274 @@
...
+            @Override
+            public Object answer(InvocationOnMock invocation) {
+                double chanceOfRainScalingFactorFromModel = 0.1;
+                return String.valueOf(inputToModel.getValue().scalar() * chanceOfRainScalingFactorFromModel);
+            }
+        });
+
+        when(humidityReader.readLine()).thenAnswer(new Answer<Object>() {
```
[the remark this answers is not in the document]
GeorgeNash: cool - changed

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/LambdaModelVertexTest.java
```diff
@@ -0,0 +1,274 @@
...
+        );
+
+        double averagePosteriorInput = posteriorSamples.getDoubleTensorSamples(inputToModel).getAverages().scalar();
+
+        Assert.assertEquals(29., averagePosteriorInput, 0.1);
+    }
+
+    private void modelExecution(Map<VertexLabel, Vertex<? extends Tensor>> inputs) {
+        //This is a placeholder for the function that would write to file.
+        //Instead we are mocking the file IO
```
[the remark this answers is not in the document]
GeorgeNash: Good idea - I've added some logic to mimic a model that writes values to file.

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/ProcessModelVertexTest.java
```diff
@@ -0,0 +1,290 @@
...
+    private BufferedReader suggestedFactorSuncream;
+
+    @Mock
+    private BufferedReader isSunnyReader;
+
+    private KeanuRandom random;
+    private DoubleVertex inputToModel;
+
+    @Before
+    public void setup() throws IOException {
```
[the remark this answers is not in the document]
GeorgeNash: good idea - renamed to: `mockFilesToReadModelOutput`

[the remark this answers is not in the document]
GeorgeNash: Yep added this to the top of the test class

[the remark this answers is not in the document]
GeorgeNash: Thanks for the help with this!

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/LambdaModelVertexTest.java
```diff
@@ -0,0 +1,231 @@
...
+
+    It takes one input, Temperature, and produces two outputs of type double, Chance of Rain & Humidity. These outputs
+    are written to file.
+
+    It also produces one integer output and one boolean output. Suggested Factor of Suncream and 'is it sunny'. These
+    are also written to file.
+     */
+
+    private KeanuRandom random;
+    private DoubleVertex inputToModel = new ConstantDoubleVertex(50.);
```
migwellian: You never actually use the value of `50.`. If instead you initialise it to `25.` in a `@Before` method then you won't need to reinitialise `inputToModel` or `weatherModel` in your first 4 tests.

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/LambdaModelVertexTest.java
```diff
@@ -0,0 +1,231 @@
...
+        inputToModel.setAndCascade(inputValue);
+        Assert.assertEquals(shouldIBringUmbrella.getValue().scalar(), 20.0, 1e-6);
+    }
+
+    @Test
+    public void canRunEvalOnTheOutputsToRecalculateTheModel() {
+        inputToModel = new ConstantDoubleVertex(25);
+        weatherModel.setInputToModel(inputToModel);
+        Map<VertexLabel, Vertex<? extends Tensor>> inputs = new HashMap<>();
+        inputs.put(new VertexLabel("Temperature"), inputToModel);
```
migwellian: one-liner:

        Map<VertexLabel, Vertex<? extends Tensor>> inputs = ImmutableMap.of(new VertexLabel("Temperature"), inputToModel);

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/ProcessModelVertexTest.java
```diff
@@ -0,0 +1,239 @@
...
+    @Before
+    public void mockFilesToReadModelOutputFrom() throws IOException {
+        random = new KeanuRandom(1);
+        weatherModel = new SimpleWeatherModel(inputToModel);
+    }
+
+    @Test
+    public void canRunAModelInAModel() {
+        inputToModel = new ConstantDoubleVertex(25);
+        weatherModel.setInputToModel(inputToModel);
```
migwellian: ditto my earlier comment - set `inputToModel` in a `@Before` method and you won't need these 2 lines

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/SimpleWeatherModel.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public double blackBoxRainModel(double temperature) {
+        return temperature * 0.1;
+    }
+
+    public double blackBoxHumidityModel(double temperature) {
+        return temperature * 2;
+    }
+
+    public int blackBoxSunCreamModel(double temperature) {
```
migwellian: can be private

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/SimpleWeatherModel.java
```diff
@@ -0,0 +1,78 @@
...
+        return (int) (temperature / 10.0);
+    }
+
+    public boolean blackBoxIsSunnyModel(double temperature) {
```
migwellian: can be private

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/LambdaModelVertexTest.java
```diff
@@ -0,0 +1,231 @@
...
+        double averagePosteriorInput = posteriorSamples.getDoubleTensorSamples(inputToModel).getAverages().scalar();
+
+        Assert.assertEquals(29., averagePosteriorInput, 0.1);
+    }
+
+    private void modelExecution(Map<VertexLabel, Vertex<? extends Tensor>> inputs) {
+        double temperature = inputs.get(new VertexLabel("Temperature")).getValue().asFlatDoubleArray()[0];
+        try {
+            double chanceOfRain = weatherModel.blackBoxRainModel(temperature);
+            double humidity = weatherModel.blackBoxHumidityModel(temperature);
```
migwellian: Wouldn't it make more sense for the model execution to be a method on `SimpleWeatherModel`? It feels wrong that `blackBoxRainModel` and `blackBoxHumidityModel` are public. I know this is only test code but it serves an important role of demonstrating how this would work in a real scenario.

keanu-project/src/main/java/io/improbable/keanu/vertices/model/LambdaModelVertex.java
```diff
@@ -0,0 +1,114 @@
...
+    public boolean hasValue() {
+        return hasValue;
+    }
+
+    @Override
+    public DoubleTensor sample(KeanuRandom random) {
+        for (Vertex<? extends Tensor> input : inputs.values()) {
+            input.sample();
+        }
+        calculate();
```
migwellian: `return calculate()`?

keanu-project/src/main/java/io/improbable/keanu/vertices/model/LambdaModelVertex.java
```diff
@@ -0,0 +1,114 @@
...
+                throw new RuntimeException("Failed during execution of the process. " + e);
+            }
+        }, updateValues);
+    }
+
+    @Override
+    public DoubleTensor calculate() {
+        run();
+        updateValues(inputs);
+        return DoubleTensor.scalar(0.0);
```
migwellian: why are you returning zero?

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/LambdaModelVertexTest.java
GeorgeNash: Nice! Updated this

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/LambdaModelVertexTest.java
GeorgeNash: nice - updated

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/SimpleWeatherModel.java
GeorgeNash: changed

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/SimpleWeatherModel.java
GeorgeNash: changed

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/ProcessModelVertexTest.java
GeorgeNash: Nice changed

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/LambdaModelVertexTest.java
GeorgeNash: Yeah very good point - looks much better with that logic living on the weather model. I've also moved across the logic of how to read the output once the model has executed.

keanu-project/src/main/java/io/improbable/keanu/vertices/model/LambdaModelVertex.java
GeorgeNash: Nice - changed

keanu-project/src/main/java/io/improbable/keanu/vertices/model/LambdaModelVertex.java
migwellian: Understood. I agree that we shouldn't enable vertices to have multiple values. But would it be OK to return the value of the first output vertex (ordered by VertexId)? Or - could the `LambdaModelVertex` receive the `VertexLabel` in its constructor so it knows how to return its output?

 If the answer is no - we need to document the semantics in javadoc here.

PR #243, opened 2018-09-14
fix findbug malicious code vulnerability: "Field isn't final but should be"

migwellian:
This static field public but not final, and could be changed by malicious code or by accident from another package. The field could be made final to avoid this vulnerability.


keanu-project/src/main/java/io/improbable/keanu/util/ProgressBar.java
```diff
@@ -12,7 +12,7 @@
 
 public class ProgressBar {
 
-    public static PrintStream defaultPrintStream = System.out;
+    private static final PrintStream DEFAULT_PRINT_STREAM = System.out;
```
gordoncaleb: This one is actually intentional. It's meant to be a global switch to a specific printstream for the progress bar.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/PartialDerivatives.java
```diff
@@ -17,7 +17,7 @@
 
 public class PartialDerivatives {
 
-    public static PartialDerivatives OF_CONSTANT = new PartialDerivatives(Collections.emptyMap());
+    public static final PartialDerivatives OF_CONSTANT = new PartialDerivatives(Collections.emptyMap());
```
gordoncaleb: This one should indeed be final.

keanu-project/src/main/java/io/improbable/keanu/util/ProgressBar.java
migwellian: I made it private, non-final and added a getter

PR #244, opened 2018-09-14
Doomed test for equality to NaN

migwellian:
This code checks to see if a floating point value is equal to the special Not A Number value (e.g., if (x == Double.NaN)). However, because of the special semantics of NaN, no value is equal to Nan, including NaN. Thus, x == Double.NaN always evaluates to false. To check to see if a value contained in x is the special Not A Number value, use Double.isNaN(x) (or Float.isNaN(x) if x is floating point precision).

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/nongradient/FitnessFunction.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/nongradient/FitnessFunction.java
@@ -44,7 +44,7 @@ public class FitnessFunction {
     }
 
     public static boolean isValidInitialFitness(double fitnessValue) {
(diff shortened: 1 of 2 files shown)


keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/nongradient/FitnessFunction.java
```diff
@@ -44,7 +44,7 @@ public MultivariateFunction fitness() {
     }
 
     public static boolean isValidInitialFitness(double fitnessValue) {
-        return fitnessValue == Double.NEGATIVE_INFINITY || fitnessValue == Double.NaN;
+        return fitnessValue == Double.NEGATIVE_INFINITY || Double.isNaN(fitnessValue);
```
gordoncaleb: Nice one 👍

PR #245, opened 2018-09-14
class overrides equals(Object), but does not override hashCode()

migwellian:
This class overrides equals(Object), but does not override hashCode(), and inherits the implementation of hashCode() from java.lang.Object (which returns the identity hash code, an arbitrary value assigned to the object by the VM).  Therefore, the class is very likely to violate the invariant that equal objects must have equal hashcodes.


keanu-project/src/test/java/io/improbable/keanu/util/csv/pojo/byrow/ObjectParserWithPublicFieldTest.java
```diff
@@ -93,9 +95,17 @@ public TestPOJOWithPublicFieldsClass(int id, String name) {
 
         @Override
         public boolean equals(Object o) {
```
gordoncaleb: Nice catch. These would ideally both be generated.

PR #246, opened 2018-09-17
LOOOOPS

migwellian:
A convenience wrapper around Plates to make it straightforward for the user to create a loop.
You must specify a max size, and the loop will (by default) throw if you reach this count without terminating.


keanu-project/src/main/java/io/improbable/keanu/plating/Plate.java
```diff
@@ -22,6 +24,17 @@ public Plate() {
...
     }
 
+
+    public <T extends Vertex<?>> void addAll(T... vertices) {
+        addAll(ImmutableList.copyOf(vertices));
+    }
+
+    public <T extends Vertex<?>> void addAll(Collection<T> vertices) {
+        for (T v : vertices) {
+            add(v);
```
christophernorth: is there a reason you've used a for loop instead of forEach()?

keanu-project/src/test/java/io/improbable/keanu/vertices/VertexMatchers.java
```diff
@@ -0,0 +1,92 @@
...
+                return valueMatcher.matches(vertex.getValue());
+            }
+
+            @Override
+            public void describeTo(Description description) {
+                description.appendText("vertex with value ").appendDescriptionOf(valueMatcher);
+            }
+        };
+    }
+}
```
christophernorth: Awesome

keanu-project/src/test/java/io/improbable/keanu/plating/LoopTest.java
```diff
@@ -0,0 +1,167 @@
...
+            }
+            BoolVertex condition = loop.getPlates().asList().get(firstFailure).get(Loop.CONDITION_LABEL);
+            condition.setAndCascade(false);
+            Double expectedOutput = new Double(firstFailure);
+            assertThat(output, VertexMatchers.hasValue(expectedOutput));
+        }
+    }
+
+    @Test
+    public void theConditionCanBeAFunctionOfThePlateVariables() {
```
GeorgeNash: cool!

keanu-project/src/main/java/io/improbable/keanu/plating/loop/Loop.java
```diff
@@ -0,0 +1,137 @@
...
+import io.improbable.keanu.plating.Plates;
+import io.improbable.keanu.tensor.Tensor;
+import io.improbable.keanu.vertices.Vertex;
+import io.improbable.keanu.vertices.VertexLabel;
+import io.improbable.keanu.vertices.bool.BoolVertex;
+
+/**
+ * A Loop object is a convenient wrapper around some Plates.
+ * See LoopTest.java for examples of how it's used.
+ * The way it works is to unroll the loop up to a maximum size and evaluate it in full
```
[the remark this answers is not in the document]
GeorgeNash: The `up to` makes it sound like it could stop early

keanu-project/src/main/java/io/improbable/keanu/plating/loop/Loop.java
```diff
@@ -0,0 +1,137 @@
...
+        if (initialState.getLabel() == null) {
+            initialState.setLabel(VALUE_OUT_LABEL);
+        }
+        return startingFrom(ImmutableList.of(initialState));
+    }
+
+    /**
+     * A factory method for creating a loop
+     *
+     * @param first  - the first Vertex (mandatory)
```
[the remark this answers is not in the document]
GeorgeNash: If it's to make sure at least one gets passed in I'd rather have a `V... vertices` argument and just assert the length `>=` 1 inside the method

keanu-project/src/main/java/io/improbable/keanu/plating/loop/Loop.java
```diff
@@ -0,0 +1,137 @@
...
+    public static <V extends Vertex<?>> LoopBuilder startingFrom(Collection<V> initialState) {
+        return new LoopBuilder(initialState);
+    }
+
+    /**
+     * @param <V> the output type
+     * @return the output of the Loop (i.e. the output Vertex from the final Plate)
+     * @throws LoopException if the loop was too short and hit its maximum unrolled size
+     */
+    public <V extends Vertex<? extends Tensor<?>>> V getOutput() throws LoopException {
```
gordoncaleb: Does this have to output a Tensor?

keanu-project/src/main/java/io/improbable/keanu/plating/Plate.java
migwellian: nope - changed

keanu-project/src/main/java/io/improbable/keanu/plating/Plate.java
```diff
@@ -22,6 +24,17 @@ public Plate() {
         this.contents = new HashMap<>();
     }
 
+
```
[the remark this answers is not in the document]
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/plating/Plate.java
```diff
@@ -22,6 +24,17 @@ public Plate() {
...
+    public <T extends Vertex<?>> void addAll(T... vertices) {
```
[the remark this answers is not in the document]
migwellian: it's not good practice to use raw types - see https://stackoverflow.com/a/1351384/741789

Effective Java, Item 23: Don't use raw types in new code

keanu-project/src/main/java/io/improbable/keanu/plating/loop/Loop.java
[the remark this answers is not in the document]
migwellian: done - "unroll the loop to a given maximum size"

keanu-project/src/main/java/io/improbable/keanu/plating/loop/Loop.java
```diff
@@ -0,0 +1,137 @@
...
+    public static final VertexLabel LOOP_LABEL = new VertexLabel("loop");
+    static final VertexLabel STILL_LOOPING_LABEL = PlateBuilder.proxyFor(LOOP_LABEL);
+    public static final int DEFAULT_MAX_COUNT = 100;
+    private final Plates plates;
+    private final boolean throwWhenMaxCountIsReached;
+
+    /**
+     * package-private because it is intended to be created by the LoopBuilder
+     *
+     * @param plates
```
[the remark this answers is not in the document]
migwellian: added a description

keanu-project/src/main/java/io/improbable/keanu/plating/loop/Loop.java
```diff
@@ -0,0 +1,137 @@
...
+
+    public Plates getPlates() {
+        return plates;
+    }
+
+    /**
+     * A factory method for creating a loop
+     * It automatically labels the initial state correctly for you.
+     *
+     * @param initialState - a single Vertex that defines the loop's base case.
```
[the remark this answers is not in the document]
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopBuilder.java
```diff
@@ -0,0 +1,180 @@
...
+        LoopBuilder2(Collection<Vertex> initialState, Function<Plate, BoolVertex> conditionFunction, Map<VertexLabel, VertexLabel> customMappings, int maxLoopCount, boolean throwWhenMaxCountIsReached){
+            this.initialState = setInitialState(initialState);
+            this.conditionFunction = conditionFunction;
+            this.customMappings = customMappings;
+            this.maxLoopCount = maxLoopCount;
+            this.throwWhenMaxCountIsReached = throwWhenMaxCountIsReached;
+        }
+
+        private ImmutableList<Vertex> setInitialState(Collection<Vertex> initialState){
+            Vertex valueOutWhenAlwaysTrue = null;
```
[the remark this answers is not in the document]
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopBuilder.java
```diff
@@ -0,0 +1,180 @@
...
+
+            try {
+                List<Vertex> outputVertices = initialState.stream().filter(v -> Loop.VALUE_OUT_LABEL.equals(v.getLabel())).collect(Collectors.toList());
```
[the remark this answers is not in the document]
migwellian: done - nice!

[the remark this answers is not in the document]
migwellian: `findFirst` is neat, however `getOnlyElement` is a stricter check as it will throw if there is more than one. But anyway, I can address your point by inlining `outputVertices` so it's a one-liner:

    Vertex<?> outputVertex = Iterables.getOnlyElement(initialState.stream().filter(v -> Loop.VALUE_OUT_LABEL.equals(v.getLabel())).collect(Collectors.toList()));

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopBuilder.java
```diff
@@ -0,0 +1,180 @@
...
+        /**
+         * A mandatory method to specify the iteration step
+         *
+         * @param iterationFunction - a lambda that takes the Proxy input vertex
+         *                          and creates and returns a new output Vertex
+         * @return the fully constructed Loop object
+         */
+        public Loop apply(Function<DoubleVertex, DoubleVertex> iterationFunction){
+            return apply((plate, valueIn) -> {
+                return iterationFunction.apply(valueIn);
```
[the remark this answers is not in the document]
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopBuilder.java
```diff
@@ -0,0 +1,180 @@
...
+        /**
+         * A mandatory method to specify the iteration step
+         *
+         * @param iterationFunction - a lambda that takes the current Plate and the Proxy input vertex
+         *                          and creates and returns a new output vertex
+         * @return the fully constructed Loop object
+         */
+        public Loop apply(BiFunction<Plate, DoubleVertex, DoubleVertex> iterationFunction){
+            Plates plates = new PlateBuilder<Integer>()
+                .withInitialState(initialState.toArray(new Vertex[0]))
```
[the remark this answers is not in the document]
migwellian: it's a gnarly-looking java trick when converting a Collection to a raw array. It bears no relation to the actual size of the collection! Without this, it thinks the raw array is of type `Object[]` :-(

It turns out in Java 8 there's an alternative way using streams:

    initialState.stream().toArray(Vertex[]::new)

and in Java 11 there will be the same without streams:

    initialState.toArray(Vertex[]::new)

I'll go with the stream version as I agree the zero size thing is confusing.

keanu-project/src/main/java/io/improbable/keanu/plating/Plate.java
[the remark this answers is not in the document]
migwellian: Here's a good example in the context of Keanu.

Syntax error:

        Vertex<?> output = getSomeVertex();
        output.setValue(0.);

No error (dangerous - is this a `Vertex<Double>` or a `Vertex<DoubleTensor>`?):

        Vertex output = getSomeVertex();
        output.setValue(0.);

No error (safer - at least you're declaring your assumption):

        DoubleVertex output = getSomeVertex();
        output.setValue(0.);

keanu-project/src/main/java/io/improbable/keanu/plating/loop/Loop.java
migwellian: Same reasons as for the VertexDictionary in my last PR. Therefore same conclusion: I'll revert it to `Vertex<?>` for now, until such time as we can reproduce the issue (if there is any) with  a ClassCastException.

keanu-project/src/main/java/io/improbable/keanu/plating/Plate.java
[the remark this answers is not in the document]
GeorgeNash: Fair point - thanks for the code snippet!

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopBuilder.java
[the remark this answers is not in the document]
GeorgeNash: Nice

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopBuilder.java
```diff
@@ -0,0 +1,180 @@
...
+        this.initialState = ImmutableList.copyOf(initialState);
+    }
+
+    /**
+     * An optional method to override the default value
+     *
+     * @param maxCount - the max number of times the loop can run
+     * @return self
+     */
+    public LoopBuilder atMost(int maxCount) {
```
gordoncaleb: The name here should be more descriptive I think. `withMaxIterations` or `atMostIterations`...

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopBuilder.java
```diff
@@ -0,0 +1,180 @@
...
+        return this;
+    }
+
+    /**
+     * An optional method to override the default behaviour
+     * If the maximum loop count is exceeded, it will log a warning instead of throwing
+     *
+     * @return self
+     */
+    public LoopBuilder dontThrowWhenMaxCountIsReached() {
```
gordoncaleb: nit: `doNotThrow...`

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopBuilder.java
```diff
@@ -0,0 +1,180 @@
...
+     * A mandatory method to specify the condition
+     *
+     * @param conditionFunction - a lambda that takes the current Plate and creates and returns a new BoolVertex
+     * @return the next stage builder
+     */
+    public LoopBuilder2 whilst(Function<Plate, BoolVertex> conditionFunction){
+        return new LoopBuilder2(initialState, conditionFunction, customMappings.build(), maxLoopCount, throwWhenMaxCountIsReached);
+    }
+
+    public class LoopBuilder2 {
```
gordoncaleb: What's the 2 mean?

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopBuilder.java
```diff
@@ -0,0 +1,180 @@
...
+        private final Collection<Vertex> initialState;
+        private final Function<Plate, BoolVertex> conditionFunction;
+        private final Map<VertexLabel, VertexLabel> customMappings;
+        private final int maxLoopCount;
+        private final boolean throwWhenMaxCountIsReached;
+        private final VertexLabel VALUE_OUT_WHEN_ALWAYS_TRUE_LABEL = new VertexLabel("loop_value_out_when_always_true");
```
gordoncaleb: We're really relying on the labels here to describe the structure of the graph. Should these special labels that describe structure be some other type of meta data on the vertex?

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopBuilder.java
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopBuilder.java
```diff
@@ -0,0 +1,180 @@
...
+        return this;
+    }
+
+    /**
+     * A mandatory method to specify the condition
+     *
+     * @param conditionSupplier - a lambda that creates and returns a new BoolVertex
+     * @return the next stage builder
+     */
+    public LoopBuilder2 whilst(Supplier<BoolVertex> conditionSupplier){
```
gordoncaleb: `iterateWhile`?

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopBuilder.java
migwellian: done (withMaxIterations)

keanu-project/src/main/java/io/improbable/keanu/plating/loop/Loop.java
```diff
@@ -0,0 +1,137 @@
...
+     * @param <V>          the input type
+     * @return a builder object
+     */
+    public static <V extends Vertex<?>> LoopBuilder startingFrom(V initialState) {
```
gordoncaleb: `withInitialConditions`?

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopBuilder.java
migwellian: Lol, George asked me the same thing.

It's a pretty common pattern in Java libraries, I think. You want to guide the user through the building process, and enforce that the method calls along the way are mandatory (`startingFrom`, `whilst`, `apply`).

I know it's not a descriptive name but it's only an internal class that no one will ever reference directly in the calling code.

keanu-project/src/main/java/io/improbable/keanu/plating/loop/Loop.java
```diff
@@ -0,0 +1,137 @@
...
+     */
+    public <V extends Vertex<? extends Tensor<?>>> V getOutput() throws LoopException {
+        Plate finalPlate = plates.asList().get(plates.size() - 1);
+        checkIfMaxCountHasBeenReached(finalPlate);
+        return finalPlate.get(VALUE_OUT_LABEL);
+    }
+
+    private void checkIfMaxCountHasBeenReached(Plate plate) throws LoopException {
+        BoolVertex stillLooping = plate.get(STILL_LOOPING_LABEL);
+        if (Arrays.stream(stillLooping.getValue().asFlatArray()).anyMatch(v -> v == true)) {
```
gordoncaleb: Can you use `BooleanTensor::allFalse`? `if (!stillLooping.getValue().allFalse()) {`

keanu-project/src/main/java/io/improbable/keanu/plating/loop/Loop.java
```diff
@@ -0,0 +1,137 @@
...
+        return new LoopBuilder(initialState);
+    }
+
+    /**
+     * @param <V> the output type
+     * @return the output of the Loop (i.e. the output Vertex from the final Plate)
+     * @throws LoopException if the loop was too short and hit its maximum unrolled size
+     */
+    public <V extends Vertex<? extends Tensor<?>>> V getOutput() throws LoopException {
+        Plate finalPlate = plates.asList().get(plates.size() - 1);
```
gordoncaleb: Can we add a `Plates::getLast`?

keanu-project/src/main/java/io/improbable/keanu/plating/loop/Loop.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    private void checkIfMaxCountHasBeenReached(Plate plate) throws LoopException {
+        BoolVertex stillLooping = plate.get(STILL_LOOPING_LABEL);
+        if (Arrays.stream(stillLooping.getValue().asFlatArray()).anyMatch(v -> v == true)) {
+            String errorMessage = "Loop has exceeded its max count " + plates.size();
+            if (throwWhenMaxCountIsReached) {
+                throw new LoopException(errorMessage);
+            } else {
+                log.warn(errorMessage);
```
gordoncaleb: Why WARN level? Maybe INFO since the user explicitly told us it's ok?

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopBuilder.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    public class LoopBuilder2 {
+        private final Collection<Vertex> initialState;
+        private final Function<Plate, BoolVertex> conditionFunction;
+        private final Map<VertexLabel, VertexLabel> customMappings;
+        private final int maxLoopCount;
+        private final boolean throwWhenMaxCountIsReached;
+        private final VertexLabel VALUE_OUT_WHEN_ALWAYS_TRUE_LABEL = new VertexLabel("loop_value_out_when_always_true");
+        private final VertexLabel VALUE_IN_WHEN_ALWAYS_TRUE_LABEL = PlateBuilder.proxyFor(VALUE_OUT_WHEN_ALWAYS_TRUE_LABEL);
```
gordoncaleb: Does the use of these labels cause problems if you have two loops in a single network?

keanu-project/src/test/java/io/improbable/keanu/plating/LoopTest.java
```diff
@@ -0,0 +1,167 @@
...
+    public void itThrowsIfYouPassInMultipleBaseCaseVertexesAndDontLabelTheOutput(){
+        expectedException.expect(VertexLabelException.class);
+        expectedException.expectMessage("You must pass in a base case, i.e. a vertex labeled as Loop.VALUE_OUT_LABEL");
+        Loop.startingFrom(ConstantVertex.of(0.), ConstantVertex.of(1.))
+            .whilst(alwaysTrue)
+            .apply(increment);
+    }
+
+    @Test
+    public void youCanLoopUntilAConditionIsTrue() {
```
gordoncaleb: Can we add a 2nd loop test? 
```
        Loop loop = Loop
            .startingFrom(startValue)
            .dontThrowWhenMaxCountIsReached()
            .whilst(alwaysTrue)
            .apply(increment);

        Loop loop2 = Loop
            .startingFrom((Vertex<?>) loop.getOutput())
            .dontThrowWhenMaxCountIsReached()
            .whilst(alwaysTrue)
            .apply(increment);

        Vertex<?> output = loop2.getOutput();

        new BayesianNetwork(output.getConnectedGraph());
```

keanu-project/src/main/java/io/improbable/keanu/plating/loop/Loop.java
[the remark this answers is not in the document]
migwellian: Yes, that's the reason. It enforces it at compile time, which is better IMO.

Happy to put it out for team discussion.

keanu-project/src/main/java/io/improbable/keanu/plating/loop/Loop.java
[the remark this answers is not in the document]
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopBuilder.java
```diff
@@ -0,0 +1,180 @@
...
+import io.improbable.keanu.vertices.Vertex;
+import io.improbable.keanu.vertices.VertexLabel;
+import io.improbable.keanu.vertices.VertexLabelException;
+import io.improbable.keanu.vertices.bool.BoolVertex;
+import io.improbable.keanu.vertices.bool.nonprobabilistic.BoolProxyVertex;
+import io.improbable.keanu.vertices.dbl.DoubleVertex;
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.DoubleProxyVertex;
+import io.improbable.keanu.vertices.generic.nonprobabilistic.If;
+
+public class LoopBuilder {
```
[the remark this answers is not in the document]
migwellian: hyphens - done

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopBuilder.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    /**
+     * An optional method to add custom mappings
+     *
+     * @param inputLabel  - the label assigned to the ProxyVertex in frame t
+     * @param outputLabel - the label assigned to a Vertex in frame t-1 which will become the ProxyVertex's parent
+     * @return self
+     */
+    public LoopBuilder mapping(VertexLabel inputLabel, VertexLabel outputLabel) {
```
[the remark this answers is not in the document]
migwellian: I think the confusion is because you're thinking of it as the input/output to the proxy vertex, while I'm thinking about it as the input/output of (two different) frames.

We could rename `inputLabel` as `proxyLabel` and `outputLabel` as `labelFromPreviousFrame` or `proxysParentLabel`?

It would also help to switch the order of the 2 lines of javadoc, e.g. this is what it would look like without any name changes:

     * @param outputLabel the label assigned to a Vertex in frame t-1 which will become the ProxyVertex's parent
     * @param inputLabel  the label assigned to the ProxyVertex in frame t

keanu-project/src/main/java/io/improbable/keanu/plating/loop/Loop.java
migwellian: good call

keanu-project/src/main/java/io/improbable/keanu/plating/loop/Loop.java
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/plating/loop/Loop.java
migwellian: fair enough - done

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopBuilder.java
migwellian: no, because everything is scoped by a unique plate ID, and all this is needed for is to join together vertices in adjacent plates.

keanu-project/src/test/java/io/improbable/keanu/plating/LoopTest.java
migwellian: That's a really good suggestion because it highlights the fact that VertexLabels are not right for Plates. You need to label the output of `loop` is the input to `loop2`, but then you have 2 vertices with the same label so the BayesNet throws.

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopBuilder.java
migwellian: I've changed it now so that Plates keeps its own mapping from VertexLabel to Vertex and doesn't rely on the Vertex's member variable VertexLabel.

keanu-project/src/test/java/io/improbable/keanu/plating/LoopTest.java
migwellian: Done now, after redesigning Plates so that it doesn't rely on the Vertex's label.

keanu-project/src/main/java/io/improbable/keanu/plating/loop/Loop.java
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopBuilder.java
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopBuilder.java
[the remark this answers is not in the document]
migwellian: I've done the rename

keanu-project/src/main/java/io/improbable/keanu/plating/Plates.java
```diff
@@ -31,4 +31,12 @@ public void add(Plate p) {
     public Iterator<Plate> iterator() {
         return containedPlates.iterator();
     }
+
+    public Plate getLastPlate() {
+        try {
+            return this.asList().get(this.size() - 1);
+        } catch (ArrayIndexOutOfBoundsException e) {
```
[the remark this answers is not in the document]
migwellian: That's interesting, I always felt that this way was more elegant. Any particular reason why you favour that way?

I'm just asking out of interest, I'm happy to change it.

migwellian:
Shall I rename `PlateException` to `PlateConstructionException`?

There are currently 3 places where I throw one:
* Vertex must contain a label in order to be added to a plate
* Vertex has already been added to another plate
* Plates is empty (when you call getLastPlate)

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopException.java
```diff
@@ -0,0 +1,11 @@
+package io.improbable.keanu.plating.loop;
+
+public class LoopException extends RuntimeException {
```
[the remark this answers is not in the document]
migwellian: It's only used in one place currently. How about `LoopDidNotTerminateException`?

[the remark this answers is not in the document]
gordoncaleb: 👍

keanu-project/src/main/java/io/improbable/keanu/plating/loop/Loop.java
```diff
@@ -0,0 +1,140 @@
...
+            VertexLabel label = vertex.getLabel();
+            if (label == null) {
+                label = new VertexLabel(String.format("base_case_vertex_%d", vertex.hashCode()));
+            }
+            baseCaseMap.put(label, vertex);
+        }
+        try {
+            return baseCaseMap.build();
+        } catch (IllegalArgumentException e) {
+            throw new VertexLabelException("Duplicate label found in base case");
```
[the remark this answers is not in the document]
gordoncaleb: 👍

keanu-project/src/main/java/io/improbable/keanu/plating/Plates.java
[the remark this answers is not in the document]
gordoncaleb: IMHO exceptions like ArrayIndexOutOfBoundsException and NullPointerException aren't meant to be caught unless you don't have a choice. It probably comes from the older days where you'd get a seg fault or something really nasty if you didn't check the bounds before accessing. I don't think the code should rely (unless there isn't a choice) on being told that there was an exception if the code can be written in a way that is safe to begin with. This might be handled this way in Python or another language but I'd prefer to avoid it here.

keanu-project/src/main/java/io/improbable/keanu/plating/loop/LoopException.java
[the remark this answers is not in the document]
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/plating/loop/Loop.java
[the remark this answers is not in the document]
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/plating/Plates.java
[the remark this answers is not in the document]
migwellian: Thanks, that's a good perspective. I'll look out for it in other parts of my code.

PR #247, opened 2018-09-20
Pass IntegerVertex to ChiSquaredVertex instead of IntegerTensor

icantsayrural:
Because we don't pass tensors to other vertex constructors?

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/ChiSquaredVertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/ChiSquaredVertex.java
@@ -45,8 +45,8 @@ public class ChiSquaredVertex extends DoubleVertex implements ProbabilisticDoubl
      *
      * @param k the number of degrees of freedom
      */
-    public ChiSquaredVertex(IntegerTensor k) {
-        this(k.getShape(), new ConstantIntegerVertex(k));
+    public ChiSquaredVertex(IntegerVertex k) {
+        this(k.getShape(), k);
     }
 
     public ChiSquaredVertex(int k) {


keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/ChiSquaredVertex.java
```diff
@@ -45,8 +45,8 @@ public ChiSquaredVertex(int[] tensorShape, int k) {
      *
      * @param k the number of degrees of freedom
      */
-    public ChiSquaredVertex(IntegerTensor k) {
-        this(k.getShape(), new ConstantIntegerVertex(k));
+    public ChiSquaredVertex(IntegerVertex k) {
+        this(k.getShape(), k);
```
christophernorth: Can we just add a new constructor rather than replacing this one (just in case there are any projects out there using the current semantics)?  I don't see anything fundamentally wrong with providing Tensor versions of the constructors if people just want to pass in a constant?

icantsayrural: Yeah, it depends on why we don't have Tensor versions of constructors for our vertices

migwellian: I agree with Joanne here - it's just weird/wrong that this one Vertex should have a different type of constructor to all the others. Breaking the API is not a big deal right now as we only have a few users - better to do it now than to bloat the API unnecessarily.

christophernorth: You could make the same argument about all the "convenience constructors" that take a constant value of some form rather than just taking a Vertex.  My thinking is that:
a) The fact this one has broken from the standard implies that someone did want a Tensor version of the constructor
b) There doesn't seem to be a good reason *not* to take Tensors and further, given that's what the rest of the API takes - Tensors are actually better than taking scalars...

I'm not suggesting that we, at this point, go through and get rid of all the constructors that take constants but it makes much more logical sense to me to have all Vertices that take constants take Tensors (input and output as the same type).

christophernorth: With that said - I don't feel particularly strongly about this particular case.  Might be worth just pinging the Applied team though so at least they know if they make use of this (or check with Caleb why he added it in the first place).

migwellian: I'm in favour of *all* Vertex types having a Tensor constructor. I'm just opposed to the inconsistency. Also, I reckon this one is only here by accident - otherwise it would already have a constructor that takes a Vertex.

migwellian: In other words, I feel (perhaps counterintuitively) that one Tensor ctor (possibly added accidentally) constitutes API bloat, whereas many Tensor ctors (added deliberately) does not!

christophernorth: Seems we're 100% on the same page then (Tensor constructors being a good thing) - just I'm putting more emphasis on de-risking the change than avoiding this inconsistency.  I'll tick it off for now, but definitely worth just a warning in the release notes at least (so perhaps create a Jira ticket if you don't have one already?).

(Just FWIW, there is a Vertex constructor - it just requires a shape to be specified)

PR #248, opened 2018-09-20
Feature/modellingapi

GeorgeNash:
Beginnings of a higher level modelling API

Lets users easily:

- Create a `LinearModel`
- Train the model (Calculate weights). Predict values of y based on values of x. Calculate a score for how accurate that prediction is (Returns the coefficient of determination).
- These currently include a `LinearRegression` and a `LogisticRegression` (Thanks Stephano for implementing this!)
- I've added a higher level version of the tests so we can compare and contrast

- `LinearModel` currently implements a `Model` interface which doesn't do a lot. I'm hoping this will flesh out as we think about different types of Model and how they will interact


keanu-project/src/test/java/io/improbable/keanu/e2e/regression/LogisticRegressionTest.java
```diff
@@ -0,0 +1,78 @@
...
+    @Test
+    public void testLogisticRegression() {
+        LogisticRegression model = new LogisticRegression(xTrain, yTrain);
+        model = model.fit();
+        double score = model.score(xTest, yTest);
+        assertTrue(score > 0.3);
+    }
+
+    @Test
+    public void testRegularizedLogisticRegression() {
```
migwellian: This test looks the same as the previous one. Is there something we can assert in this test that would not be true in the previous one?

keanu-project/src/main/java/io/improbable/keanu/model/LogisticRegression.java
```diff
@@ -0,0 +1,111 @@
...
+
+    private DoubleVertex computeProbabilities(DoubleTensor x) {
+        int numFeatures = x.getShape()[0];
+        double[] sigma = new double[numFeatures];
+        for (int i = 0; i < numFeatures; i++) {
+            DoubleTensor xColumn = x.slice(0, i);
+            sigma[i] = priorSigma / xColumn.abs().average() / regularization;
+        }
+        DoubleVertex intercept = new GaussianVertex(priorMu, priorSigma);
+        intercept.setLabel(INTERCEPT_LABEL);
```
migwellian: inline

keanu-project/src/main/java/io/improbable/keanu/model/LogisticRegression.java
```diff
@@ -0,0 +1,111 @@
...
+        DoubleVertex weights = new GaussianVertex(priorMu, ConstantVertex.of(sigma));
+        weights.setLabel(WEIGHTS_LABEL);
```
migwellian: inline

keanu-project/src/main/java/io/improbable/keanu/model/LogisticRegression.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Override
+    public DoubleTensor predict(DoubleTensor x) {
+        if (isFit) {
+            DoubleVertex weights = (DoubleVertex) net.getVertexByLabel(WEIGHTS_LABEL);
+            DoubleVertex intercept = (DoubleVertex) net.getVertexByLabel(INTERCEPT_LABEL);
+            return weights.matrixMultiply(ConstantVertex.of(x)).plus(intercept).getValue().sigmoid();
+        } else {
+            throw new RuntimeException("The model must be fit before attempting to predict.");
```
migwellian: can you add a test case for this?

keanu-project/src/main/java/io/improbable/keanu/model/LogisticRegression.java
```diff
@@ -0,0 +1,111 @@
...
+import io.improbable.keanu.tensor.bool.BooleanTensor;
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+import io.improbable.keanu.vertices.ConstantVertex;
+import io.improbable.keanu.vertices.VertexLabel;
+import io.improbable.keanu.vertices.bool.probabilistic.BernoulliVertex;
+import io.improbable.keanu.vertices.dbl.DoubleVertex;
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.ConstantDoubleVertex;
+import io.improbable.keanu.vertices.dbl.probabilistic.GaussianVertex;
+
+public class LogisticRegression implements LinearModel {
```
migwellian: It doesn't seem right that this class is quite different to LinearRegression. Logistic regression is effectively just linear regression with a sigmoid function wrapped around it. Can we do some more code reuse between the two?

keanu-project/src/main/java/io/improbable/keanu/model/LinearRegression.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @Override
+    public BayesianNetwork construct() {
+        int numberOfFeatures = x.getShape()[0];
+        int[] shapeOfWeights = new int[]{1, numberOfFeatures};
+        DoubleVertex weights = new GaussianVertex(shapeOfWeights, MU_OF_WEIGHTS, sigmaOnPrior).setLabel(WEIGHTS_LABEL);
+        DoubleVertex intercept = new GaussianVertex(MU_OF_WEIGHTS, sigmaOnPrior).setLabel(INTERCEPT_LABEL);
+        DoubleVertex xMu = weights.getValue().isScalar() ? weights.times(ConstantVertex.of(x)) : weights.matrixMultiply(ConstantVertex.of(x));
+        DoubleVertex yVertex = new GaussianVertex(xMu.plus(intercept), sigmaOnPrior).setLabel(Y_OBSERVATION_LABEL);
+        net = new BayesianNetwork(yVertex.getConnectedGraph());
```
migwellian: `return new BayesianNetwork(yVertex.getConnectedGraph());` - that way you can make `net` final

keanu-project/src/main/java/io/improbable/keanu/model/LinearRegression.java
```diff
@@ -0,0 +1,85 @@
...
+
+    public LinearRegression(DoubleTensor x, DoubleTensor y) {
+        this(x, y, DEFAULT_PRIOR_ON_WEIGHTS);
+    }
+
+    public LinearRegression(DoubleTensor x, DoubleTensor y, double sigmaOnPrior) {
+        this.x = x;
+        this.y = y;
+        this.sigmaOnPrior = sigmaOnPrior;
+        construct();
```
migwellian: `this.net = construct()` - that way you can make `net` final

keanu-project/src/main/java/io/improbable/keanu/model/LinearRegression.java
```diff
@@ -0,0 +1,85 @@
...
+    private static final double MU_OF_WEIGHTS = 0.0;
+    private static final VertexLabel WEIGHTS_LABEL = new VertexLabel("weights");
+    private static final VertexLabel INTERCEPT_LABEL = new VertexLabel("intercept");
+    private static final VertexLabel Y_OBSERVATION_LABEL = new VertexLabel("y");
+
+    private DoubleTensor x;
+    private DoubleTensor y;
+    private BayesianNetwork net;
+    private boolean isFit = false;
+    private double sigmaOnPrior;
```
migwellian: you can make all of these `final` except for `fit`

keanu-project/src/main/java/io/improbable/keanu/model/LinearRegression.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Override
+    public DoubleTensor predict(DoubleTensor x) {
+        if (isFit) {
+            DoubleVertex weights = (DoubleVertex) net.getVertexByLabel(WEIGHTS_LABEL);
+            DoubleVertex intercept = (DoubleVertex) net.getVertexByLabel(INTERCEPT_LABEL);
+            return weights.getValue().times(x).plus(intercept.getValue());
+        } else {
+            throw new RuntimeException("The model must be fit before attempting to predict.");
```
migwellian: can you add a test case for this?

keanu-project/src/main/java/io/improbable/keanu/model/LinearRegression.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }
+
+    public DoubleVertex getWeights() {
```
migwellian: unused - remove?

keanu-project/src/main/java/io/improbable/keanu/model/LinearRegression.java
```diff
@@ -0,0 +1,85 @@
...
+        return (DoubleVertex) net.getVertexByLabel(WEIGHTS_LABEL);
+    }
+
+    public DoubleVertex getIntercept() {
+        return (DoubleVertex) net.getVertexByLabel(INTERCEPT_LABEL);
+    }
+
+    public BayesianNetwork getNet() {
```
migwellian: unused - remove?

keanu-project/src/main/java/io/improbable/keanu/model/LogisticRegression.java
```diff
@@ -0,0 +1,111 @@
...
+            weights.times(xVertex).plus(intercept).sigmoid() :
+            weights.matrixMultiply(xVertex).plus(intercept).sigmoid();
+    }
+
+    @Override
+    public LogisticRegression fit() {
+        BooleanTensor observedOutcomes = y.greaterThan(0.5);
+        net.getVertexByLabel(OUTCOMES_LABEL).observe(observedOutcomes);
+        GradientOptimizer optimizer = GradientOptimizer.of(net);
+        optimizer.maxAPosteriori();
```
migwellian: why do we do `maxAPosteriori` for `LogisticRegression` but `maxLikelihood` for `LinearRegression`?

keanu-project/src/main/java/io/improbable/keanu/model/LinearRegression.java
```diff
@@ -0,0 +1,85 @@
...
+        construct();
+    }
+
+    @Override
+    public BayesianNetwork construct() {
+        int numberOfFeatures = x.getShape()[0];
+        int[] shapeOfWeights = new int[]{1, numberOfFeatures};
+        DoubleVertex weights = new GaussianVertex(shapeOfWeights, MU_OF_WEIGHTS, sigmaOnPrior).setLabel(WEIGHTS_LABEL);
+        DoubleVertex intercept = new GaussianVertex(MU_OF_WEIGHTS, sigmaOnPrior).setLabel(INTERCEPT_LABEL);
+        DoubleVertex xMu = weights.getValue().isScalar() ? weights.times(ConstantVertex.of(x)) : weights.matrixMultiply(ConstantVertex.of(x));
```
migwellian: aside: It would be nice if `matrixMultiply` just did the right thing for scalars and you didn't need the branch here.

This pattern comes up in other places too (e.g. MultivariateGaussian)

keanu-project/src/main/java/io/improbable/keanu/model/LinearRegression.java
```diff
@@ -0,0 +1,85 @@
...
+
+import io.improbable.keanu.algorithms.variational.optimizer.gradient.GradientOptimizer;
+import io.improbable.keanu.network.BayesianNetwork;
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+import io.improbable.keanu.vertices.ConstantVertex;
+import io.improbable.keanu.vertices.VertexLabel;
+import io.improbable.keanu.vertices.dbl.DoubleVertex;
+import io.improbable.keanu.vertices.dbl.probabilistic.GaussianVertex;
+
+public class LinearRegression implements LinearModel {
```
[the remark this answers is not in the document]
GeorgeNash: It's a linear regression we've made a few times in the `e2e` tests. Just extended to work for many columns of `x` values

keanu-project/src/main/java/io/improbable/keanu/model/LogisticRegression.java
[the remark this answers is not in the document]
GeorgeNash: From Stefano: https://github.com/improbable/applied-core/blob/logistic-regression/java/src/main/java/io/improbable/applied/models/logistic/LogisticRegression.java

They've pulled out a few models into the `applied-core` repo

keanu-project/src/main/java/io/improbable/keanu/model/LogisticRegression.java
```diff
@@ -0,0 +1,111 @@
...
+    private double regularization;
+
+
+    public LogisticRegression(DoubleTensor x, DoubleTensor y, double regularization) {
+        this.x = x;
+        this.y = y;
+        this.priorMu = DEFAULT_MU;
+        this.priorSigma = DEFAULT_SIGMA;
+        this.regularization = regularization;
+        construct();
```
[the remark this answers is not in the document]
GeorgeNash: nice!!

keanu-project/src/main/java/io/improbable/keanu/model/LogisticRegression.java
```diff
@@ -0,0 +1,111 @@
...
+    private boolean isFit = false;
+    private double regularization;
+
+
+    public LogisticRegression(DoubleTensor x, DoubleTensor y, double regularization) {
+        this.x = x;
+        this.y = y;
+        this.priorMu = DEFAULT_MU;
+        this.priorSigma = DEFAULT_SIGMA;
+        this.regularization = regularization;
```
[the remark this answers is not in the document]
GeorgeNash: Yep! Added the other constructors

keanu-project/src/main/java/io/improbable/keanu/model/LogisticRegression.java
GeorgeNash: My mistake - changed to MAP on both.

keanu-project/src/main/java/io/improbable/keanu/model/LogisticRegression.java
GeorgeNash: Good spot. I've made the `Logistic` call out to `Linear` under the hood and perform the extra sigmoid op's at the end

keanu-project/src/main/java/io/improbable/keanu/model/LogisticRegression.java
```diff
@@ -0,0 +1,111 @@
...
+        if (isFit) {
+            DoubleVertex weights = (DoubleVertex) net.getVertexByLabel(WEIGHTS_LABEL);
+            DoubleVertex intercept = (DoubleVertex) net.getVertexByLabel(INTERCEPT_LABEL);
+            return weights.matrixMultiply(ConstantVertex.of(x)).plus(intercept).getValue().sigmoid();
+        } else {
+            throw new RuntimeException("The model must be fit before attempting to predict.");
+        }
+    }
+
+    public DoubleVertex getWeights() {
```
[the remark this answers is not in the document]
GeorgeNash: I think this will be useful. Added an example in the tests.

keanu-project/src/main/java/io/improbable/keanu/model/LogisticRegression.java
```diff
@@ -0,0 +1,111 @@
...
+        return (DoubleVertex) net.getVertexByLabel(WEIGHTS_LABEL);
+    }
+
+    public DoubleVertex getIntercept() {
```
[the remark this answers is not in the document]
GeorgeNash: Agreed - let's remove until needed.

keanu-project/src/main/java/io/improbable/keanu/model/LogisticRegression.java
```diff
@@ -0,0 +1,111 @@
...
+        return (DoubleVertex) net.getVertexByLabel(INTERCEPT_LABEL);
+    }
+
+    public BayesianNetwork getNet() {
```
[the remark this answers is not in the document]
GeorgeNash: I see this being useful for connecting your `LogisticRegression` model up to a wider model.

keanu-project/src/main/java/io/improbable/keanu/model/LogisticRegression.java
```diff
@@ -0,0 +1,111 @@
...
+        return net;
+    }
+
+    public double getWeight(int index) {
```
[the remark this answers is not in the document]
GeorgeNash: I see this being useful - I imagine prediction / scoring / calculating weights are the most common use cases

keanu-project/src/main/java/io/improbable/keanu/model/LogisticRegression.java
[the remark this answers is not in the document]
GeorgeNash: Think you've commented on the same one twice

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/DiabetesLinearRegression.java
```diff
@@ -47,4 +51,41 @@ public void doesLinearRegressionOnBMI() {
...
+    @Test
+    public void canPredictFutureValuesWithLinearRegression() {
+        Data data = ReadCsv
+            .fromResources("data/datasets/diabetes/diabetes_standardized_training.csv")
+            .asVectorizedColumnsDefinedBy(Data.class)
+            .load(true);
+
+        int sizeOfTestData = 100;
+
+        List<DoubleTensor> splitXData = data.bmi.split(1, (int) data.bmi.getLength() - sizeOfTestData, (int) data.bmi.getLength() - 1);
```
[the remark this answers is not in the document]
GeorgeNash: For sure - I've added that to the tasking in the google doc for this spike.

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/LogisticRegressionTest.java
GeorgeNash: Hmm. Good point. It's designed to make the model perform better on "unseen" data. Upping the regularisation seems to make it worse. I'll have a chat with Stefano....

keanu-project/src/main/java/io/improbable/keanu/model/LinearRegression.java
GeorgeNash: I think this will be used a lot

keanu-project/src/main/java/io/improbable/keanu/model/LinearRegression.java
GeorgeNash: Agreed - let me think about this. Possibly a separate task.

keanu-project/src/main/java/io/improbable/keanu/model/LinearRegression.java
GeorgeNash: This may be used by people to link up their `LinearRegression` model with a wider bayes net

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/LogisticRegressionTest.java
GeorgeNash: Updated the test to ensure that the regularized version has smaller weights (Larger regularization punishes larger weights)

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/LogisticRegressionTest.java
```diff
@@ -0,0 +1,100 @@
...
+import io.improbable.keanu.vertices.dbl.KeanuRandom;
+import io.improbable.keanu.vertices.dbl.probabilistic.GaussianVertex;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import static junit.framework.TestCase.assertTrue;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class LogisticRegressionTest {
```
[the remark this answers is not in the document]
GeorgeNash: Made them quicker - down to < 300ms!

keanu-project/src/main/java/io/improbable/keanu/model/LogisticRegression.java
```diff
@@ -0,0 +1,83 @@
...
+    private final DoubleTensor y;
+    private final double priorSigma;
+    private final double regularization;
+    private final LinearRegression regression;
+    private final BayesianNetwork net;
+
+    public LogisticRegression(DoubleTensor x, DoubleTensor y, double regularization, double priorMu, double priorSigma) {
+        this.y = y;
+        this.priorSigma = priorSigma;
+        this.regularization = regularization;
```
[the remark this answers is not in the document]
GeorgeNash: Yes it's faster but this is only ran once and I think for someone looking at the code seeing `inverseRegularization` will be rather confusing

keanu-project/src/main/java/io/improbable/keanu/model/LinearRegression.java
```diff
@@ -0,0 +1,112 @@
...
+    private static final double DEFAULT_SIGMA = 2.0;
+    private static final VertexLabel WEIGHTS_LABEL = new VertexLabel("weights");
+    private static final VertexLabel INTERCEPT_LABEL = new VertexLabel("intercept");
+    private static final VertexLabel Y_OBSERVATION_LABEL = new VertexLabel("y");
+
+    private final DoubleTensor x;
+    private final DoubleTensor y;
+    private final double priorOnSigma;
+    private final double priorOnMu;
+    private final double[] priorOnSigmaForWeights;
```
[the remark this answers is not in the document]
GeorgeNash: nice - removed

keanu-project/src/main/java/io/improbable/keanu/model/linear/LinearRegression.java
```diff
@@ -0,0 +1,68 @@
...
+
+public class LinearRegression implements LinearModel {
+
+    private static final double DEFAULT_MU = 0.0;
+    private static final double DEFAULT_SIGMA = 2.0;
+
+    private final BayesianNetwork net;
+    private final DoubleTensor y;
+
+    public LinearRegression(DoubleTensor x, DoubleTensor y) {
```
johannespetrat: A bit more of a design question here: Following the idea of 'pre-defined' models in Keanu would we perhaps want to decouple the model design from the data that goes in? This would be more similar to the way models are initialised in scikit-learn and would probably look something like this:
```
Model model = new LinearRegression()
DoubleTensor x = some data
DoubleTensor y = some labels
model.fit(x, y)
```
At the same time, the model structure (i.e. number of coefficients) is determined by the data that comes in. This would justify the currently implementation of the class. Or instead we'd have to only create the priors on the fly when the `fit` function is called.

PR #98, opened 2018-06-19
Add Bayesian Methods For Hackers Examples

alex-remedios-improbable:
I have been hacking together some of the examples from Bayesian methods for hackers chapters 1 and 2.

camdavidsonpilon.github.io/Probabilistic-Programming-and-Bayesian-Methods-for-Hackers/

I have had some unexpected results so have created this PR to help us identify where any issues lie.

Hopefully with some cleaning up, these examples will have some value to others also -- please feel free to provide feedback on style/quality

PR #252, opened 2018-09-24
Fix failing tests for Bayesian Methods for Hackers

Pinpickle:
The tests in #98 were failing in non-obvious ways. After digging into it and talking to @gordoncaleb, we determined that the proposal distribution for the Metropolis Hastings sampler (which was just the prior) was too wide, and made bad proposals, resulting in bad samples.

As the API for making a new proposal distribution is pretty clunky, this just changes the priors so that the proposal distribution is more appropriate.

### DIFF
--- a/keanu-examples/bayesianMethodsForHackers/src/main/java/examples/ChallengerDisaster.java
+++ b/keanu-examples/bayesianMethodsForHackers/src/main/java/examples/ChallengerDisaster.java
@@ -27,8 +27,8 @@ public class ChallengerDisaster {
         // as its proposal distribution. The suggested parameters were too wide, resulting
         // in bad proposals and by extension bad samples.
         // When it is easier to decouple the prior from the proposal distribution, we should revisit this
-        final double betaSigma = tauToSigma(0.01);
-        final double alphaSigma = tauToSigma(0.005);
+        final double betaSigma = convertTauToSigma(0.01);
+        final double alphaSigma = convertTauToSigma(0.005);
 
         GaussianVertex alpha = new GaussianVertex(0, alphaSigma);
         GaussianVertex beta = new GaussianVertex(0, betaSigma);
@@ -54,7 +54,7 @@ public class ChallengerDisaster {
         return cp;
     }
 
-    private static double tauToSigma(double tau) {
+    private static double convertTauToSigma(double tau) {
         return Math.sqrt(1.0 / tau);
     }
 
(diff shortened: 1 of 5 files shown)


keanu-examples/bayesianMethodsForHackers/src/main/java/examples/ChallengerDisaster.java
```diff
@@ -56,6 +54,10 @@ public static ChallengerPosteriors run() {
         return cp;
     }
 
+    private static double tauToSigma(double tau) {
```
gordoncaleb: Nice. `convertTauToSigma`?

keanu-examples/bayesianMethodsForHackers/src/test/java/examples/ChallengerDisasterTest.java
```diff
@@ -22,7 +23,7 @@ public void testWhenChallengerDisasterIsRunThenBetaIsSmallAndPositive() {
         System.out.println("mapBeta " + posteriors.mapBeta);
 
         // assert
-        assertThat(posteriors.mapBeta).isBetween(0.1, 0.4);
-        assertThat(posteriors.mapAlpha).isBetween(-25.0, -5.0);
+        assertThat(posteriors.mapBeta).isCloseTo(0.25, within(0.15));
```
gordoncaleb: Nice work with AssertJ but we usually use Hamcrest. I think AssertJ is here from the original commit by Alex. I'll change it before it goes into develop. Just an FYI :)

keanu-examples/bayesianMethodsForHackers/src/main/java/examples/TextMessaging.java
```diff
@@ -44,28 +47,31 @@ public static TextMessagingResults run() {
...
-        int mostProbableSwitchPoint = posteriorSamples.getIntegerTensorSamples(switchPoint).getScalarMode();
-
-        WriteCsv.asSamples(posteriorSamples, switchPoint)
-            .withHeader("switchpoint")
-            .toFile("switchpoitsamples.csv");
-
-        return new TextMessagingResults(mostProbableSwitchPoint);
+        return new TextMessagingResults(
+            posteriorSamples.getIntegerTensorSamples(switchPoint).getScalarMode(),
+            posteriorSamples.getDoubleTensorSamples(earlyRate).getMode().scalar(),
```
gordoncaleb: Getting the mode of a double might give you some strange behaviour without a bin size. Does this work when using getAverages instead?

Pinpickle:
GitHub thinks the travis checks are still running but they [have](https://travis-ci.org/improbable-research/keanu/builds/432984276?utm_source=github_status&utm_medium=notification) in fact [passed](https://travis-ci.org/improbable-research/keanu/builds/432984253?utm_source=github_status&utm_medium=notification)

So going ahead with merge

PR #256, opened 2018-09-26
Feature/multivariate gaussian proposal

migwellian:
This PR enables you to use a Multivariate Gaussian distribution as a ProposalDistribution for the Metropolis Hastings algorithm. It is a precursor to doing adaptive MCMC but for now, the covariance matrix is pre-defined (by the user) and constant.

So for example you can now do Random Walk MCMC, where your covariance matrix would be diagonal.


keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/MultivariateGaussian.java
```diff
@@ -11,10 +11,11 @@
     private final DoubleTensor mu;
     private final DoubleTensor covariance;
 
-    public static ContinuousDistribution withParameters(DoubleTensor mu, DoubleTensor covariance) {
+    public static MultivariateGaussian withParameters(DoubleTensor mu, DoubleTensor covariance) {
         return new MultivariateGaussian(mu, covariance);
     }
     private MultivariateGaussian(DoubleTensor mu, DoubleTensor covariance) {
+        TensorShapeValidation.checkShapesCanBeMatrixMultiplied(covariance.getShape(), mu.getShape());
```
migwellian: this check was missing previously - there was a check in MultivariateGaussianVertex but no check here.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/ScalarDoubleTensor.java
```diff
@@ -431,11 +431,12 @@ public double product() {
...
     @Override
     public DoubleTensor slice(int dimension, int index) {
-        if (dimension == 0 && index == 0) {
-            return duplicate();
-        } else {
-            throw new IllegalStateException("Slice is only valid for dimension and index zero in a scalar");
+        if (dimension == 0 || dimension == 1) {
+            if (index == 0) {
+                return duplicate();
+            }
```
migwellian: I needed to allow slicing on dimension 1.

Note that our "scalars" are always rank 2 (i.e. their shape is 1x1) so this will not fail.

keanu-project/src/main/java/io/improbable/keanu/vertices/Covariance.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.List;
+
+import org.nd4j.linalg.api.blas.BlasException;
+
+import com.google.common.collect.ImmutableList;
+
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+import io.improbable.keanu.tensor.dbl.ScalarDoubleTensor;
+
+public class Covariance {
```
migwellian: I needed a new class so that I can get a smaller covariance matrix for a subset of the dimensions. This is necessary as the MCMC algorithms can select a subset of vertices as a proposal.

migwellian:
I'm withdrawing this PR as I think we are only going to need a (univariate) Gaussian proposal distribution.

PR #259, opened 2018-09-27
added GaussianProposalDistribution

migwellian:
I withdrew the PR for a MultivariateGaussian proposal distribution, and instead submitted this one for the (univariate) Gaussian.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/proposal/GaussianProposalDistribution.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/proposal/GaussianProposalDistribution.java
@@ -1,0 +1,41 @@
+package io.improbable.keanu.algorithms.mcmc.proposal;
+
+import java.util.Set;
+
+import io.improbable.keanu.distributions.ContinuousDistribution;
+import io.improbable.keanu.distributions.continuous.Gaussian;
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+import io.improbable.keanu.vertices.Probabilistic;
+import io.improbable.keanu.vertices.Vertex;
+import io.improbable.keanu.vertices.dbl.DoubleVertex;
+import io.improbable.keanu.vertices.dbl.KeanuRandom;
+
+public class GaussianProposalDistribution implements ProposalDistribution {
+
+    private final DoubleTensor sigma;
+
+    public GaussianProposalDistribution(DoubleTensor sigma) {
+        this.sigma = sigma;
+    }
+
+    @Override
+    public Proposal getProposal(Set<Vertex> vertices, KeanuRandom random) {
+        Proposal proposal = new Proposal();
+        for (Vertex vertex : vertices) {
+            ContinuousDistribution proposalDistribution = Gaussian.withParameters((DoubleTensor) vertex.getValue(), sigma);
+            proposal.setProposal((DoubleVertex) vertex, proposalDistribution.sample(vertex.getShape(), random));
+        }
+        return proposal;
+    }
+
+    @Override
+    public <T> double logProb(Probabilistic<T> vertex, T ofValue, T givenValue) {
+        if (!(ofValue instanceof DoubleTensor)) {
+            throw new ClassCastException("Only DoubleTensor values are supported - not " + ofValue.getClass().getSimpleName());
+        }
+
+        ContinuousDistribution proposalDistribution = Gaussian.withParameters((DoubleTensor) ofValue, sigma);
+        return proposalDistribution.logProb((DoubleTensor) givenValue).sum();
+    }
+
+}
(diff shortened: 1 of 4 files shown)


keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/proposal/GaussianProposalDistributionTest.java
```diff
@@ -0,0 +1,67 @@
...
+        assertThat(logProb, equalTo(expectedLogProb.sum()));
+    }
+
+    @Test
+    public void theLogProbAtFromIsGaussianAroundTheGivenPoint() {
+        double logProb = proposalDistribution.logProbAtFromGivenTo(proposal);
+        DoubleTensor expectedLogProb = Gaussian.withParameters(proposedState, sigma).logProb(currentState);
+        assertThat(logProb, equalTo(expectedLogProb.sum()));
+    }
+}
```
christophernorth: Nit: do we want a newline here?

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/proposal/GaussianProposalDistributionTest.java
```diff
@@ -0,0 +1,67 @@
...
+    public void setRandomSeed() throws Exception {
+        KeanuRandom.setDefaultRandomSeed(0);
+    }
+
+    @Before
+    public void setUpMocks() throws Exception {
+        when(vertex1.getValue()).thenReturn(DoubleTensor.scalar(currentState.getValue(0)));
+        when(vertex2.getValue()).thenReturn(DoubleTensor.scalar(currentState.getValue(1)));
+
+        proposal = new Proposal();
```
christophernorth: Should the proposal setup be done in its own function (given the name of the function is to do with Mocks and this looks like just the test input).

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastingsTest.java
```diff
@@ -167,6 +169,40 @@ public void samplesFromPriorWithObservedDeterministic() {
...
+            bayesNet,
+            Arrays.asList(A, B),
+            1000
+        );
+
+        double averagePosteriorA = posteriorSamples.getDoubleTensorSamples(A).getAverages().scalar();
+        double averagePosteriorB = posteriorSamples.getDoubleTensorSamples(B).getAverages().scalar();
+
+        double actual = averagePosteriorA + averagePosteriorB;
+        assertEquals(44.0, actual, 0.1);
```
christophernorth: Can you add a quick comment on how you derived the 44.0 value here please?

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/proposal/GaussianProposalDistribution.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public GaussianProposalDistribution(DoubleTensor sigma) {
+        this.sigma = sigma;
+    }
+
+    @Override
+    public Proposal getProposal(Set<Vertex> vertices, KeanuRandom random) {
+        Proposal proposal = new Proposal();
+        for (Vertex vertex : vertices) {
+            ContinuousDistribution proposalDistribution = Gaussian.withParameters((DoubleTensor) vertex.getValue(), sigma);
```
christophernorth: Casts are icky.  I wonder if we could fix up the class hierarchy somewhat to remove the need for this?

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastingsTest.java
migwellian: um... ask @gordoncaleb? I copied it from the test above!

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/proposal/GaussianProposalDistributionTest.java
migwellian: done

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/proposal/GaussianProposalDistributionTest.java
migwellian: The order is important (I need to set up the mocks before the proposal) so they can't go in separate `@Before` methods.

I've renamed the method `setUpProposal` which is more accurate and hopefully addresses your point.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/proposal/GaussianProposalDistribution.java
migwellian: Agreed, casts are icky. The problem here (and many other places) is that we need a collection of vertices of mixed types. 

In fact I wonder if I ought to be filtering out those that are not `DoubleVertex` here. @gordoncaleb - can a proposal contain non-double-values vertices?

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastingsTest.java
gordoncaleb: If you think about the result of the sum of A and B then you can see that the mu of C is also a gaussian of Gaussian(40, 2). Then if you take the log of their pdfs and sum them you get `(-2(x-40)^2)/8 - ((46-x)^2)/2`. If you find the root of the derivative of that (aka maximum) you will find it's 44. You can see this in WolfgramAlpha `D[(-2 (x - 40)^2)/8 - (46 - x)^2/2, x] == 0`. Maybe put this as a comment on the code?

Issue #181, opened 2018-08-20
Create easy way to take x samples


mhutti1:
**Is your feature request related to a problem? Please describe.**
Currently if I need to take 100 samples of a random variable I do this:
```
speedVertex.setValue(new double[100]);
speedVertex.setValue(speedVertex.sample().average());
```
But this is two lines for a rather simple thing.

**Describe the solution you'd like**
Some kind of single line method like
``` speedVertex.sample(100).average()```

PR #261, opened 2018-09-28
Add sampling multiple scalars into a tensor

Pinpickle:
#181 introduced a need to make multiple samples at once easily.

After some discussion with @gordoncaleb, we figured the best way for now is to sample multiple values into a tensor of arbitrary shape. As such, this ability has only been added to tensor vertexes.

This adds a new method, `.sampleScalarValuesAsTensor`, which when given a tensor shape, gives you a tensor of that shape filled with samples.

It only works on vertexes of a scalar shape.

As this can't be generic, all tensor vertices must have code and tests added independently. Because of that, I don't know if this is worthwhile maintaining. So while in review, could people let me know their opinion on merging this in?


keanu-project/src/main/java/io/improbable/keanu/vertices/intgr/IntegerVertex.java
```diff
@@ -179,4 +195,22 @@ public int getValue(int... index) {
...
     }
 
+    public IntegerTensor sampleScalarValuesAsTensor(int[] shape, KeanuRandom random) {
+        if (!TensorShape.isScalar(this.getShape())) {
+            throw new IllegalArgumentException("Vertex to sample must be scalar");
+        }
+
+        final int length = Math.toIntExact(TensorShape.getLength(shape));
+        final int[] samples = new int[length];
+        for (int i = 0; i < length; i += 1) {
```
icantsayrural: small int: use foreach loop?

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/BoolVertex.java
```diff
@@ -91,6 +93,24 @@ public boolean getValue(int... index) {
         return getValue().getValue(index);
     }
 
+    public BooleanTensor sampleScalarValuesAsTensor(int[] shape, KeanuRandom random) {
+        if (!TensorShape.isScalar(this.getShape())) {
```
migwellian: if (!this.getValue().isScalar())

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/BoolVertex.java
```diff
@@ -91,6 +93,24 @@ public boolean getValue(int... index) {
         return getValue().getValue(index);
     }
 
+    public BooleanTensor sampleScalarValuesAsTensor(int[] shape, KeanuRandom random) {
```
migwellian: I think this could be a parameterised method on `Vertex`: 

    public T sample(int[] shape, KeanuRandom random)

Pinpickle: That's the road I went down initially, but things got hairy pretty quickly as the interface to create tensors uses arrays. This is primarily due to generics requiring boxed primitives, but the tensor constructors requiring unboxed primitives.

Finally, as this thing returns a tensor, it doesn't make sense to exist on a vertex that doesn't expose something that can be stored in a tensor (e.g. an enum or a user-defined object).

Richer type systems would let us solve these problems but this is the world we live in.

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/BoolVertex.java
Pinpickle: Can't `getValue` return `null` in some circumstances?

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/BoolVertex.java
Pinpickle: If there's a way we can have this as a generic function regardless of these issues, I'd love to hear it (this duplicated code isn't ideal)

GeorgeNash: We can probably have a go at removing a lot of the shared code here. What we're essentially doing is taking many samples and then reshaping them into a tensor. We can have a method on Vertex that knew how to take many samples and apply a function (in this case to go from tensor to scalar).

```
    public <U> List<U> sampleMany(int n, KeanuRandom random, Function<T, U> func) {
        ArrayList<U> samples = new ArrayList<>();
        for (int i = 0; i < n; i++) {
            samples.add(func.apply(sample(random)));
        }
        return samples;
    }
```

Then we can start doing:

```
    public BooleanTensor sampleScalarValuesAsTensor(int[] shape, KeanuRandom random) {
        if (!TensorShape.isScalar(this.getShape())) {
            throw new IllegalArgumentException("Vertex to sample must be scalar");
        }

        final int length = Math.toIntExact(TensorShape.getLength(shape));
        List<Boolean> samples = sampleMany(length, random, i -> i.scalar());
        return BooleanTensor.create(Booleans.toArray(samples), shape);
    }
```

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/BoolVertex.java
migwellian: Hmm good question. Also, it has side effects as it might call `lazyEval()`.

So your way is better. I wonder if we should wrap this as `this.isScalar()` so that it becomes the obvious choice.

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/BoolVertex.java
migwellian: I think that's a great idea.

I'd probably make a new class for this, rather than putting it into `Vertex`. Also (optional) it could take a `Vertex` rather than a `Function<T, U>` - that way you can put the shape checking into the method too, rather than repeat it 3 times.

    package io.improbable.keanu.vertices;
    
    import java.util.ArrayList;
    import java.util.List;
    
    import io.improbable.keanu.tensor.Tensor;
    import io.improbable.keanu.tensor.TensorShape;
    
    public class ScalarSampler<U> {
    
        public List<U> sample(Vertex<? extends Tensor<U>> vertex, int[] shape, KeanuRandom random) {
            if (!TensorShape.isScalar(vertex.getShape())) {
                throw new IllegalArgumentException("Vertex to sample must be scalar");
            }
    
            final int length = Math.toIntExact(TensorShape.getLength(shape));
            ArrayList<U> samples = new ArrayList<>();
            for (int i = 0; i < length; i += 1) {
                samples.add(vertex.sample(random).scalar());
            }
            return samples;
        }
    }

Pinpickle:
@GeorgeNash @migwellian thanks for your ideas - I ended up going with something in between what the two of you suggested. Feel free to have another look!

gordoncaleb:
I think we should just expose the sample with shape method on the Distributions.

```
public interface Distribution<T> {
    T sample(int[] shape, KeanuRandom random);
    DoubleTensor logProb(T x);
}
```

The sample with shape should be available on our distribution vertices (e.g. GaussianVertex) but not on all operations due to our lacking appropriate semantics.

PR #262, opened 2018-10-01
Add determinant vertex

Pinpickle:
This PR adds support for a determinant vertex. It takes a vertex shaped like a matrix as input, and outputs a scalar determinant.

Reverse auto diff has been implemented on this vertex as well. It's slightly problematic as it involves the inverse of our input matrix, which fails when the determinant is 0 (despite the gradient being well-defined at that point). This seems to be the expected way to perform such a calculation, and none of the literature even addresses this as a problem. @gordoncaleb and I couldn't think of a case where this would be a problem, and I added some random testing to verify this at least in a small case.

Forward differentiation has not been implemented.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
@@ -33,6 +33,7 @@ import io.improbable.keanu.vertices.dbl.nonprobabilistic.operators.unary.ExpVert
 import io.improbable.keanu.vertices.dbl.nonprobabilistic.operators.unary.FloorVertex;
 import io.improbable.keanu.vertices.dbl.nonprobabilistic.operators.unary.LogGammaVertex;
 import io.improbable.keanu.vertices.dbl.nonprobabilistic.operators.unary.LogVertex;
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.operators.unary.MatrixDeterminantVertex;
 import io.improbable.keanu.vertices.dbl.nonprobabilistic.operators.unary.MatrixInverseVertex;
 import io.improbable.keanu.vertices.dbl.nonprobabilistic.operators.unary.ReshapeVertex;
 import io.improbable.keanu.vertices.dbl.nonprobabilistic.operators.unary.RoundVertex;
@@ -75,6 +76,17 @@ public abstract class DoubleVertex extends Vertex<DoubleTensor> implements Doubl
 
     public DoubleVertex matrixInverse() {
         return new MatrixInverseVertex(this);
+    }
+
+    /**
+     * Get a vertex that is the determinant of this vertex.
+     * <p>
+     * Gradient calculations (and thus gradient-based optimisations) will fail if the matrixDeterminant is 0.
+     *
+     * @return a new vertex with the value of the determinant of this vertex
+     */
+    public DoubleVertex matrixDeterminant() {
+        return new MatrixDeterminantVertex(this);
     }
 
     public DoubleVertex divideBy(DoubleVertex that) {
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
@@ -78,13 +78,6 @@ public abstract class DoubleVertex extends Vertex<DoubleTensor> implements Doubl
         return new MatrixInverseVertex(this);
     }
 
-    /**
-     * Get a vertex that is the determinant of this vertex.
-     * <p>
-     * Gradient calculations (and thus gradient-based optimisations) will fail if the matrixDeterminant is 0.
-     *
-     * @return a new vertex with the value of the determinant of this vertex
-     */
     public DoubleVertex matrixDeterminant() {
         return new MatrixDeterminantVertex(this);
     }
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
@@ -13,16 +13,30 @@ import io.improbable.keanu.vertices.dbl.nonprobabilistic.diff.PartialDerivatives
 /**
  * A vertex that takes on the value of the matrixDeterminant of the value of its input matrix
  * <p>
- * Gradient calculations (and thus gradient-based optimisations) will fail if the matrixDeterminant is 0.
+ * Gradient calculations (and thus gradient-based optimisations) will fail if the determinant is 0.
+ * <p>
+ * Reverse derivatives are implemented according to https://www.cs.ox.ac.uk/files/723/NA-08-01.pdf
+ * <p>
+ * Forward mode differentiation is not implemented due to requiring a tensor trace, which is not yet implemented
  */
 public class MatrixDeterminantVertex extends DoubleUnaryOpVertex {
     public MatrixDeterminantVertex(DoubleVertex vertex) {
         super(Tensor.SCALAR_SHAPE, vertex);
+        assertInputValueIsSquareMatrix(vertex.getValue());
     }
 
-    @Override
-    protected DoubleTensor op(DoubleTensor value) {
-        return DoubleTensor.scalar(value.determinant());
+    private static void assertInputValueIsSquareMatrix(DoubleTensor value) {
+        final IllegalArgumentException exception = new IllegalArgumentException("Input tensor must be a square matrix");
+
+        if (value == null) {
+            throw exception;
+        }
+
+        final int[] shape = value.getShape();
+
+        if ((shape.length != 2) || (shape[0] != shape[1])) {
+            throw exception;
+        }
     }
 
     @Override
@@ -30,14 +44,15 @@ public class MatrixDeterminantVertex extends DoubleUnaryOpVertex {
         throw new UnsupportedOperationException();
     }
 
-    /**
-     * Implemented according to https://www.cs.ox.ac.uk/files/723/NA-08-01.pdf
-     *
-     * @param derivativeOfOutputsWithRespectToSelf derivative of outputs with respect to self
-     * @return derivative of outputs with respect to inputs
-     */
+    @Override
+    protected DoubleTensor op(DoubleTensor value) {
+        assertInputValueIsSquareMatrix(value);
+        return DoubleTensor.scalar(value.determinant());
+    }
+
     @Override
     public Map<Vertex, PartialDerivatives> reverseModeAutoDifferentiation(PartialDerivatives derivativeOfOutputsWithRespectToSelf) {
+        assertInputValueIsSquareMatrix(inputVertex.getValue());
         DoubleTensor inverseTranspose = inputVertex.getValue().transpose().matrixInverse();
 
         PartialDerivatives derivativeOfOutputsWithRespectToInputs = derivativeOfOutputsWithRespectToSelf
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
@@ -1,0 +1,51 @@
+package io.improbable.keanu.vertices.dbl.nonprobabilistic.operators.unary;
+
+import java.util.Collections;
+import java.util.Map;
+
+import io.improbable.keanu.tensor.Tensor;
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+import io.improbable.keanu.vertices.Vertex;
+import io.improbable.keanu.vertices.dbl.DoubleVertex;
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.diff.DualNumber;
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.diff.PartialDerivatives;
+
+/**
+ * A vertex that takes on the value of the matrixDeterminant of the value of its input matrix
+ * <p>
+ * Gradient calculations (and thus gradient-based optimisations) will fail if the matrixDeterminant is 0.
+ */
+public class MatrixDeterminantVertex extends DoubleUnaryOpVertex {
+    public MatrixDeterminantVertex(DoubleVertex vertex) {
+        super(Tensor.SCALAR_SHAPE, vertex);
+    }
+
+    @Override
+    protected DoubleTensor op(DoubleTensor value) {
+        return DoubleTensor.scalar(value.determinant());
+    }
+
+    @Override
+    protected DualNumber dualOp(DualNumber dualNumber) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public Map<Vertex, PartialDerivatives> reverseModeAutoDifferentiation(PartialDerivatives derivativeOfOutputsWithRespectToSelf) {
+        // As this is the inverse, this will fail if the matrixDeterminant is zero
+        DoubleTensor inverseTranspose = inputVertex.getValue().transpose().matrixInverse();
+
+        // Reverse mode auto diff of matrix matrixDeterminant: https://www.cs.ox.ac.uk/files/723/NA-08-01.pdf
+        // Using variables from the paper
+        // Abar = derivativeOfOutputsWithRespectToInputs (what we want)
+        // A = inputVertex
+        // C = det(inputVertex)
+        // Cbar = derivativeOfOutputsWithRespectToSelf
+        // Abar_i,j = C * Cbar_i,j * inverse(transpose(A))
+        PartialDerivatives derivativeOfOutputsWithRespectToInputs = derivativeOfOutputsWithRespectToSelf
+            .multiplyBy(inputVertex.getValue().determinant())
+            .multiplyAlongWrtDimensions(inverseTranspose, this.inputVertex.getShape());
+
+        return Collections.singletonMap(inputVertex, derivativeOfOutputsWithRespectToInputs);
+    }
+}
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
@@ -26,16 +26,12 @@ public class MatrixDeterminantVertex extends DoubleUnaryOpVertex {
     }
 
     private static void assertInputValueIsSquareMatrix(DoubleTensor value) {
-        final IllegalArgumentException exception = new IllegalArgumentException("Input tensor must be a square matrix");
-
-        if (value == null) {
-            throw exception;
-        }
-
         final int[] shape = value.getShape();
+        final boolean isMatrix = shape.length == 2;
+        final boolean isSquare = isMatrix && shape[0] == shape[1];
 
-        if ((shape.length != 2) || (shape[0] != shape[1])) {
-            throw exception;
+        if (!isMatrix || !isSquare) {
+            throw new IllegalArgumentException("Input tensor must be a square matrix");
         }
     }
 
@@ -46,13 +42,11 @@ public class MatrixDeterminantVertex extends DoubleUnaryOpVertex {
 
     @Override
     protected DoubleTensor op(DoubleTensor value) {
-        assertInputValueIsSquareMatrix(value);
         return DoubleTensor.scalar(value.determinant());
     }
 
     @Override
     public Map<Vertex, PartialDerivatives> reverseModeAutoDifferentiation(PartialDerivatives derivativeOfOutputsWithRespectToSelf) {
-        assertInputValueIsSquareMatrix(inputVertex.getValue());
         DoubleTensor inverseTranspose = inputVertex.getValue().transpose().matrixInverse();
 
         PartialDerivatives derivativeOfOutputsWithRespectToInputs = derivativeOfOutputsWithRespectToSelf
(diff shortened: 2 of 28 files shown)


keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
```diff
@@ -0,0 +1,51 @@
...
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public Map<Vertex, PartialDerivatives> reverseModeAutoDifferentiation(PartialDerivatives derivativeOfOutputsWithRespectToSelf) {
+        // As this is the inverse, this will fail if the matrixDeterminant is zero
+        DoubleTensor inverseTranspose = inputVertex.getValue().transpose().matrixInverse();
+
+        // Reverse mode auto diff of matrix matrixDeterminant: https://www.cs.ox.ac.uk/files/723/NA-08-01.pdf
+        // Using variables from the paper
```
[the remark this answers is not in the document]
icantsayrural: Looks useful, might be better to put it in a javadoc for the method?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
```diff
@@ -75,6 +76,17 @@ public DoubleVertex matrixInverse() {
         return new MatrixInverseVertex(this);
     }
 
+    /**
```
gordoncaleb: The method name here should be enough to describe what's going on. I'd just leave the javadoc for the class itself.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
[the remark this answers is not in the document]
gordoncaleb: Method javadocs are better than inline comments. With javadoc a user can see the comments in the docs instead of having to dig through the source.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    protected DualNumber dualOp(DualNumber dualNumber) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public Map<Vertex, PartialDerivatives> reverseModeAutoDifferentiation(PartialDerivatives derivativeOfOutputsWithRespectToSelf) {
+        // As this is the inverse, this will fail if the matrixDeterminant is zero
```
gordoncaleb: This comment should be on the matrixInverse() method instead of here I think. You could argue that that's a property of the function instead of a quirk of the code that needs documenting.

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertexTest.java
```diff
@@ -0,0 +1,73 @@
...
+    @Test
+    public void canOptimiseOutOfTheBoxEvenWithRiskOfSingularMatrices() {
+        // The standard method for calculating the gradient of the determinant fails when the matrixDeterminant is 0.
+        // Regardless, the gradient is defined when the matrixDeterminant is 0.
+        // We run the optimizer many times to create some confidence that we won't run into problems in a real-world
+        // use-case.
+        final int runCount = 100;
+        IntStream.range(0, runCount).forEach(i -> {
+            int[] shape = new int[]{2, 2};
+            DoubleVertex input = new GaussianVertex(shape, Math.random() * 2 - 4, Math.random() * 2 + 0.1);
```
gordoncaleb: Nondeterministic tests are really a pain. Think about the edge cases this random test might find and then test them explicitly and deterministically.

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertexTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Test
+    public void canDifferentiateWrtTensor() {
+        final int[] shape = new int[]{2, 2};
+        final DoubleVertex input = new UniformVertex(shape, 0, 10);
+        input.setValue(DoubleTensor.create(new double[]{1, 2, 3, 4}, shape));
+        final DoubleVertex output = input.matrixDeterminant().times(input);
+
+        // For calculation of derivative, see
+        // https://www.wolframalpha.com/input/?i=x+%3D+1;+y+%3D+2;+z+%3D+3;+w+%3D+4;+d%2Fdx+(%7B%7Bx,y%7D,%7Bz,w%7D%7D+*+det(%7B%7Bx,y%7D,%7Bz,w%7D%7D))
```
gordoncaleb: Awesome work with the wolframalpha link!

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertexTest.java
Pinpickle: That's a good point. This loop is also making the test take much longer than it needs to.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    protected DoubleTensor op(DoubleTensor value) {
+        return DoubleTensor.scalar(value.determinant());
+    }
+
+    @Override
+    protected DualNumber dualOp(DualNumber dualNumber) {
+        throw new UnsupportedOperationException();
```
christophernorth: Potentially a few reasons to implement this:
1) To provide extra confidence to our gradient tests (we can check that forward = reverse etc)
2) Given we have a limitation in reverse - might be best if people can still get the gradient if they hit the 0 case...
3) There will be certain problems where forward accumulation is quicker.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
```diff
@@ -0,0 +1,49 @@
...
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.diff.DualNumber;
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.diff.PartialDerivatives;
+
+/**
+ * A vertex that takes on the value of the matrixDeterminant of the value of its input matrix
+ * <p>
+ * Gradient calculations (and thus gradient-based optimisations) will fail if the matrixDeterminant is 0.
+ */
+public class MatrixDeterminantVertex extends DoubleUnaryOpVertex {
+    public MatrixDeterminantVertex(DoubleVertex vertex) {
```
GeorgeNash: We've only implemented determinant for rank 2 tensors right? Let's put in an check to ensure `vertex` is a matrix and throw an `IllegalArgumentException` if not

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertexTest.java
```diff
@@ -0,0 +1,73 @@
...
+    public void canOptimiseOutOfTheBoxStartingAtZero() {
+        assertOptimizerWorks(0);
+    }
+
+    private void assertReverseAutoDiffMatches(DoubleVertex output, DoubleVertex input, DoubleTensor expectedDerivative) {
+        final DoubleTensor derivative = Differentiator.reverseModeAutoDiff(output, input).withRespectTo(input);
+        Assert.assertThat(derivative, TensorMatchers.elementwiseEqualTo(expectedDerivative));
+    }
+
+    private void assertOptimizerWorks(double inputGaussianMu) {
```
GeorgeNash: This doesn't really mean anything - can we call it `assertOptimizerWorksWithDeterminant`?

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertexTest.java
```diff
@@ -0,0 +1,73 @@
...
+            Optimizer.of(net).maxLikelihood();
+            Assert.assertEquals(input.getValue().determinant(), 2.2, 0.1);
+        });
+    }
+
+    private void assertReverseAutoDiffMatches(DoubleVertex output, DoubleVertex input, DoubleTensor expectedDerivative) {
+        final DoubleTensor derivative = Differentiator.reverseModeAutoDiff(output, input).withRespectTo(input);
+        Assert.assertThat(derivative, TensorMatchers.elementwiseEqualTo(expectedDerivative));
+    }
+}
```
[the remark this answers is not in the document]
GeorgeNash: `finiteDifferenceMatchesGradient`

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertexTest.java
```diff
@@ -0,0 +1,73 @@
...
+import io.improbable.keanu.network.BayesianNetwork;
+import io.improbable.keanu.tensor.TensorMatchers;
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+import io.improbable.keanu.vertices.dbl.Differentiator;
+import io.improbable.keanu.vertices.dbl.DoubleVertex;
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.ConstantDoubleVertex;
+import io.improbable.keanu.vertices.dbl.probabilistic.GaussianVertex;
+import io.improbable.keanu.vertices.dbl.probabilistic.UniformVertex;
+
+public class MatrixDeterminantVertexTest {
```
GeorgeNash: Nice tests! In all your tests the `determinant` is never the end of the graph and is never the thing we are calculating the gradient `of`. Can we add in a test where the determinant is at the end?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
```diff
@@ -0,0 +1,49 @@
...
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+import io.improbable.keanu.vertices.Vertex;
+import io.improbable.keanu.vertices.dbl.DoubleVertex;
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.diff.DualNumber;
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.diff.PartialDerivatives;
+
+/**
+ * A vertex that takes on the value of the matrixDeterminant of the value of its input matrix
+ * <p>
+ * Gradient calculations (and thus gradient-based optimisations) will fail if the matrixDeterminant is 0.
```
migwellian: Can you demonstrate this in a test?

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertexTest.java
```diff
@@ -0,0 +1,73 @@
...
+    @Test
+    public void calculatesDeterminant() {
+        final DoubleVertex input = new ConstantDoubleVertex(DoubleTensor.create(new double[]{1, 2, 3, 4}, 2, 2));
```
migwellian: It's probably worth having a test that uses a matrix larger than 2x2

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertexTest.java
```diff
@@ -0,0 +1,73 @@
...
```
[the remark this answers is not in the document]
Pinpickle: I think that makes reading this harder (in the case of 4 items), as we have to call `.arange(1,5).reshape(shape)`. The exclusivity of the range in particular makes this harder to read.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
Pinpickle: This was a good idea and @gordoncaleb and I spent some time investigating this. At this stage it's too much effort to implement. I'll make this clearer.

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertexTest.java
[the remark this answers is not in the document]
christophernorth: That's the badger

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
christophernorth: Cool - thanks.

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertexTest.java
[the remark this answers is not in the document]
Pinpickle: This is muuuch nicer

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
Pinpickle: Yes good catch!

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
Pinpickle: Went back and forth on whether to do this a few times. In the end I decided to. If it's in our javadoc, it's part of the API contract, so it should be tested for. Unfortunately due to the way forward mode differentiation is implemented, I can't test for that not being implemented (I don't think)

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
```diff
@@ -0,0 +1,64 @@
...
+    public MatrixDeterminantVertex(DoubleVertex vertex) {
+        super(Tensor.SCALAR_SHAPE, vertex);
+        assertInputValueIsSquareMatrix(vertex.getValue());
+    }
+
+    private static void assertInputValueIsSquareMatrix(DoubleTensor value) {
+        final IllegalArgumentException exception = new IllegalArgumentException("Input tensor must be a square matrix");
+
+        if (value == null) {
+            throw exception;
```
gordoncaleb: `Preconditions.checkNotNull(...)`

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        final int[] shape = value.getShape();
+
+        if ((shape.length != 2) || (shape[0] != shape[1])) {
```
gordoncaleb: Maybe use two booleans `isNotMatrix` and `isNotSquare` and clean this predicate up.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
```diff
@@ -0,0 +1,64 @@
...
+ * Forward mode differentiation is not implemented due to requiring a tensor trace, which is not yet implemented
+ */
+public class MatrixDeterminantVertex extends DoubleUnaryOpVertex {
+    public MatrixDeterminantVertex(DoubleVertex vertex) {
+        super(Tensor.SCALAR_SHAPE, vertex);
+        assertInputValueIsSquareMatrix(vertex.getValue());
+    }
+
+    private static void assertInputValueIsSquareMatrix(DoubleTensor value) {
+        final IllegalArgumentException exception = new IllegalArgumentException("Input tensor must be a square matrix");
```
gordoncaleb: No need to create this on every call to this assert...

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
Pinpickle: Going to drop this as we discussed as null checking isn't something we really do in the codebase

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertexTest.java
```diff
@@ -0,0 +1,111 @@
...
+        Assert.assertThat(input.matrixDeterminant().getValue(), TensorMatchers.isScalarWithValue(-2d));
```
migwellian: You can drop the `TensorMatchers.` for better readability. 

i.e. then you will have
`import io.improbable.keanu.tensor.TensorMatchers.isScalarWithValue`
instead of
`import io.improbable.keanu.tensor.TensorMatchers`

migwellian: likewise you can drop the `Assert.`

migwellian: You can use `VertexMatchers#hasValue` instead:
`assertThat(input.matrixDeterminant(), hasValue(-2d));`

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    private static void assertInputValueIsSquareMatrix(DoubleTensor value) {
+        final int[] shape = value.getShape();
+        final boolean isMatrix = shape.length == 2;
+        final boolean isSquare = isMatrix && shape[0] == shape[1];
+
+        if (!isMatrix || !isSquare) {
+            throw new IllegalArgumentException("Input tensor must be a square matrix");
+        }
```
migwellian: It would be better to separate out the 2 failure conditions into separate error messages. So you end up with:

    if (shape.length != 2) {
            throw new IllegalArgumentException("Input tensor must be a matrix");
    } else if (shape[0] != shape[1]) {
            throw new IllegalArgumentException("Input matrix must be square");
    }

migwellian: The natural place for this method to live is `TensorShapeValidation`

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertexTest.java
```diff
@@ -0,0 +1,111 @@
...
+    @Test
+    public void calculatesDeterminantOnLargeMatrix() {
+        final double[] values = new double[]{
+            1, 8, 4, 5, 6,
+            1, 100, 232, 4, 15,
+            -10, 3, 57, 68, 10,
+            7, 8, 9, 20, 10,
+            12, 32, 43, -2, 5};
+        final DoubleVertex input = new ConstantDoubleVertex(DoubleTensor.create(values, 5, 5));
+        Assert.assertThat(input.matrixDeterminant().getValue(), TensorMatchers.isScalarWithValue(Matchers.closeTo(6120880d, 1e-5)));
```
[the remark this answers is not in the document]
Pinpickle: Determinant calculations end up with big numbers for larger matrices - this calculation in particular ended up being not exact (so an exact match would fail)

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/MatrixDeterminantVertex.java
Pinpickle: Good call - moved and added tests

PR #266, opened 2018-10-02
Fix setWithMaskInPlace so that it doesn't accept mask with different length than the tensor

icantsayrural:


icantsayrural:
Fixes #270

PR #268, opened 2018-10-03
[RFC] Add automatic code formatting (Google style)

Pinpickle:
This is a possible illustration of how we might add automatic code formatting to our workflow.

This is part of a group of PRs to demonstrate different implementations. This one uses a gradle plugin ([Spotless](https://github.com/diffplug/spotless/tree/master/plugin-gradle)) and a particular code formatter ([google-java-format](https://github.com/google/google-java-format)) configured to the Google style. The primary difference from the Android style is that indents are 2 spaces. In my opinion this increases readability, but requires IDE configuration for any sensible developer experience.

I'm open to suggestions, but google-java-format is the only formatter I could find that was opinionated enough to prevent bikeshedding.

Wildcard imports can't be fixed automatically, so there's a rule to check for them and prompts the user to fix it.

See the `README.md` and `build.gradle` in the diff for specifics. Everything else is the code formatting itself, which allows you to determine if you like it or not.


Pinpickle:
@migwellian, @gordoncaleb pointed out to me that IntelliJ respects editorconfig by default, so we can change ours to 2 spaces and default intelliJ will be mostly passable.

It's also an option to install the formatter plugin (instructions in readme), but we can't expect developers to do that.

christophernorth:
I'm really not a fan of 2 space indents personally - I think they only lead to badly formatted code in the long run (I'm more on the side of the fence as the linux kernel guys, but I think 8 spaces is a tad too much sometimes).

PR #269, opened 2018-10-03
[Do not merge] An example of bad formatting failing

Pinpickle:
This PR includes changes that violate our formatting rules. It's an example of how tests will fail in this case.


Pinpickle:
You heard it from Travis, it's failing:

image

migwellian:
Does it tell you on the command line what the violations were?

Pinpickle:
@migwellian yes. If you run `gradle check`, it will give you the same message as screenshot above

migwellian:
Yes, but what I meant was: does it explain what's wrong, i.e. which rule has been violated.

I tried it out myself - the answer is no, but I guess that doesn't matter as the user doesn't really need to know/care about the details, they just need to run `spotlessApply` (which the error message instructs them to do).

And of course in the case of your custom rule, you get the custom error message.

PR #274, opened 2018-10-04
Refactor Kotlin operators and add booleans

Pinpickle:
This PR adds operators to booleans in Kotlin.

Unfortunately, Kotlin doesn't allow you to overload the `&&` and `||` operators, so instead we have _infix functions_, e.g.

```
val vertex3 = vertex1 and vertex2
```

This PR also has a bunch of refactoring around how our Kotlin is setup. It's more generic, the `ArithmeticX` classes are documented, existing code has been trimmed down, tests have been split up and a bug or two has been fixed.

Tests have been left more or less the same, but they could also be improved (as they didn't catch a multiplication being implemented as a subtraction). This PR has already become bigger than expected so I'm opting to leave that for future work.


keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/DualNumber.java
```diff
@@ -431,4 +431,13 @@ private DoubleTensor takeFromPartial(DoubleTensor from, int... indices) {
             .reshape(takeShape);
     }
 
+    @Override
+    public DualNumber doubleMinusThis(double that) {
+        throw new UnsupportedOperationException();
```
Pinpickle: As dual numbers are being thrown out in another PR, I didn't implement this here

keanu-project/src/test/java/io/improbable/keanu/kotlin/IntegerOperatorOverloadingTest.kt
```diff
@@ -0,0 +1,240 @@
...
+
+import io.improbable.keanu.DeterministicRule
+import io.improbable.keanu.vertices.intgr.nonprobabilistic.ConstantIntegerVertex
+import io.improbable.keanu.vertices.intgr.probabilistic.PoissonVertex
+import io.improbable.keanu.vertices.intgr.probabilistic.UniformIntVertex
+import junit.framework.TestCase
+import org.junit.Rule
+import org.junit.Test
+
+class IntegerOperatorOverloadingTest {
```
migwellian: good call to pull this out into a separate file

keanu-project/src/test/java/io/improbable/keanu/tensor/TensorMatchers.java
```diff
@@ -90,6 +91,10 @@ public void describeTo(Description description) {
         };
     }
 
+    public static <T> Matcher<Tensor<T>> tensorEqualTo(Tensor<T> tensor) {
```
migwellian: I would favour overriding `Tensor#equals` rather than add this matcher. Otherwise the user has to remember the different semantics of `tensorEqualTo` vs `equalTo`.

Would like to put that idea to the rest of the team though, as changing the semantics of `Tensor#equals` might have implications.

keanu-project/src/test/java/io/improbable/keanu/vertices/VertexMatchers.java
```diff
@@ -60,6 +61,10 @@ public void describeTo(Description description) {
         return hasValue(Arrays.stream(values).map(v -> equalTo(v)).collect(Collectors.toList()));
     }
 
+    public static <DATATYPE, TENSOR extends Tensor<DATATYPE>> Matcher<Vertex<TENSOR>> hasTensorValue(TENSOR tensor) {
```
migwellian: Similar comment as for `TensorMatchers` - if we override `Tensor#equals` then your tests would work with `hasValue` and you wouldn't need `hasTensorValue`.

keanu-project/src/test/java/io/improbable/keanu/kotlin/BooleanOperatorOverloadingTest.kt
```diff
@@ -0,0 +1,70 @@
...
+        assertEquals(arithmeticFalse or arithmeticTrue, arithmeticTrue)
+        assertEquals(arithmeticFalse or false, arithmeticFalse)
+        assertEquals(false or arithmeticTrue, arithmeticTrue)
+    }
+
+    @Test
+    fun boolVertexPrefixOperators() {
+        val boolVertex = ConstantBoolVertex(true)
+
+        TestCase.assertEquals(false, not(boolVertex).value.scalar())
```
migwellian: you can do `!boolVertex` instead of `not(boolVertex)`

keanu-project/src/test/java/io/improbable/keanu/kotlin/BooleanOperatorOverloadingTest.kt
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Test
+    fun arithmeticBooleanPrefixOperators() {
+        val arithmeticBoolean = ArithmeticBoolean(true)
+
+        TestCase.assertEquals(false, not(arithmeticBoolean).value)
```
migwellian: you can do `!arithmeticBoolean` instead of `not(arithmeticBoolean)`

keanu-project/src/main/java/io/improbable/keanu/kotlin/BooleanOperators.kt
```diff
@@ -0,0 +1,26 @@
+package io.improbable.keanu.kotlin
+
+/**
+ * Custom operators for booleans.
+ *
+ * Built-in operators (&&, ||) cannot be overridden so infix functions are used instead.
```
migwellian: This is interesting, might be worth including a link in the javadoc: https://discuss.kotlinlang.org/t/overloading-and-and-or-operators/1623/2

> Overloading ‘and’ and ‘or’ is recommended against in most languages because short-circuiting is lost, which changes the usual symantics of the operators.

keanu-project/src/test/java/io/improbable/keanu/kotlin/BooleanOperatorOverloadingTest.kt
Pinpickle: For some reason I assumed this impossible, but I was conflating the absence of some boolean operators for absence of them all. Thanks for the reminder!

keanu-project/src/main/java/io/improbable/keanu/kotlin/BooleanOperators.kt
Pinpickle: This is a really interesting point about short-circuiting. I'll link to the discussion

keanu-project/src/test/java/io/improbable/keanu/tensor/TensorMatchers.java
Pinpickle: Is there a way to use `equalTo` and keep detailed error messages about diffs? The reason I used the tensor matchers here is because it will give you a detailed error message about how they differ (whether it's the shape or a value, in which index is the value different).

migwellian: If you use `equalTo` then it displays an error message that includes the object's `toString`. So I'd suggest that the better solution here is to write a better `Tensor#toString`.

migwellian: But that won't tell you in which index the value is different.

migwellian: I just noticed that I already wrote a `TensorMatchers#elementwiseEqualTo` but it's not being used anywhere. It checks the values but not the shape.

Pinpickle: You'll notice that this matcher uses `elementWiseEqualTo` under the hood

keanu-project/src/main/java/io/improbable/keanu/kotlin/ArithmeticDouble.kt
```diff
@@ -1,73 +1,54 @@
...
+ *    add5(ConstantDoubleVertex(0.0))
+ *    add5(ArithmeticDouble(5.0))
+ */
+data class ArithmeticDouble(val value: Double) : DoubleOperators<ArithmeticDouble> {
 
-class ArithmeticDouble(val value: Double) : DoubleOperators<ArithmeticDouble> {
-    override fun exp(): ArithmeticDouble {
-        return ArithmeticDouble(Math.exp(this.value))
-    }
+    override fun exp() = ArithmeticDouble(Math.exp(this.value))
```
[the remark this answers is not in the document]
gordoncaleb: 👍

keanu-project/src/main/java/io/improbable/keanu/kotlin/DoubleOperators.kt
```diff
@@ -1,17 +1,62 @@
...
 
-    operator fun minus(that: Double): T
     operator fun plus(that: Double): T
+
+    operator fun minus(that: Double): T
+
+    /**
+     * @return that - this
+     */
+    fun doubleMinusThis(that: Double): T
```
gordoncaleb: This is a very cool feature. Doesn't get used often but it's pretty key when it is. It's called reverse minus in nd4j and may have another name numpy. Can you find a reasonable name and use that?

Pinpickle: Yes! Will try and generalise it to division as well

Pinpickle: Looks like it's `rsub` in numpy too. I'll go with `reverseMinus` and `reverseDiv`

PR #275, opened 2018-10-05
Add argMax function to tensors

Pinpickle:
This PR adds a function to number tensors, `argMax`, that returns the indices of maximum values in a tensor. It's been adopted to follow the semantics of numpy's argmax method, which takes a single `axis` argument as opposed to an array of `dimensions`. It took me a while to grok how this method worked, so I tried to document it to help others a little more.

Shaping is a little strange with these semantics - the returned tensor can be rank 1 if the input was rank 2. Let me know if that's a problem and what we might want to do instead.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/tensor/NumberTensor.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/tensor/NumberTensor.java
@@ -39,9 +39,10 @@ public interface NumberTensor<N extends Number, T extends NumberTensor<N,T>> ext
      * // [[5]]
      * </pre>
      *
-     * @return A scalar tensor with the index of the largest value
+     * @return A scalar tensor with the index of the largest value. If there are multiple largest values, it will be
+     *         the first index.
      */
-    IntegerTensor argMax();
+    int argMax();
 
     /**
      * Find the indices into the tensor of the largest values in a specified axis (dimension), e.g.
@@ -57,7 +58,8 @@ public interface NumberTensor<N extends Number, T extends NumberTensor<N,T>> ext
      *
      * @param axis The axis (dimension) to find the largest values in
      * @return A tensor where each value is the location of the maximum value in the vector at that location in the
-     *         specified dimension in the original tensor
+     *         specified dimension in the original tensor. If there are multiple largest values in this vector, it will
+     *         be the first index.
      * @see <a href="https://www.geeksforgeeks.org/numpy-argmax-python/">An article about argmax over an axis in numpy</a>
      */
     IntegerTensor argMax(int axis);
--- a/keanu-project/src/main/java/io/improbable/keanu/tensor/NumberTensor.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/tensor/NumberTensor.java
@@ -29,6 +29,39 @@ public interface NumberTensor<N extends Number, T extends NumberTensor<N,T>> ext
     
     T abs();
 
+    /**
+     * Find the index into the flattened array of the tensor of the largest value, e.g.
+     * <pre>
+     * DoubleTensor tensor = DoubleTensor.arange(0, 6).reshape(2, 3);
+     * // [[0., 1., 2.],
+     * //  [3., 4., 5.]]
+     * IntegerTensor max = tensor.argMax();
+     * // [[5]]
+     * </pre>
+     *
+     * @return A scalar tensor with the index of the largest value
+     */
+    IntegerTensor argMax();
+
+    /**
+     * Find the indices into the tensor of the largest values in a specified axis (dimension), e.g.
+     * <pre>
+     * DoubleTensor tensor = DoubleTensor.arange(0, 6).reshape(2, 3);
+     * // [[0., 1., 2.],
+     * //  [3., 4., 5.]]
+     * IntegerTensor maxesFor0 = tensor.argMax(0);
+     * // [[1, 1, 1]]
+     * IntegerTensor maxFor1 = tensor.argMax(1);
+     * // [[2, 2]]
+     * </pre>
+     *
+     * @param axis The axis (dimension) to find the largest values in
+     * @return A tensor where each value is the location of the maximum value in the vector at that location in the
+     *         specified dimension in the original tensor
+     * @see <a href="https://www.geeksforgeeks.org/numpy-argmax-python/">An article about argmax over an axis in numpy</a>
+     */
+    IntegerTensor argMax(int axis);
+
     T getGreaterThanMask(T greaterThanThis);
 
     T getGreaterThanOrEqualToMask(T greaterThanThis);
--- a/keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
@@ -1,10 +1,8 @@
 package io.improbable.keanu.tensor;
 
-import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.List;
 
-import com.google.common.primitives.Ints;
+import org.apache.commons.lang3.ArrayUtils;
 
 public class TensorShape {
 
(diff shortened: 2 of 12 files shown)


keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
```diff
@@ -215,5 +217,16 @@ public static boolean isScalar(int[] shape) {
...
     }
 
+    /**
+     * Removes a dimension from a shape
+     *
+     * @param shape the shape to remove the dimension from
+     * @param dimension the dimension to remove
+     * @return the shape without the given dimension
+     */
+    public static int[] argMaxShape(int[] shape, int dimension) {
```
GeorgeNash: Let's call this `removeDimensionFromShape`?

keanu-project/src/test/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensorTest.java
```diff
@@ -935,6 +936,48 @@ public void canFindElementWiseMinAndMax() {
         assertArrayEquals(new double[]{2, 3, 3, 4}, max.asFlatDoubleArray(), 1e-6);
     }
 
+    @Test
+    public void canArgFindMaxOfVector() {
```
GeorgeNash: nit: `canFindArgMaxOfVector`

keanu-project/src/main/java/io/improbable/keanu/tensor/NumberTensor.java
```diff
@@ -29,6 +29,39 @@
...
+     * DoubleTensor tensor = DoubleTensor.arange(0, 6).reshape(2, 3);
+     * // [[0., 1., 2.],
+     * //  [3., 4., 5.]]
+     * IntegerTensor max = tensor.argMax();
+     * // [[5]]
+     * </pre>
+     *
+     * @return A scalar tensor with the index of the largest value
+     */
+    IntegerTensor argMax();
```
GeorgeNash: Why doesn't this just return an `int`?

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/ScalarDoubleTensor.java
```diff
@@ -358,6 +359,17 @@ public double min() {
...
 
+    @Override
+    public IntegerTensor argMax() {
+        return IntegerTensor.scalar(0);
+    }
+
+    @Override
+    public IntegerTensor argMax(int axis) {
+        TensorShapeValidation.checkDimensionExistsInShape(axis, this.getShape());
+        return IntegerTensor.create(0).reshape(1);
```
GeorgeNash: This looks weird. ND4J doesn't support rank 1's (vectors). Why doesn't this just return an `int` anyway?

keanu-project/src/test/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensorTest.java
```diff
@@ -935,6 +936,48 @@ public void canFindElementWiseMinAndMax() {
...
+        DoubleTensor tensor = DoubleTensor.create(1, 3, 4, 5, 2);
+
+        assertThat(tensor.argMax(), isScalarWithValue(3));
+        assertThat(tensor.argMax(0), tensorEqualTo(IntegerTensor.zeros(5).reshape(5)));
```
GeorgeNash: We shouldn't have a reshape of just `5`, that's not supported by ND4J. It'll need to be a rank 2 vector, `[1,5]` or `[5,1]`

keanu-project/src/main/java/io/improbable/keanu/tensor/NumberTensor.java
```diff
@@ -29,6 +29,39 @@
...
+     * IntegerTensor maxFor1 = tensor.argMax(1);
+     * // [[2, 2]]
+     * </pre>
+     *
+     * @param axis The axis (dimension) to find the largest values in
+     * @return A tensor where each value is the location of the maximum value in the vector at that location in the
+     *         specified dimension in the original tensor
+     * @see <a href="https://www.geeksforgeeks.org/numpy-argmax-python/">An article about argmax over an axis in numpy</a>
+     */
+    IntegerTensor argMax(int axis);
```
GeorgeNash: Is this returning a rank 1 vector? I think you'll need to change this to a rank 2

keanu-project/src/main/java/io/improbable/keanu/tensor/NumberTensor.java
GeorgeNash: What happens if there's two values that are the max? are we meant to return both indices?

eg:
```
DoubleTensor tensor = DoubleTensor.create(1, 2, 4, 3, 4, 1, 3, 1).reshape(2, 4);
tensor.argMax();
```

Which index of `4` do we return?

keanu-project/src/main/java/io/improbable/keanu/tensor/NumberTensor.java
christophernorth: Yep - the code as it stands will get confused by anything rank < 2

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
gordoncaleb: This is more of an argMax concern than a TensorShape concern. I'd either move this to the class it's needed in or make a more general purpose method like @GeorgeNash is suggesting. However, since it's really just a wrapper on `ArrayUtils.remove(shape, dimension)`, which is a single line, I'd say it might not be worth the coupling/abstraction it causes.

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShapeValidation.java
```diff
@@ -45,6 +45,19 @@ public static void checkTensorsMatchNonScalarShapeOrAreScalar(int[] proposalShap
...
 
+    /**
+     * Check if the given dimension exists within the shape
+     *
+     * @param dimension Proposed dimension
+     * @param shape Shape to check
+     * @throws IllegalArgumentException if the dimension exceeds the rank of the shape
+     */
+    public static void checkDimensionExistsInShape(int dimension, int[] shape) {
+        if (dimension > shape.length - 1) {
```
gordoncaleb: `dimension >= shape.length` is a bit more simple.

keanu-project/src/main/java/io/improbable/keanu/tensor/NumberTensor.java
Pinpickle: This will return the first instance of 4, (so `2`). There are tests for this but I'll also document this.

Good call on returning an `int`, I'll get that in

keanu-project/src/main/java/io/improbable/keanu/tensor/NumberTensor.java
Pinpickle: Wasn't sure exactly what restrictions were here, I'll fix that up!

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
Pinpickle: Turns out this isn't even something we need to do if we throw out the weird rank 1 behaviour, so I'm throwing this function out.

Pinpickle: I stand corrected, it is something we need to keep! The logic is slightly more complex so I'll keep it here in the more generic form

Pinpickle: Actually I do need to implement this, I'll go with the more general version as the logic is more complicated now.

Pinpickle:
@GeorgeNash I've fixed up the strange rank situation and addressed your other comments. Could you please have another look?

PR #271, opened 2018-10-03
[#188] Change Categorical to take GenericTensor as value

icantsayrural:

PR #276, opened 2018-10-05
Make all tensors implement `setWithMask` and `setWithMaskInPlace`

icantsayrural:
Required for #271

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/tensor/NumberTensor.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/tensor/NumberTensor.java
@@ -29,17 +29,13 @@ public interface NumberTensor<N extends Number, T extends NumberTensor<N,T>> ext
     
     T abs();
 
-    T getGreaterThanMask(T greaterThanThis);
+    DoubleTensor getGreaterThanMask(T greaterThanThis);
 
-    T getGreaterThanOrEqualToMask(T greaterThanThis);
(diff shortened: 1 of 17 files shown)


keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -1029,7 +1029,7 @@ public BooleanTensor elementwiseEquals(Tensor that) {
         }
     }
 
-    static INDArray unsafeGetNd4J(DoubleTensor that) {
+    static public INDArray unsafeGetNd4J(DoubleTensor that) {
```
gordoncaleb: Does this need to be public? This should be implementation specific and therefore private to this class.

keanu-project/src/main/java/io/improbable/keanu/tensor/NumberTensor.java
```diff
@@ -29,17 +29,13 @@
     
     T abs();
 
-    T getGreaterThanMask(T greaterThanThis);
+    DoubleTensor getGreaterThanMask(T greaterThanThis);
```
gordoncaleb: I think IntegerTensor as a mask makes more sense. Why double?

icantsayrural:
Closing, because we are no longer making all tensors implement `setWithMask*`. This is because we'd want the mask to be the same type as the tensor for other operations.

PR #277, opened 2018-10-08
Feature/broadcasting bugs

GeorgeNash:
Fixes bugs in broadcast.

Moves the `broadcast` operations into their own testing class.

Makes it explicit in the tests which types of broadcast we can't do by annotating the test with: `@Test(expected = AssertionError.class)`


keanu-project/src/test/java/io/improbable/keanu/tensor/dbl/Nd4jBroadcastTensorTest.java
```diff
@@ -0,0 +1,513 @@
...
+        DoubleTensor expected = Nd4jDoubleTensor.create(new double[]{
+            -4, -1, -2, -6, -7,
+            -3, 0, -1, -5, -6,
+            -2, 1, 0, -4, -5
+        }, new int[]{3, 5});
+
+        assertEquals(expected, diff);
+    }
+
+    @Test(expected = AssertionError.class)
```
gordoncaleb: Should this be `@Ignore`? http://junit.sourceforge.net/javadoc/org/junit/Ignore.html

PR #278, opened 2018-10-09
check tensor rank >= 2 on construction

migwellian:
Added a new method in `TensorShapeValidation` and call it from each Tensor ctor so we catch these issues earlier.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShapeValidation.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShapeValidation.java
@@ -8,6 +8,8 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Stream;
 
+import com.google.common.base.Preconditions;
+
 public class TensorShapeValidation {
 
     private TensorShapeValidation() {
@@ -147,5 +149,8 @@ public class TensorShapeValidation {
         }
     }
 
+    public static void checkRankIsAtLeastTwo(int[] shape) {
+        Preconditions.checkArgument(shape.length > 1, "Tensors must have rank >=2 : " + Arrays.toString(shape));
 
+    }
 }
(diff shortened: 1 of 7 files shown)


keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShapeValidation.java
```diff
@@ -147,5 +149,8 @@ public static void checkIndexIsValid(int[] shape, int... index) {
         }
     }
 
+    public static void checkRankIsAtLeastTwo(int[] shape) {
+        Preconditions.checkArgument(shape.length > 1, "Tensors must have rank >=2 : " + Arrays.toString(shape));
```
Pinpickle: I didn't know we could do this, nice!

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -84,10 +85,11 @@ public static Nd4jDoubleTensor arange(double start, double end, double stepSize)
     private INDArray tensor;
 
     public Nd4jDoubleTensor(double[] data, int[] shape) {
-        this.tensor = TypedINDArrayFactory.create(data, shape, BUFFER_TYPE);
+        this(TypedINDArrayFactory.create(data, shape, BUFFER_TYPE));
     }
 
     public Nd4jDoubleTensor(INDArray tensor) {
+        TensorShapeValidation.checkRankIsAtLeastTwo(tensor.shape());
```
Pinpickle: It's great that we're doing this now, takes the human out of making sure we don't do this (and end-users are informed that things will break). Should we have tests to validate that this is working as expected?

migwellian: good call - added

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -84,10 +85,11 @@ public static Nd4jDoubleTensor arange(double start, double end, double stepSize)
     private INDArray tensor;
 
     public Nd4jDoubleTensor(double[] data, int[] shape) {
-        this.tensor = TypedINDArrayFactory.create(data, shape, BUFFER_TYPE);
+        this(TypedINDArrayFactory.create(data, shape, BUFFER_TYPE));
```
gordoncaleb: Does ND4J throw an exception if the shape is zero length? If it doesn't throw an exception and returns a 1x1 then we might want to do the check before the ND4J INDArray creation.

migwellian: ND4J doesn't throw, it returns an INDArray whose shape is an empty `int[]`.

PR #279, opened 2018-10-10
Fix bug in concat reverse mode autodiff

gordoncaleb:
Previously we were splitting on the accidentally correct dimension in our tests. For higher rank tensors this fails.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/ConcatenationVertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/ConcatenationVertex.java
@@ -139,7 +139,7 @@ public class ConcatenationVertex extends DoubleVertex implements Differentiable,
             splitPartials.put(operands[i], new PartialDerivatives(new HashMap<>()));
         }
 
-        int wrtDimensionToSliceOn = operands[0].getShape().length + dimension;
+        int wrtDimensionToSliceOn = -operands[0].getShape().length + dimension;
         for (Map.Entry<VertexId, DoubleTensor> entry : derivativeOfOutputsWithRespectToSelf.asMap().entrySet()) {
             DoubleTensor partial = entry.getValue();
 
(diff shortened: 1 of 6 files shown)


keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/ConcatenationVertex.java
```diff
@@ -139,7 +139,7 @@ private static DoubleTensor concatPartialDerivatives(int dimension, List<DoubleT
             splitPartials.put(operands[i], new PartialDerivatives(new HashMap<>()));
         }
 
-        int wrtDimensionToSliceOn = operands[0].getShape().length + dimension;
+        int wrtDimensionToSliceOn = -operands[0].getShape().length + dimension;
```
christophernorth: It's not immediately obvious how this works - comment or one extra step?

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/binary/ConcatenationVertexTest.java
```diff
@@ -169,6 +173,23 @@ public void canConcatenateSimpleAutoDiff() {
...
+        b.setValue(DoubleTensor.arange(0, 8).reshape(2, 2, 2));
+
+        DoubleVertex c = a.times(ConstantVertex.of(DoubleTensor.linspace(0, 1, 12).reshape(2, 2, 3)));
+        DoubleVertex d = b.plus(ConstantVertex.of(DoubleTensor.linspace(1, 2, 8).reshape(2, 2, 2)));
+
+        DoubleVertex concat = new ConcatenationVertex(2, c, d);
+        DoubleVertex sum = concat.sum(1);
+
+        finiteDifferenceMatchesForwardAndReverseModeGradient(Arrays.asList(a,b), sum, 10.0, 1e-10);
+    }
```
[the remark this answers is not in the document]
gordoncaleb: We don't support negative dimension indexing in the concat or sum at the moment. I've documented this for both sum and concat. Maybe we should support this?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/multiple/ConcatenationVertex.java
gordoncaleb: 👍

PR #280, opened 2018-10-10
ND4J Upgrade

GeorgeNash:
Shapes are now `longs`
Indices are now `longs`
dimensions are still `int`


keanu-project/src/main/java/io/improbable/keanu/tensor/generic/GenericTensor.java
```diff
@@ -63,13 +65,13 @@ public boolean isShapePlaceholder() {
     }
 
     @Override
-    public T getValue(int... index) {
-        return data[getFlatIndex(shape, stride, index)];
+    public T getValue(long... index) {
+        return data[Ints.saturatedCast(getFlatIndex(shape, stride, index))];
```
gordoncaleb: We should do a check here and throw if it's above or below max or min instead of saturated casting.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/MultivariateGaussianVertex.java
```diff
@@ -53,7 +53,7 @@ public MultivariateGaussianVertex(DoubleVertex mu, DoubleVertex covariance) {
      * @param covariance the scale of the identity matrix
      */
     public MultivariateGaussianVertex(DoubleVertex mu, double covariance) {
-        this(mu, ConstantVertex.of(DoubleTensor.eye(mu.getShape()[0])).times(covariance));
+        this(mu, ConstantVertex.of(DoubleTensor.eye((int)mu.getShape()[0])).times(covariance));
```
gordoncaleb: `Nd4j.eye(...)` now takes a long. Can we have our `.eye` take a long as well?

keanu-project/src/main/java/io/improbable/keanu/model/linear/LinearRegression.java
```diff
@@ -60,7 +62,7 @@ private BayesianNetwork buildModel(DoubleTensor x, double priorOnMu, double prio
     }
 
     private static double[] fillPriorOnWeights(DoubleTensor x, double priorOnSigma) {
-        double[] priorWeights = new double[x.getShape()[0]];
+        double[] priorWeights = new double[Ints.saturatedCast(x.getShape()[0])];
```
christophernorth: I think there's a better cast here: checkedCast()

keanu-project/src/main/java/io/improbable/keanu/model/linear/LogisticRegression.java
```diff
@@ -60,7 +62,7 @@ private BayesianNetwork buildModel(DoubleTensor x, double priorMu, double priorS
     }
 
     private double[] regularizeSigma(DoubleTensor x, double priorSigma, double regularization) {
-        int numFeatures = x.getShape()[0];
+        int numFeatures = Ints.saturatedCast(x.getShape()[0]);
```
christophernorth: as above

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Wed Jun 06 15:50:00 BST 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.1-bin.zip
```
migwellian: what's this change for?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/MultivariateGaussianVertex.java
christophernorth: should the function just take a 'long'

keanu-project/src/main/java/io/improbable/keanu/model/linear/LinearRegression.java
migwellian: I think `Math.toIntExact` would be better, since you're using it to construct an array.

    public static int toIntExact(long value)
    Returns the value of the long argument; throwing an exception if the value overflows an int.

https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#toIntExact-long-

    public static int saturatedCast(long value)
    Returns the same value cast to int if it is in the range of the int type, Integer.MAX_VALUE if it is too large, or Integer.MIN_VALUE if it is too small

https://google.github.io/guava/releases/19.0/api/docs/com/google/common/primitives/Ints.html

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -84,8 +86,9 @@ public static Nd4jDoubleTensor arange(double start, double end, double stepSize)
 
     private INDArray tensor;
 
-    public Nd4jDoubleTensor(double[] data, int[] shape) {
-        this(TypedINDArrayFactory.create(data, shape, BUFFER_TYPE));
+    public Nd4jDoubleTensor(double[] data, long[] shape) {
+        TensorShapeValidation.checkRankIsAtLeastTwo(shape);
+        this.tensor = TypedINDArrayFactory.create(data, shape, BUFFER_TYPE);
```
migwellian: This looks like it needs merging from develop

keanu-project/src/main/java/io/improbable/keanu/tensor/generic/GenericTensor.java
migwellian: see my comment above - use `Math.toIntExact`

gradle/wrapper/gradle-wrapper.properties
christophernorth: I've always thought it was weird we used 'all' rather than just 'bin' so this looks like a good change to me!

keanu-project/src/main/java/io/improbable/keanu/model/linear/LinearRegression.java
GeorgeNash: Agreed - I've ended up going for `checkedCast` which is the same behaviour as `Math.toIntExact`

gradle/wrapper/gradle-wrapper.properties
GeorgeNash: I don't know, gradle must've made this change for me...I'll change it back seeing as I didn't intend to do it

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
GeorgeNash: Good spot - bad merge on my part

keanu-project/src/main/java/io/improbable/keanu/model/linear/LinearRegression.java
migwellian: The thing is, we are already using `toIntExact` elsewhere in the code base - I think it's important to maintain consistency. I don't have a strong opinion on which of the two we use.

gradle/wrapper/gradle-wrapper.properties
migwellian: In which case you'll need to revert `gradlew.bat` as well

PR #281, opened 2018-10-10
Beta log prob as graph

gordoncaleb:
This PR introduces a fundamental change to the way calculate log probabilities in Keanu. At the moment the log probability calculation is done as effectively a single operation in the graph. This change makes the log prob calculation a graph of simple operations, which can be used to achieve the same thing as before but now we can convert the calculation to other computational graphs (e.g. tensorflow) as well as use auto diff on the graph and remove the need for dLogProb (maybe?). I've only converted Gaussian as a start but the others will follow the same pattern. There will be logic duplication until we are ready to actually use the log prob as a graph, at which point we can remove the old log prob as an operation code.


springcoil:
It might be worth @gordoncaleb having a look at the implementation notes here https://github.com/stripe/rainier/blob/develop/docs/impl.md since that's that a JVM based project too. 

I'm not an expert neccesarily in the intricacies of autodiff, but I am of the opinion that something like Tensorflow or Theano could be a good idea, since they have been heavily optimized for doing autograd.

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/probabilistic/GaussianVertexTest.java
```diff
@@ -50,6 +53,25 @@ public void matchesKnownLogDensityOfVector() {
...
+
+        NormalDistribution distribution = new NormalDistribution(0.0, 1.0);
+        double expectedLogDensity = distribution.logDensity(0.25) + distribution.logDensity(-0.75);
+
+        DoubleVertex mu = ConstantVertex.of(0.0);
+        DoubleVertex sigma = ConstantVertex.of(1.0);
+        GaussianVertex tensorGaussianVertex = new GaussianVertex(mu, sigma);
+
+        LogProbGraph logProbGraph = tensorGaussianVertex.logProbGraph();
+        logProbGraph.getInput(mu).setValue(DoubleTensor.scalar(0.0));
```
icantsayrural: Should we have a `setInput(input, value)` for `LogProbGraph`?

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/LogNormal.java
```diff
@@ -8,50 +8,62 @@
 import io.improbable.keanu.distributions.ContinuousDistribution;
 import io.improbable.keanu.distributions.hyperparam.Diffs;
 import io.improbable.keanu.tensor.dbl.DoubleTensor;
+import io.improbable.keanu.vertices.dbl.DoubleVertex;
 import io.improbable.keanu.vertices.dbl.KeanuRandom;
 
 public class LogNormal implements ContinuousDistribution {
```
joncoello83: The ContinuousDistribution interface has been removed from the Gaussian distribution. Should it be remove here as well?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/PlaceHolderDoubleVertex.java
```diff
@@ -0,0 +1,19 @@
+package io.improbable.keanu.vertices.dbl.nonprobabilistic;
+
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+import io.improbable.keanu.vertices.NonProbabilistic;
+import io.improbable.keanu.vertices.dbl.DoubleVertex;
+import io.improbable.keanu.vertices.dbl.KeanuRandom;
+
+public class PlaceHolderDoubleVertex extends DoubleVertex implements NonProbabilistic<DoubleTensor> {
```
joncoello83: Interesting idea. Two questions. 1) Could we build the LogProbGraph using the hyperparameters rather than placeholders? This might offer some advantages for autodiffing the rate of change of logProb with respect to the value of upstream latent vertices. Could even have the sum operation in the graph. 2) Is there a technical reason why ConstantDoubleVertex couldn't be used as a placeholder? Appreciate that it wouldn't be as nice to instantiate with a value that would always be overwritten.

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/Beta.java
```diff
@@ -7,66 +7,99 @@
...
 
-    private final DoubleTensor alpha;
-    private final DoubleTensor beta;
-    private final DoubleTensor xMin;
-    private final DoubleTensor xMax;
+    private final DoubleVertex x;
+    private final DoubleVertex alpha;
+    private final DoubleVertex beta;
+    private final DoubleVertex xMin;
+    private final DoubleVertex xMax;
```
migwellian: I feel uneasy about introducing `Vertex` classes into the `Distribution` classes. The way it is now feels neat, that `Distributions` deal with `Tensor`s and need know nothing about `Vertex`es.

migwellian: 2 possible ways to change it:
1. Make `logProbGraph` a method on `BetaVertex` rather than on `Beta`
2. Build your log prob graph out of something other than `Vertex`es (as per my comment later on in `LogProbGraph.java`)

keanu-project/src/main/java/io/improbable/keanu/vertices/LogProbGraph.java
```diff
@@ -0,0 +1,35 @@
...
+
+    private final Map<Vertex<?>, Vertex<?>> inputs;
+    private final DoubleVertex logProbOutput;
+
+    public LogProbGraph(DoubleVertex logProbOutput) {
+        this.inputs = new HashMap<>();
+        this.logProbOutput = logProbOutput;
+    }
+
+    public <T> LogProbGraph addInput(Vertex<T> from, Vertex<T> to) {
```
migwellian: It would be nice if `LogProbGraph` was immutable. You could create a builder class with an `addInput` method.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/PlaceHolderDoubleVertex.java
migwellian: Can't you just use a `ConstantDoubleVertex` and set its value?

migwellian: +1 to "Is there a technical reason why ConstantDoubleVertex couldn't be used as a placeholder"

keanu-project/src/main/java/io/improbable/keanu/vertices/LogProbGraph.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+import io.improbable.keanu.vertices.dbl.DoubleVertex;
+
+public class LogProbGraph {
+
+    private final Map<Vertex<?>, Vertex<?>> inputs;
```
migwellian: If I understand this correctly, there's no need to use `Vertex` as the map key. It would be enough to key on `VertexId` or even `VertexLabel`.

keanu-project/src/main/java/io/improbable/keanu/vertices/LogProbGraph.java
```diff
@@ -0,0 +1,35 @@
...
+        this.inputs = new HashMap<>();
+        this.logProbOutput = logProbOutput;
+    }
+
+    public <T> LogProbGraph addInput(Vertex<T> from, Vertex<T> to) {
+        inputs.put(from, to);
+        return this;
+    }
+
+    public Map<Vertex<?>, Vertex<?>> getInputs() {
```
migwellian: This method is dangerous as the map could be changed by the calling code. It's also unused at the moment - can we remove it?

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/Beta.java
```diff
@@ -7,66 +7,99 @@
...
 
-    public static ContinuousDistribution withParameters(DoubleTensor alpha, DoubleTensor beta, DoubleTensor xMin, DoubleTensor xMax) {
-        return new Beta(alpha, beta, xMin, xMax);
+    private final LogProbGraph logProbGraph;
+
+    public static Beta withParameters(DoubleVertex x, DoubleVertex alpha, DoubleVertex beta, DoubleVertex xMin, DoubleVertex xMax) {
+        return new Beta(x, alpha, beta, xMin, xMax);
```
migwellian: The value of `x` is never used. Its only purpose is to act as a key in your `LogProbGraph` map. I think therefore it doesn't need to be a constructor argument - we can keep the constructor signature unchanged.

keanu-project/src/main/java/io/improbable/keanu/vertices/LogProbGraph.java
gordoncaleb: Not all vertices have labels and if you key off of VertexId you will just end up calling `getInput(vertex.getId())` instead of `getInput(vertex)`.

keanu-project/src/main/java/io/improbable/keanu/vertices/LogProbGraph.java
```diff
@@ -0,0 +1,34 @@
+package io.improbable.keanu.vertices;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import io.improbable.keanu.vertices.dbl.DoubleVertex;
+
+public class LogProbGraph {
+
+    private final Map<Vertex<?>, Vertex<?>> inputs;
```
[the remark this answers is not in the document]
gordoncaleb: A graph without random variables or probabilistic would still have `vertices` right? A Bayesian Network is just a special case of a graph and a graph is composed of vertices. I think we need to rename some of our interfaces so that we either have `RandomVariables<T>` or `Operation<T>` which will both be a `Vertex<T>`. In that case this mapping on inputs could still be either from RandomVariable to Operation or from Operation to Operation.

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/Beta.java
gordoncaleb: DoubleTensors is a computation concern where as Vertex is a "problem description" (maybe need a better name for this) concern. The question is where does the Beta distribution live? The BetaVertex is clearly on the "problem description" side and it will depend on some distribution functionality, which will now require a graph. If the distribution supplies a graph then I think it stands to reason that it should also live on the description side. This might mean that we can just move the graph based code from the distribution into the random variable vertex and then drop the distribution classes. So where does the DoubleTensor computation functionality now live? Well, if our inference/computation requires a compiled graph then we would only ever need the DoubleTensor functions in our compiled graph, which may be a compiled Tensorflow graph or something of our own until we can get full functionality out of Tensorflow. If its TF then we don't need the DoubleTensor code at all and if its something of our own then I'd say we'd want to generate the DoubleTensor code based on the same graph that we would use for the TF conversion.

keanu-project/src/main/java/io/improbable/keanu/vertices/LogProbGraph.java
gordoncaleb: 👍

keanu-project/src/main/java/io/improbable/keanu/vertices/LogProbGraph.java
gordoncaleb: If the inputs map is an ImmutableMap then it's not a problem right? I use it in the TF conversion code, which isn't on this branch so we can drop it until that gets merged in.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/PlaceHolderDoubleVertex.java
gordoncaleb: No technical reason at the moment but I'd say a ConstantVertex shouldn't be changed by an algorithm where as a PlaceHolder requires the algorithm to give it a value. Should we disallow a ConstantVertex value change after construction?

gordoncaleb: Could but that's semantically wrong I think. It would save use creating new classes, but it'd be confusing.

keanu-project/src/main/java/io/improbable/keanu/vertices/LogProbGraph.java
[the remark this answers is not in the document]
migwellian: Agreed - let's attack the problem in terms of interfaces.
* What interface is only for BayesNet Vertices? `sample`, `observe` etc.
* What interface is common to both types? `eval`, `addChild` etc.

keanu-project/src/main/java/io/improbable/keanu/vertices/LogProbGraph.java
migwellian: Yes but what I mean is, you could just as well use `String` for your key as `Vertex`.

keanu-project/src/main/java/io/improbable/keanu/vertices/LogProbGraph.java
migwellian: Right.

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/Beta.java
migwellian: Yes, I would say that the distribution classes are a computation concern.

Our discussion below offers a 3rd option, which is that we redefine `Vertex` so that it's something more general with a smaller API. Then I would be comfortable mixing `Vertex` with `Distribution`.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/PlaceHolderDoubleVertex.java
migwellian: Do we even require our Constant Vertices to be constant (i.e. immutable)? Suppose we renamed them to something like `SimpleVertex`, `ValueVertex` or `VariableVertex` - would you then be happy to use them here too?

gordoncaleb: A variable and constant are semantically different so I think you need to be able to express both of them separately. We could call these place holder vertices variable vertices but those two concepts will also need to be separate when we start adding in control flow where place holders would be function arguments and variables would be stateful.

keanu-project/src/main/java/io/improbable/keanu/vertices/LogProbGraph.java
gordoncaleb: Then we have to have a consistent vertex to string mapping, which can be done (and I have on the tensorflow branch) but it's more complicated than we need at the moment.

keanu-project/src/main/java/io/improbable/keanu/distributions/continuous/Beta.java
gordoncaleb: `Distribution` as a DoubleTensor only class will go away when everything is a graph. Until everything is a graph, we need it to still provide the old `DoubleTensor` API. When everything is a graph, the issue won't be wether `Distribution` should be coupled to `Vertex` because there won't be any non-`Vertex` functionality left for the `Distribution`.

keanu-project/src/main/java/io/improbable/keanu/vertices/LogProbGraph.java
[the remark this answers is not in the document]
gordoncaleb: I was suggesting that we refactor the way we've currently set interfaces on our Vertex classes. There should be two types of vertices, RandomVariables and Operation with the only difference being the way we allow the user to observe them. Currently we've defined them as Probabilistic and NonProbabilistic but I think that's wrong. For the purposes of this mapping, in both cases it's a mapping of vertex to vertex. The tightest you could bound it would be Map<Vertex<?>, Operation<?>> but I don't think the tighter bounds gives you anything except added complexity.

PR #282, opened 2018-10-11
Add sampling many scalar tensors into one tensor

### LINKED ISSUE #181: Create easy way to take x samples

Pinpickle:
Fixes #181 

This supersedes #261 

This adds a method to take multiple scalar samples from a tensor vertex, returning a tensor matching the given shape. It only works for vertices with a scalar shape.


keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShapeValidation.java
```diff
@@ -60,6 +60,18 @@ public static void checkDimensionExistsInShape(int dimension, long[] shape) {
         }
     }
 
+    public static void checkTensorsAreScalar(long[] ...shapes) {
```
migwellian: whitespace - should be `long[]... shapes`. If you format the file in IntelliJ it will make this change.

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShapeValidation.java
```diff
@@ -60,6 +60,18 @@ public static void checkDimensionExistsInShape(int dimension, long[] shape) {
...
+        checkTensorsAreScalar("Expected shapes to be scalar");
+    }
+
+    public static void checkTensorsAreScalar(String message, long[] ...shapes) {
```
migwellian: same comment about whitespace

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/probabilistic/ProbabilisticBooleanVertex.java
```diff
@@ -0,0 +1,24 @@
+package io.improbable.keanu.vertices.bool.probabilistic;
+
+import io.improbable.keanu.tensor.TensorShapeValidation;
+import io.improbable.keanu.tensor.bool.BooleanTensor;
+import io.improbable.keanu.vertices.bool.BoolVertex;
+import io.improbable.keanu.vertices.dbl.KeanuRandom;
+
+public abstract class ProbabilisticBooleanVertex extends BoolVertex implements ProbabilisticBoolean {
```
migwellian: I am not keen on introducing more abstract classes into the hierarchy. Since `sample` is a member of `Vertex`, that seems like the natural place for these new methods too.

keanu-project/src/test/java/io/improbable/keanu/vertices/bool/probabilistic/ProbabilisticBooleanVertexTest.java
```diff
@@ -0,0 +1,72 @@
...
+        vertex.sampleManyScalars(new long[]{1, 1});
+    }
+
+    private static class ProbabilisticBooleanVertexImplementation extends ProbabilisticBooleanVertex {
+        ProbabilisticBooleanVertexImplementation(long... shape) {
+            this.setValue(BooleanTensor.placeHolder(shape));
+        }
+
+        @Override
+        protected BooleanTensor sampleWithShape(long[] shape, KeanuRandom random) {
```
migwellian: It doesn't really make sense to implement the concrete class `ProbabilisticBooleanVertexImplementation` - your tests are then dependent on its logic, rather than focussing on the logic of `ProbabilisticBooleanVertex`.

What you could do instead is instantiate a `mock(ProbabilisticBooleanVertex.class)` and then verify that its `sampleWithShape` method is called with the appropriate `shape` argument.

migwellian: Ditto for the Integer and Double versions.

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShapeValidation.java
Pinpickle: This makes much more sense, `...` is really a type modifier

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/KDEVertex.java
```diff
@@ -103,7 +104,7 @@ public DoubleTensor sample(int nSamples, KeanuRandom random) {
     }
 
     @Override
-    public DoubleTensor sample(KeanuRandom random) {
+    public DoubleTensor sampleWithShape(long[] shape, KeanuRandom random) {
         return sample(1, random);
```
migwellian: It seems wrong that this one doesn't use `shape`. Any thoughts on this @GeorgeNash / @gordoncaleb ?

Pinpickle: I believe this was a mistake with my find + replace

migwellian: The `GaussianKDE` algorithm asserts that the shape of the samples must be scalar. The `KDEVertex` provides a method `sample(int, KeanuRandom)` which returns a tensor of shape [1, n]. I therefore think that `KDEVertex` shouldn't implement `SamplableWithManyScalars`. I've changed it to implement `Samplable` instead.

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShapeValidation.java
migwellian: fixed

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShapeValidation.java
migwellian: fixed

PR #283, opened 2018-10-15
Feature/python ci

migwellian:
Build the shadow jar and run the python tests on Travis and Appveyor


keanu-python/tests/test_appveyor_hanging.py
```diff
@@ -0,0 +1,25 @@
+import keanu as kn
+import numpy as np
+import sys
+import logging
+
+
+def test_appveyor_hanging():
```
GeorgeNash: Do we need this now that AppVeyor is no longer hanging - or do you want to keep it just incase it starts happening again?

keanu-python/tests/test_appveyor_hanging.py
```diff
@@ -0,0 +1,25 @@
...
+    print("test_appveyor_hanging")
+    logging.warn("test_appveyor_hanging")
```
GeorgeNash: If we are keeping this - can you swap this for `logging.warning` - it's chucking out some annoying deprecated messages when you build

.travis.yml
```diff
@@ -1,9 +1,17 @@
...
-  - 'if [[ ! -z $SONAR_TOKEN ]]; then sonar-scanner; fi'
+matrix:
+  include:
+    - language: java
+      jdk: oraclejdk8
+      addons:
+        sonarcloud:
+          organization: "improbable-keanu"
+      script:
+        - ./gradlew clean build -x :keanu-python:build -i
```
GeorgeNash: Why do we need to run `keanu-python:build` here - aren't we running this again in the python section of the build?

keanu-python/tests/test_appveyor_hanging.py
migwellian: Good call. I will remove it and we can always refer back to this PR if we need to reinstate it.

.travis.yml
migwellian: The `-x` means exclude

PR #285, opened 2018-10-16
Add a GPB object representation for a graph

christophernorth:
First stab at a GPB representation of a graph.  No code here to actually generate them or load them, but wanted to start a discussion on representation thats not blocked on the rest of the code.

### DIFF
--- a/keanu-project/build.gradle
+++ b/keanu-project/build.gradle
@@ -6,6 +6,7 @@ buildscript {
 
     dependencies {
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
+        classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.6'
     }
 }
 
@@ -15,6 +16,11 @@ plugins {
     id 'jacoco'
 }
 
+apply plugin: 'java'
+apply plugin: 'jacoco'
+apply plugin: 'com.google.protobuf'
+
+group = 'io.improbable'
 archivesBaseName = "keanu"
 
 dependencies {
@@ -23,6 +29,7 @@ dependencies {
     compile 'org.nd4j:nd4j-native-platform:1.0.0-beta2'
     compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
     compile 'com.google.guava:guava:20.0'
+    compile 'com.google.protobuf:protobuf-java:3.6.1'
     compile group: 'com.opencsv', name: 'opencsv', version: '4.2'
     compile group: 'org.slf4j', name: 'slf4j-nop', version: '1.7.10'
 
--- a/keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
+++ b/keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
@@ -1,0 +1,38 @@
+syntax = "proto3";
+
+package keanu;
+
+option java_package = "io.improbable.keanu";
+option java_outer_classname = "KeanuSavedBayesNet";
+
+message BayesNet {
+    repeated Vertex vertices = 1;
+}
+
+message Vertex {
+    uint32 id = 1;
+    string label = 2;
+    string vertexType = 3;
+    VertexValue value = 4;
+    repeated HyperParams hyperParameters = 5;
+}
+
+message VertexValue {
+    enum ValueType {
+        DOUBLE = 0;
+        INT = 1;
+        BOOL = 2;
+        STRING = 3;
+    }
+
+    ValueType type = 1;
+    double doubleVal = 2;
+    int64 intVal = 3;
+    bool boolVal = 4;
+    string stringVal = 5;
+}
+
+message HyperParams {
+    string name = 1;
+    uint32 parentID = 2;
+}


keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
```diff
@@ -0,0 +1,38 @@
...
+    VertexValue value = 4;
+    repeated HyperParams hyperParameters = 5;
+}
+
+message VertexValue {
+    enum ValueType {
+        DOUBLE = 0;
+        INT = 1;
+        BOOL = 2;
+        STRING = 3;
```
migwellian: Do we need a `GENERIC` ValueType?

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    ValueType type = 1;
+    double doubleVal = 2;
+    int64 intVal = 3;
+    bool boolVal = 4;
+    string stringVal = 5;
```
migwellian: Yuk. Can't you just store the value in a bytes field? (disclaimer: I know nothing about protobufs)

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
```diff
@@ -0,0 +1,38 @@
...
+
+message BayesNet {
+    repeated Vertex vertices = 1;
+}
+
+message Vertex {
+    uint32 id = 1;
+    string label = 2;
+    string vertexType = 3;
+    VertexValue value = 4;
```
migwellian: do we need an `isObserved` field?

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
```diff
@@ -0,0 +1,38 @@
...
+
+    ValueType type = 1;
+    double doubleVal = 2;
+    int64 intVal = 3;
+    bool boolVal = 4;
+    string stringVal = 5;
+}
+
+message HyperParams {
+    string name = 1;
```
GeorgeNash: what's name here? Like `mu` or `sigma`?. Why do we need to store that?

keanu-project/build.gradle
```diff
@@ -15,6 +16,11 @@ plugins {
     id 'jacoco'
 }
 
+apply plugin: 'java'
```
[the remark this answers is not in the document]
christophernorth: Bad merge - fixed

keanu-project/build.gradle
```diff
@@ -15,6 +16,11 @@ plugins {
...
+apply plugin: 'jacoco'
+apply plugin: 'com.google.protobuf'
+
+group = 'io.improbable'
```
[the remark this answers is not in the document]
christophernorth: As above

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
```diff
@@ -0,0 +1,38 @@
...
+    uint32 parentID = 2;
```
GeorgeNash: Same comment here about `long` you probs want `int64`

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
joncoello83: I guess this depends on whether we are separating data and model structure definition. If we are then `isObserved` should probably be in with the data.

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    ValueType type = 1;
+    double doubleVal = 2;
+    int64 intVal = 3;
+    bool boolVal = 4;
+    string stringVal = 5;
+}
+
+message HyperParams {
```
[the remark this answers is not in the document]
christophernorth: Yeah agreed on Hyper Params - I don't think our current usage of "Parents" really encompasses what I'm trying to achieve though.  Parents in Keanu are sort of "unordered" (ie just those graph nodes that feed in to this one) whereas we need to know the exact semantic meaning to recreate the vertex later.  So yeah, they're the parents, but I was worried that would be confusing.  I've called them NamedInputs for now.

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
christophernorth: Yep - good call - I've added support for generic via a byte array.

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
christophernorth: Byte arrays are even more yucky (you're throwing away type info).  I've changed this to a oneof but the semantics are exactly the same as this.

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
christophernorth: We thought this should live in the "data layer" rather than the model description layer - it does mean until this comes in we cant save observations but I think thats ok.

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
```diff
@@ -0,0 +1,38 @@
...
+
+option java_package = "io.improbable.keanu";
+option java_outer_classname = "KeanuSavedBayesNet";
+
+message BayesNet {
+    repeated Vertex vertices = 1;
+}
+
+message Vertex {
+    uint32 id = 1;
```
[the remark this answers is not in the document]
christophernorth: yeah - silly slip, thanks

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
```diff
@@ -0,0 +1,38 @@
...
+    string label = 2;
+    string vertexType = 3;
```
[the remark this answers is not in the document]
christophernorth: This is effectively a class name (Gaussian etc)

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
christophernorth: Yeah, this is like mu/sigma/or a and b for minus etc.  We need this so that when we load we can correctly hook up the inputs (we can't just have a list of parent ids as we might use the mu as the sigma etc).

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
christophernorth: 👍

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
migwellian: That makes sense, although I think I don't yet understand why `value` must be part of the description layer.

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
```diff
@@ -0,0 +1,33 @@
+syntax = "proto3";
+
+package keanu;
+
+option java_package = "io.improbable.keanu";
+option java_outer_classname = "KeanuSavedBayesNet";
+
+message BayesNet {
```
gordoncaleb: BayesNet -> BayesianNetwork?

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
joncoello83: I've been on the fence about this for a bit and have had some good chats with a few people to inform my thinking. I'm currently favouring having a totally separate topology definition, then having additional data layers that are essentially k-v pairs that define data of different types. Migwellian and I just had a chat and we came up with a few different types of data that could potentially occupy separate layers. Interested in other people's thoughts: 
- constants ('real' constants and model defaults that users probably shouldn't consider changing)
- priors (latent variable hyperparameter values)
- observations (data used for model calibration)
- state (the specific values of latent variables at any point during an inference algorithm run, potentially stored for fault tolerance, etc.)

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
christophernorth: 👍

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
```diff
@@ -0,0 +1,33 @@
...
+    string label = 2;
+    string vertexType = 3;
+    VertexValue value = 4;
+    repeated NamedInputs inputs = 5;
+}
+
+message VertexValue {
+    oneof values {
+        double doubleVal = 1;
+        int64 intVal = 2;
```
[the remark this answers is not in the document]
christophernorth: this was a mistake - given our tensors only cover 32bits this should be an int32 (and the name should stay imo).

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
```diff
@@ -0,0 +1,33 @@
...
+        bool boolVal = 3;
+        string stringVal = 4;
+        bytes genericVal = 5;
+    }
+}
+
+message NamedInputs {
+    string name = 1;
+    int64 parentID = 2;
```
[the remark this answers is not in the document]
christophernorth: NamedParent is good with me - thanks.

PR #286, opened 2018-10-16
Automated release

GeorgeNash:
- Add an encrypted version of the gpg keys to the repo that can be decrypted by Travis
- If Travis is ran on master, it will run `uploadArchives` to release to: https://oss.sonatype.org/#stagingRepositories
- Otherwise, runs the classic `./gradlew check` command
- Add a script to automate the `git flow` steps

Testing
- Tested on a cloned version of the repo to make sure it works
- Also tested the `if branch == x` with the current branch name to make sure that works


.travis.yml
```diff
@@ -1,9 +1,10 @@
 language: java
 jdk:
-  - oraclejdk8
+- oraclejdk8
```
migwellian: this doesn't look right?

.travis.yml
```diff
@@ -1,9 +1,10 @@
...
 addons:
   sonarcloud:
-    organization: "improbable-keanu"
-script:
-  - ./gradlew check
-  - 'if [[ ! -z $SONAR_TOKEN ]]; then sonar-scanner; fi'
+    organization: improbable-keanu
```
migwellian: I don't think the quote marks make a difference, do they?

deployment/buildAndDeploy.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+if [[ $TRAVIS_BRANCH == 'master' ]]; then
+  ./gradlew uploadArchives -PnexusUser=$NEXUS_USER -PnexusPassword=$NEXUS_PASSWORD -Psigning.keyId=$SIGNING_KEY_ID -Psigning.password=$SIGNING_PASSWORD -Psigning.secretKeyRingFile=../deployment/secret-keys-keanu.gpg
```
migwellian: It looks like `uploadArchives` depends on `compileJava` but it doesn't depend on `check`, which we should still be running on master.

I think the best command to use is `./gradlew clean build uploadArchives`

Also worth sticking a `--info --stacktrace` on the end too.

deployment/buildAndDeploy.sh
```diff
@@ -0,0 +1,11 @@
...
+else
+  ./gradlew check
```
migwellian: FYI I've changed this on another branch (building Python) so will change it here too once I merge yours into mine.

deployment/release.sh
```diff
@@ -0,0 +1,9 @@
+#  version="USER INPUT"
+read -p "Enter version number: " version
```
migwellian: You know what would be really sweet is if it automatically worked out the version number. You can use `git describe --abbrev=0`

deployment/release.sh
```diff
@@ -0,0 +1,9 @@
...
+
+git checkout master
+git flow release start $version
+git flow release publish $version
+git flow release finish $version
+git push origin --tags
+git push
```
migwellian: why do you need to push twice?

.travis.yml
```diff
@@ -1,9 +1,10 @@
...
+script: deployment/buildAndDeploy.sh
+before_install:
+- openssl aes-256-cbc -K $encrypted_eec04b57ffc0_key -iv $encrypted_eec04b57ffc0_iv
+  -in deployment/secret-keys-keanu.gpg.enc -out deployment/secret-keys-keanu.gpg -d
```
migwellian: How does this work? Is `$encrypted_eec04b57ffc0_key` something that lives on the Travis server that nobody can steal?

migwellian: It's probably worth putting `**/*.gpg` in the `.gitignore`, even though it's unlikely (impossible?) that someone would generate it locally and accidently commit it.

deployment/release.sh
christophernorth: Also would be sweet if this fixed the version numbers in all he places we have to do it? :-)

Might be something for another ticket though!

.travis.yml
GeorgeNash: Funny reformat by travis - I'll revert

deployment/buildAndDeploy.sh
GeorgeNash: Nice I'll add that in

.travis.yml
GeorgeNash: Nah - travis auto-deleted them when I added the encrypted file

deployment/release.sh
GeorgeNash: `git push` doesn't push the tags so I have to do an explicit tag push

.travis.yml
GeorgeNash: Yep exactly, when you run `travis encrypt` on a file it creates that `.enc` version of the file and creates these two environmental variables for you.

For more: https://docs.travis-ci.com/user/encrypting-files/

GeorgeNash: Worth adding just in case - added.

deployment/release.sh
GeorgeNash: Yeah that would be cool - let me add a ticket for that

GeorgeNash: I'd rather be able to control it for a few reasons.

- Adds a manual step, so people don't commit to master and release by mistake
- Gives us more control over bumping to major versions
- We may want to skip versions etc

deployment/buildAndDeploy.sh
GeorgeNash: Nice!

deployment/release.sh
GeorgeNash: I'll print out the current version as an indicator at least!!!

christophernorth: For now, would it be worth just printing a reminder?  Something like:

Please update the version number to $ver in:
 - {Keanu locations}
 - {keanu starter}
 - {keanu docs stuff}

deployment/release.sh
migwellian: yeah, but what use is the second `git push`?

.travis.yml
migwellian: 👍

deployment/release.sh
GeorgeNash: Oh nice I thought it just pushed the tags!! I'll remove the extra push

deployment/buildAndDeploy.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+if [[ $TRAVIS_BRANCH == 'master' ]]; then
+  ./gradlew clean build -x :keanu-python:build -i uploadArchives -PnexusUser=$NEXUS_USER -PnexusPassword=$NEXUS_PASSWORD -Psigning.keyId=$SIGNING_KEY_ID -Psigning.password=$SIGNING_PASSWORD -Psigning.secretKeyRingFile=../deployment/secret-keys-keanu.gpg --info --stacktrace
```
migwellian: `-i` is a duplicate of `--info`. Also I'd prefer to see all the tasks together, i.e. `uploadArchives` should come directly after `build`

Trying out the funky new "suggestion" feature on github:

```suggestion
  ./gradlew clean build -x :keanu-python:build uploadArchives -PnexusUser=$NEXUS_USER -PnexusPassword=$NEXUS_PASSWORD -Psigning.keyId=$SIGNING_KEY_ID -Psigning.password=$SIGNING_PASSWORD -Psigning.secretKeyRingFile=../deployment/secret-keys-keanu.gpg --info --stacktrace
```

deployment/buildAndDeploy.sh
```diff
@@ -0,0 +1,11 @@
...
+else
+  ./gradlew check
```
migwellian:
```suggestion
  ./gradlew clean build -x :keanu-python:build --info --stacktrace
```

PR #288, opened 2018-10-17
Add ridge/lasso regression and reformat model API

CharlieCrisp:
This PR introduces a refactor to change how Keanu users can create linear and logistic regression models. It also introduces the ability to create models that use ridge and lasso regression which both use regularisation to minimise the weight vector norms.


keanu-project/src/test/java/io/improbable/keanu/e2e/regression/DiabetesLinearRegression.java
```diff
@@ -57,23 +55,20 @@ public void doesLinearRegressionOnBMI() {
 
     @Test
     public void doesLinearRegressionOnBMIAsModel() {
-        Data data = ReadCsv
-            .fromResources("data/datasets/diabetes/diabetes_standardized_training.csv")
-            .asVectorizedColumnsDefinedBy(Data.class)
-            .load(true);
+        Data data = readData();
 
-        LinearRegression regression = new LinearRegression(data.bmi, data.y, 0.0, 2.0, 100);
```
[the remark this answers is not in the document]
CharlieCrisp: I believe this was a sigma for weights and was actually the default value anyway which is why it has been removed

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/LinearRegressionTestUtils.java
```diff
@@ -0,0 +1,118 @@
...
+import io.improbable.keanu.vertices.ConstantVertex;
+import io.improbable.keanu.vertices.dbl.DoubleVertex;
+import io.improbable.keanu.vertices.dbl.probabilistic.GaussianVertex;
+import io.improbable.keanu.vertices.dbl.probabilistic.LaplaceVertex;
+import io.improbable.keanu.vertices.dbl.probabilistic.UniformVertex;
+import lombok.Value;
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+class LinearRegressionTestUtils {
```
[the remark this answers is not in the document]
CharlieCrisp: Could you clarify how that would help? I'm not sure I see how that would differ from the current implementation?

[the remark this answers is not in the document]
migwellian: Admittedly it's only a half-thought-through idea, but basically: the purpose of this class is to set something up at the start of the test, and assert something at the end of the test. That is essentially what Rules do. For example, take a look at the implementation of `org.junit.rules.ExpectedException`.

This is very much an optional extra so don't feel like you have to pursue it if you don't have time.

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/DiabetesLinearRegression.java
```diff
@@ -24,17 +25,11 @@
...
  */
 public class DiabetesLinearRegression {
-
-    private static Data readData() {
-        return ReadCsv
-            .fromResources("data/datasets/diabetes/diabetes_standardized_training.csv")
-            .asVectorizedColumnsDefinedBy(Data.class)
-            .load(true);
-    }
+    @Rule CsvDataResource<Data> csvDataResource = new CsvDataResource("data/datasets/diabetes/diabetes_standardized_training.csv", Data.class);
```
migwellian: Very nice!

keanu-project/src/main/java/io/improbable/keanu/model/regression/RegressionGraph.java
```diff
@@ -13,7 +13,7 @@
 import lombok.Getter;
 import lombok.Value;
 
-public class LinearRegressionGraph<OUTPUT> implements ModelGraph<DoubleTensor, OUTPUT> {
+public class RegressionGraph<OUTPUT> implements ModelGraph<DoubleTensor, OUTPUT> {
```
migwellian: I think `LinearRegressionGraph` is the better name. Logistic Regression is still linear so it's appropriate to use a LinearRegressionGraph in both models.

keanu-project/src/main/java/io/improbable/keanu/model/regression/RegressionModelBuilder.java
```diff
@@ -0,0 +1,217 @@
...
+            setPriorOnWeights(RegressionWeights.fillPriorOnWeights(this.inputTrainingData.getShape(), DEFAULT_MU), RegressionWeights.fillPriorOnWeights(this.inputTrainingData.getShape(), DEFAULT_SCALE_PARAMETER));
+        }
+
+        if (priorOnInterceptMean == null || priorOnInterceptScaleParameter == null) {
+            setPriorOnIntercept(DEFAULT_MU, DEFAULT_SCALE_PARAMETER);
+        }
+    }
+
+    private DoubleVertex getInterceptVertex() {
+        switch (this.regularization) {
```
migwellian: The trouble with enums is you end up with ugly `switch` statements. If you use a class hierarchy instead then you can do it all with method dispatch. By adding methods to your enum you can get the best of both worlds. Then your method would just be:

    private DoubleVertex getInterceptVertex() {
        return this.regularization.getInterceptVertex(priorOnInterceptMean, priorOnInterceptScaleParameter)
    }

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/LogisticModelScoreTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+    @Test
+    public void accuracyOfEverythingWrongIs0() {
+        BooleanTensor test = BooleanTensor.create(true, false, true, false);
+        assertEquals(ModelScoring.accuracy(test, test.not()), 0, 1e-10);
+    }
+
+    @Test
+    public void accuracyOfHalfIsHalf() {
+        BooleanTensor test = BooleanTensor.create(true, false, true, false);
```
migwellian: unused

keanu-project/src/main/java/io/improbable/keanu/model/regression/RegressionModelBuilder.java
```diff
@@ -0,0 +1,217 @@
...
+    private RegressionRegularization regularization = RegressionRegularization.NONE;
+    private double observationSigma = DEFAULT_SCALE_PARAMETER;
+    private double[] priorOnWeightsScaleParameters;
+    private double[] priorOnWeightsMeans;
+    private Double priorOnInterceptScaleParameter;
+    private Double priorOnInterceptMean;
+    private DoubleTensor inputTrainingData;
+    private DoubleTensor outputLinearTrainingData;
+    private BooleanTensor outputLogisticTrainingData;
+
```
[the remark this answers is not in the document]
migwellian: I think we can achieve a builder API like this:

        RegressionModel<> model = RegressionModel.withTrainingData(inputData, outputData)
            .withRegularization(RegressionRegularization.RIDGE)
            .withPriorOnWeightsAndIntercept(0, 100)
            .withObservationSigma(100)
            .build();

keanu-project/src/main/java/io/improbable/keanu/model/regression/RegressionModelBuilder.java
```diff
@@ -0,0 +1,143 @@
...
+        return this.regularization.getInterceptVertex(priorOnInterceptMean, priorOnInterceptScaleParameter);
+    }
+
+    private DoubleVertex getWeightsVertex() {
+        return this.regularization.getWeightsVertex(getFeatureCount(), priorOnWeightsMeans, priorOnWeightsScaleParameters);
+    }
+
+    private void performDataFitting(LinearRegressionGraph<OUTPUT> regressionGraph, OUTPUT outputTrainingData) {
+        ModelFitter<DoubleTensor, OUTPUT> fitter;
+        switch (this.regularization) {
```
migwellian: You could get rid of this switch statement, and instead have:

    ModelFitter<DoubleTensor, OUTPUT> fitter = this.regularization.createFitterFor(regressionGraph);
    fitter.fit(inputTrainingData, outputTrainingData);

PR #290, opened 2018-10-18
Added the ability to just set the vetex label as a string - auto-crea…

christophernorth:
…ting the label object (syntatic sugar only)


GeorgeNash:
LGTM!

PR #291, opened 2018-10-18
Feature/python optimiser

GeorgeNash:
- Wraps the gradient optimizer and non gradient optimizer builders.
- Lets users call map and max likelihood.

Any ideas on how to test that the private fields of the optimizer are being set correctly by the builder??


keanu-python/keanu/optimization.py
```diff
@@ -0,0 +1,58 @@
...
+k = KeanuContext().jvm_view()
+
+java_import(k, "io.improbable.keanu.algorithms.variational.optimizer.gradient.GradientOptimizer")
+java_import(k, "io.improbable.keanu.algorithms.variational.optimizer.nongradient.NonGradientOptimizer")
+
+
+class Optimizer:
+    def __init__(self, optimizer):
+        self.optimizer = optimizer
+        self.net = optimizer.net
```
migwellian: unused

keanu-python/keanu/optimization.py
```diff
@@ -0,0 +1,58 @@
...
+        return self.optimizer.maxLikelihood()
+
+
+class GradientOptimizer(Optimizer):
+    def __init__(self, net, max_evaluations=None, relative_threshold=None, absolute_threshold=None):
+        builder = k.GradientOptimizer.builder()
+        if isinstance(net, BayesNet):
+            builder = builder.bayesianNetwork(net.unwrap())
+        elif isinstance(net, Vertex):
+            builder = builder.bayesianNetwork(BayesNet(net.unwrap().getConnectedGraph()).unwrap())
```
migwellian: `else raise` something?

keanu-python/keanu/optimization.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def max_likelihood(self):
+        return self.optimizer.maxLikelihood()
+
+
+class GradientOptimizer(Optimizer):
+    def __init__(self, net, max_evaluations=None, relative_threshold=None, absolute_threshold=None):
+        builder = k.GradientOptimizer.builder()
+        if isinstance(net, BayesNet):
+            builder = builder.bayesianNetwork(net.unwrap())
```
migwellian: you don't need the `builder = ` here or on subsequent lines. Once you have a reference to the builder (line 27) you're good.

keanu-python/keanu/optimization.py
```diff
@@ -0,0 +1,58 @@
...
+
+class NonGradientOptimizer(Optimizer):
+    def __init__(self, net, max_evaluations=None, bounds_range=None, initial_trust_region_radius=None, stopping_trust_region_radius=None):
+        builder = k.NonGradientOptimizer.builder()
+        if isinstance(net, BayesNet):
+            builder = builder.bayesianNetwork(net.unwrap())
+        elif isinstance(net, Vertex):
+            builder = builder.bayesianNetwork(BayesNet(net.unwrap().getConnectedGraph()).unwrap())
+        if max_evaluations is not None:
+            builder = builder.maxEvaluations(max_evaluations)
```
migwellian: It would be nice to factor out the code common with  `GradientOptimizer.__init__`

keanu-python/keanu/optimization.py
```diff
@@ -0,0 +1,60 @@
...
+        return self.optimizer.maxAPosteriori()
+
+    def max_likelihood(self):
+        return self.optimizer.maxLikelihood()
+
+def build_bayes_net(builder, net):
+    if isinstance(net, BayesNet):
+        return builder.bayesianNetwork(net.unwrap()), net
+    elif isinstance(net, Vertex):
+        net = BayesNet(net.unwrap().getConnectedGraph())
```
icantsayrural: You only have to unwrap `JavaObjectWrapper` when you are passing it as an argument. If you are just calling the method, the unwrapping is done for you in `__getattr__`.

keanu-python/keanu/optimization.py
```diff
@@ -0,0 +1,60 @@
...
+        self.optimizer = optimizer
+        self.net = net
+
+    def max_a_posteriori(self):
+        return self.optimizer.maxAPosteriori()
+
+    def max_likelihood(self):
+        return self.optimizer.maxLikelihood()
+
+def build_bayes_net(builder, net):
```
icantsayrural: Should this be private?

keanu-python/keanu/inference.py
```diff
@@ -5,16 +5,14 @@
 k = KeanuContext().jvm_view()
 
 java_import(k, "io.improbable.keanu.network.BayesianNetwork")
-
+java_import(k, "io.improbable.keanu.algorithms.mcmc.MetropolisHastings")
 
 class BayesNet(JavaObjectWrapper):
     def __init__(self, vertices):
+        self.vertices = vertices
```
icantsayrural: If we just need it for tests, I think it'll be better if we just call Java's `.getVertices().size()` directly?

keanu-python/keanu/optimization.py
```diff
@@ -0,0 +1,62 @@
...
+        self.net = net
+
+    def max_a_posteriori(self):
+        return self.optimizer.maxAPosteriori()
+
+    def max_likelihood(self):
+        return self.optimizer.maxLikelihood()
+
+    @staticmethod
+    def build_bayes_net(builder, net):
```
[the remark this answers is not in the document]
GeorgeNash: As discussed - doing one underscore

keanu-python/keanu/optimization.py
```diff
@@ -0,0 +1,62 @@
...
+        self.net = net
+
+    def max_a_posteriori(self):
+        return self.optimizer.maxAPosteriori()
+
+    def max_likelihood(self):
+        return self.optimizer.maxLikelihood()
+
+    @staticmethod
+    def _build_bayes_net(builder, net):
```
migwellian: double underscore is preferred for private methods - it does name mangling.

https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-a-single-and-a-double-underscore-before-an-object-name

keanu-python/keanu/optimization.py
```diff
@@ -0,0 +1,62 @@
...
+        if isinstance(net, BayesNet):
+            return builder.bayesianNetwork(net.unwrap()), net
```
migwellian: this line appears twice, so it would be neater to move it to the end of the method

keanu-python/keanu/optimization.py
GeorgeNash: As discussed offline - will leave this as a single underscore to avoid the nasty `_Optimizer__build_bayes_net`

GeorgeNash: As discussed offline - will leave this as a single underscore to avoid the nasty `_Optimizer__build_bayes_net`

PR #292, opened 2018-10-18
Feature/python operator overload

migwellian:
I'm posting this early for review: I haven't implemented all operators yet but this PR captures the most interesting things about making it work with numbers and ndarrays.

Currently supported: `+ - * / > <`

Also included is a change to `Const` so that it's not a class. It's now just a function that returns a `Vertex`. This makes it easier to use consts with the operators.


keanu-python/keanu/vertex.py
```diff
@@ -1,18 +1,67 @@
 import numpy as np
 import numbers
 
+import keanu as kn
 from keanu.context import KeanuContext
 from keanu.base import JavaObjectWrapper
-from keanu.const import Const
```
migwellian: I had to remove this to avoid a circular import Vertex <--> Const

keanu-python/tests/keanu_assert.py
```diff
@@ -0,0 +1,44 @@
+"""
+Some utility functions to compare tensor-like things
+They might be Keanu tensors or Numpy arrays
+"""
+def tensors_equal(t1, t2):
+    if not shapes_equal(__get_shape(t1), __get_shape(t2)):
+        return False
+        print("Shapes don't match", __get_shape(t1), __get_shape(t2))
+        return False
+    it1 = __as_list(t1)
```
christophernorth: Should we use generators here rather than lists?  I guess the tensors could get quite big in theory so not having to copy them would be good?

keanu-python/tests/keanu_assert.py
```diff
@@ -0,0 +1,44 @@
...
+    it2 = __as_list(t2)
+    for i in range(len(it1)):
```
christophernorth: perhaps a bit more pythonic to use:

for idx, value = enumerate(zip(t1, t2)) ?

keanu-python/tests/keanu_assert.py
```diff
@@ -0,0 +1,44 @@
+"""
+Some utility functions to compare tensor-like things
+They might be Keanu tensors or Numpy arrays
+"""
+def tensors_equal(t1, t2):
+    if not shapes_equal(__get_shape(t1), __get_shape(t2)):
+        return False
```
christophernorth: I think this line might be a typo?

keanu-python/tests/test_vertex_operations.py
```diff
@@ -0,0 +1,79 @@
...
+    (kn.Const(np.array([15., 10.])), kn.Const(np.array([2., 4.])), np.array([[7.5], [2.5 ]])),
+    (kn.Const(np.array([15., 10.])),          np.array([2., 4.]) , np.array([[7.5], [2.5 ]])),
+    (         np.array([15., 10.]) , kn.Const(np.array([2., 4.])), np.array([[7.5], [2.5 ]])),
+    (kn.Const(np.array([15., 10.])),                    2.       , np.array([[7.5], [5.  ]])),
+    (                   15.,         kn.Const(np.array([2., 4.])), np.array([[7.5], [3.75]])),
+])
+def test_can_do_division(lhs, rhs, expected_result):
+    result = lhs / rhs
+    assert type(result) == kn.Vertex
+    assert tensors_equal(result.getValue(), expected_result)
```
christophernorth: Should we add a compound test (ie one that does lots of different ops on a line?)

keanu-python/tests/keanu_assert.py
migwellian: nice!

keanu-python/tests/keanu_assert.py
migwellian: Good call. You mean I should use generator expressions instead of list comprehensions inside the `__as_list` function, right?

keanu-python/tests/keanu_assert.py
migwellian: 👍

keanu-python/tests/test_vertex_operations.py
migwellian: sure

keanu-python/tests/keanu_assert.py
christophernorth: Yep exactly - sorry - should have been clearer!

keanu-python/keanu/vertex.py
icantsayrural: You could just also import it inline

keanu-python/keanu/vertex.py
```diff
@@ -30,9 +79,9 @@ def __parse_args(args):
     @staticmethod
     def __parse_arg(arg):
         if isinstance(arg, np.ndarray):
-            return Const(arg).unwrap()
+            return kn.Const(arg).unwrap()
         elif isinstance(arg, numbers.Number):
-            return arg
+            return Vertex.__parse_arg(np.array([[arg]]))
```
icantsayrural: If we make all numbers passed to a vertex ndarray, `__infer_tensor_from_scalar` in tenso.py is never reached

icantsayrural: Oh, nevermind it can still be used in `observe`

icantsayrural: I think it was my mistake - but could you fix the import statement for `observe`?

keanu-python/keanu/vertex.py
migwellian: Good idea. In this case I will leave it since doing `import keanu as kn` instead breaks the circular dependency, and I use it elsewhere in this file too.

keanu-python/tests/test_vertex_operations.py
migwellian: done

keanu-python/keanu/vertex.py
migwellian: done

keanu-python/keanu/vertex.py
```diff
@@ -34,9 +119,9 @@ def __parse_args(args):
     @staticmethod
     def __parse_arg(arg):
         if isinstance(arg, np.ndarray):
```
GeorgeNash: This logic of checking the type for `numpy` arrays and checking for `number` and casting that to a rank 2 `numpy` array feels identical to the logic in the constructor of `Const`. Can we pull it out into a method?

keanu-python/tests/keanu_assert.py
```diff
@@ -0,0 +1,46 @@
...
+
+def shapes_equal(s1, s2):
+    if __get_length(s1) != __get_length(s2):
+        print("Ranks don't match: %s vs %s" % (__as_list(s1), __as_list(s2)))
+        return False
+    for i in range(len(s2)):
+        if s1[i] != s2[i]:
+            print("Mismatch in dimension %d: %d != %d" % (i, s1[i], s2[i]))
+            return False
+        return True
```
GeorgeNash: Think your indentation is wrong here - is this always returning `True` after the first iteration? You need to bring it in one.

keanu-python/tests/test_vertex_operations.py
```diff
@@ -0,0 +1,227 @@
+import keanu as kn
+import numpy as np
+import pytest
+import math
+from tests.keanu_assert import tensors_equal
+
+
+### Comparisons
+
+@pytest.mark.parametrize("lhs, rhs, expected_result", [
```
GeorgeNash: Really cool use of `parametrize`!

keanu-python/tests/keanu_assert.py
migwellian: good spot!

migwellian: this file actually gets nuked in my later PR, but I've made the change as you suggest

keanu-python/keanu/vertex.py
```diff
@@ -1,30 +1,115 @@
...
+                raise NotImplementedError("NumPy ufunc of type %s not implemented" % ufunc.__name__)
+        else:
+            raise NotImplementedError("NumPy ufunc method %s not implemented" % method)
 
     def __add__(self, other):
-        from keanu.generated.vertex import Addition
-        return Addition(self, other)
+        return kn.generated.vertex.Addition(self, other)
+
+    def __radd__(self, other):
```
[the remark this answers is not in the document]
migwellian: not really, because `self` has a special meaning.

It would be possible for commutative operations like `+` - i.e. here I could do:

    def __radd__(self, other):
        return self.__add__(other)

but that would not work for non-commutative operations like `-`. I could do:

    def __rsub__(self, other):
        return Vertex.__sub__(other, self)

but that seems like an abuse of the concept of `self`. Overall I think it all looks better as it is.

PR #294, opened 2018-10-19
Document BOBYQA Non Gradient Optimizer better

CharlieCrisp:
This PR introduces some basic documentation to explain how the `NonGradientOptimizer` works. This includes exposing the fact that it uses BOBYQA under the hood, how it optimizes latent variables to maximise the probability of the observed variables, and how it will mutate the Bayesian network in place.


CharlieCrisp:
Realised there's actually more work to be done here

PR #296, opened 2018-10-22
Refactor/refactor particle filtering classes

CharlieCrisp:
This PR introduces a suggestion as to how we could refactor the ParticleFilter Code to make it easier to understand from a Keanu developer and user point of view.

This is open to inputs if anyone disagrees with this way of structuring code but the thinking is that this brings the API in line with the `RegressionModel` API.


keanu-project/src/main/java/io/improbable/keanu/algorithms/particlefiltering/Particle.java
```diff
@@ -0,0 +1,83 @@
...
+     * @return log probability of the subgraph occuring in it's current state
+     */
+    public double logProb() {
+        return sumLogPOfSubgraph;
+    }
+
+    /**
+     * @return probability of the subgraph occuring in it's current state
+     */
+    public double prob() { return Math.exp(sumLogPOfSubgraph); }
```
gordoncaleb: This is going to be very numerically unstable. Most interesting networks are going to return 0 here.

CharlieCrisp: How do you reckon we should go about dealing with this? We could remove this functionality for the moment and then raise a ticket to fix this problem?

keanu-project/src/main/java/io/improbable/keanu/algorithms/particlefiltering/Particle.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Represents a Particle used in a Particle Filter.
+ * A Particle can be thought of as a particular instance of the network with state and an associated probability.
+ */
+public class Particle {
```
GeorgeNash: Why doesn't this class have a constructor?

PR #297, opened 2018-10-23
Broken regression test

gordoncaleb:
One of the regression model tests used a lambda (scale) that meant it didn't get close enough to the expected intercept value. I've decreased the scale causing it to have a tighter constraint on the intercept. I've also removed my name from the maven pom.

### DIFF
--- a/keanu-project/build.gradle
+++ b/keanu-project/build.gradle
@@ -84,9 +84,9 @@ uploadArchives {
 
                 developers {
                     developer {
-                        id 'caleb'
-                        name 'Caleb Walker'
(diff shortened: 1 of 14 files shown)


keanu-project/build.gradle
```diff
@@ -84,9 +84,9 @@ uploadArchives {
 
                 developers {
                     developer {
-                        id 'caleb'
-                        name 'Caleb Walker'
-                        email 'caleb@improbable.io'
+                        id 'keanu-engineering'
```
GeorgeNash: RIP!

PR #298, opened 2018-10-23
Check for invalid sampling arguments

gordoncaleb:
Disallow negative sample dropping, non-positive down sample intervals and dropping more samples than requesting.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/NetworkSamples.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/NetworkSamples.java
@@ -1,5 +1,6 @@
 package io.improbable.keanu.algorithms;
 
+import com.google.common.base.Preconditions;
 import io.improbable.keanu.network.NetworkState;
 import io.improbable.keanu.tensor.dbl.DoubleTensor;
 import io.improbable.keanu.tensor.intgr.IntegerTensor;
@@ -62,6 +63,10 @@ public class NetworkSamples {
     }
 
     public NetworkSamples drop(int dropCount) {
+        Preconditions.checkArgument(dropCount >= 0, "Cannot drop %s samples. Drop count must be positive.", dropCount);
+        if (dropCount == 0) {
+            return this;
+        }
 
         final Map<VertexId, List<?>> withSamplesDropped = samplesByVertex.entrySet().parallelStream()
             .collect(toMap(
@@ -74,6 +79,7 @@ public class NetworkSamples {
     }
 
     public NetworkSamples downSample(final int downSampleInterval) {
+        Preconditions.checkArgument(downSampleInterval > 0, "Down sample interval of %s is invalid. Sample interval must be positive.", downSampleInterval);
 
         final Map<VertexId, List<?>> withSamplesDownSampled = samplesByVertex.entrySet().parallelStream()
             .collect(toMap(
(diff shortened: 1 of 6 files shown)


keanu-project/src/main/java/io/improbable/keanu/algorithms/NetworkSamples.java
```diff
@@ -62,6 +63,10 @@ public IntegerTensorVertexSamples getIntegerTensorSamples(VertexId vertexId) {
     }
 
     public NetworkSamples drop(int dropCount) {
+        Preconditions.checkArgument(dropCount >= 0, "Cannot drop %s samples. Drop count must be positive.", dropCount);
+        if (dropCount == 0) {
+            return this;
```
GeorgeNash: Can we chuck a `log.warn` statement in here saying "You're trying to drop zero samples, this will do nothing"

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/NetworkSamplesGenerator.java
```diff
@@ -36,7 +28,39 @@ public NetworkSamplesGenerator(SamplingAlgorithm algorithm, Supplier<ProgressBar
         this.progressBarSupplier = progressBarSupplier;
     }
 
+    public int dropCount() {
```
GeorgeNash: `getDropCount`?

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/NetworkSamplesGenerator.java
```diff
@@ -36,7 +28,39 @@ public NetworkSamplesGenerator(SamplingAlgorithm algorithm, Supplier<ProgressBar
...
+    public NetworkSamplesGenerator dropCount(int dropCount) {
+        Preconditions.checkArgument(dropCount >= 0,
+            "Drop count of %s is invalid. Cannot drop negative samples.",
+            dropCount
+        );
+        this.dropCount = dropCount;
+        return this;
+    }
+
+    public int downSampleInterval() {
```
GeorgeNash: `getDownSampleInterval`?

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/LinearLassoRegressionTest.java
```diff
@@ -55,7 +55,7 @@ public void findsExpectedParamsForManyWeights() {
 
         RegressionModel linearRegressionModel = RegressionModel.withTrainingData(data.xTrain, data.yTrain)
             .withRegularization(RegressionRegularization.LASSO)
-            .withPriorOnIntercept(0, 40)
+            .withPriorOnIntercept(0, 20)
```
GeorgeNash: Think you fixed this on another PR?

keanu-project/src/main/java/io/improbable/keanu/algorithms/NetworkSamples.java
gordoncaleb: That's part of the contract of drop I'd say. It's not ambiguous what drop(0) means and it is valid so I don't think we should bother the user with it. I'll javadoc the method to call out that dropping 0 does nothing.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/NetworkSamplesGenerator.java
gordoncaleb: This came from from delomboking. If it's getDropCount then it should be setDropCount, which I didn't like and isn't backwards compatible.

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/LinearLassoRegressionTest.java
gordoncaleb: Yup, it was just to get it building locally. I'll pull develop in.

PR #299, opened 2018-10-24
Feature/python lorenz

migwellian:
implement the Lorenz example in python


migwellian:
see feature/python-with-numpy-values which includes the Lorenz stuff

PR #308, opened 2018-10-26
Refactor categorical to know tensor implementation class

gordoncaleb:
This PR uses some java generic magic and a few instanceof in order to make the Categorical vertex generic AND take advantage of the typed tensor classes (e.g. DoubleTensor)


keanu-project/src/main/java/io/improbable/keanu/vertices/generic/probabilistic/discrete/CategoricalVertex.java
```diff
@@ -21,50 +26,55 @@
 import static java.util.stream.Collectors.toMap;
 
 
-public class CategoricalVertex<T> extends Vertex<Tensor<T>> implements Probabilistic<Tensor<T>> {
+public class CategoricalVertex<T, TENSOR extends Tensor<T>> extends Vertex<TENSOR> implements Probabilistic<TENSOR> {
```
icantsayrural: I wanted to avoid doing this, because I'm not a big fan of forcing people to specify both the element type and the tensor type when instantiating a categorical vertex. But I guess there is no other way...

gordoncaleb: I think this is necessary but I agree it's ugly and not something the user should care about. What we can do is define a Typed class that has T and TENSOR filled in. `e.g. CategoricalIntegerVertex extends CategoricalVertex<Integer, IntegerTensor>`.

keanu-project/src/main/java/io/improbable/keanu/tensor/generic/GenericTensor.java
```diff
@@ -19,14 +19,22 @@
...
 
+    public static <T> GenericTensor<T> create(T[] data, long[] shape) {
+        return new GenericTensor<>(data, shape);
+    }
+
     public static <T> GenericTensor<T> scalar(T data) {
         return new GenericTensor<>(data);
     }
 
+    public static <T> GenericTensor<T> placeHolder(long[] shape) {
```
migwellian: @gordoncaleb is removing the `placeHolder` methods in another PR - it looks like you've already merged his changes - so it feels like the wrong design choice to add it back in here.

keanu-project/src/main/java/io/improbable/keanu/vertices/generic/probabilistic/discrete/CategoricalVertex.java
```diff
@@ -21,50 +26,55 @@
...
+        return new CategoricalVertex<>(toDoubleVertices(selectableValues));
     }
 
-    public static <T> CategoricalVertex<T> of(DirichletVertex vertex, List<T> categories) {
+    private static <T> Map<T, DoubleVertex> toDoubleVertices(Map<T, Double> selectableValues) {
+        Map<T, DoubleVertex> copy = new LinkedHashMap<>();
+        for (Map.Entry<T, Double> entry : selectableValues.entrySet()) {
+            copy.put(entry.getKey(), ConstantVertex.of(entry.getValue()));
+        }
+        return copy;
```
migwellian: return selectableValues.entrySet().stream()
            .collect(toMap(
                Map.Entry::getKey, 
                e -> ConstantVertex.of(e.getValue())
            ));

keanu-project/src/main/java/io/improbable/keanu/tensor/generic/GenericTensor.java
```diff
@@ -19,14 +19,22 @@
     private long[] shape;
     private long[] stride;
 
-    public static <T> GenericTensor<T> create(T data, long[] shape) {
+    public static <T> GenericTensor<T> createFilled(T data, long[] shape) {
```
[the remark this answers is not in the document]
gordoncaleb: create(T data, long[] shape) gets confused when you pass SomeType[]. It thinks T is SomeType[] instead of SomeType. You're right that if we change this here then it should be consistent with the other types. I'll rename them too?

PR #313, opened 2018-10-30
Extract keanu version from sources

CharlieCrisp:
This PR uses a current_version variable to extract the keanu version from links. It also removes the need for the `current: true` line in our `previous_versions.yml` data file.


docs/_config.yml
```diff
@@ -1,3 +1,5 @@
+current_version: 0.0.15
```
migwellian: gradle uses `gitVersion()` to set its version parameter. Can we want to use that to set `current_version` here?

The `gitVersion()` is the output of `git describe` which includes the git hash (e.g. `0.0.15-1-g7dd6ebfa` which means we are 1 commit past 0.0.15 and its git hash is 7dd6ebfa). You can also do `git describe --abbrev=0` which gives you just the major version, i.e. `0.0.15`

CharlieCrisp: I'm not sure if that's a possibility with Jekyll/its use of YAML. All i've seen so far is that it's very inflexible in the dynamic sense. I may be wrong though...

PR #316, opened 2018-10-30
Feature/seed random in gradient test

GeorgeNash:
To stop the flaky test failing!


keanu-python/keanu/keanu_random.py
```diff
@@ -13,3 +13,6 @@ def __init__(self, seed=None):
             super(KeanuRandom, self).__init__(k.jvm_view().KeanuRandom())
         else:
             super(KeanuRandom, self).__init__(k.jvm_view().KeanuRandom(seed))
+
+    def set_default_random_seed(self, seed):
+    	self.unwrap().setDefaultRandomSeed(seed)
```
icantsayrural: This is instantiating the `KeanuRandom` object and calling a static method on the instance, which is bad practice.

We'd want something like `k.jvm_view().KeanuRandom.setDefaultRandomSeed(seed)`?

GeorgeNash: much better - changed

PR #318, opened 2018-10-31
Feature/tensor equals scalar

GeorgeNash:
Enables a user to do an equals between:

- `DoubleTensor` and `scalar`
- `IntegerTensor` and `scalar`


keanu-project/src/test/java/io/improbable/keanu/tensor/intgr/Nd4jIntegerTensorTest.java
```diff
@@ -638,4 +641,20 @@ public void youCanFixACustomValidationIssueByReplacingTheValue() {
         containsMinusOne = validator.validate(containsMinusOne);
         assertThat(containsMinusOne, equalTo(expectedResult));
     }
+
+    @Test
+    public void comparesIntegerTensorWithScalar() {
+        IntegerTensor value = IntegerTensor.create(1, 2, 3);
+        IntegerTensor differentValue = IntegerTensor.create(1);
+        BoolVertex result = new EqualsVertex<>(ConstantVertex.of(value), ConstantVertex.of(differentValue));
```
icantsayrural: If we are using `EqualsVertex`, shouldn't this be in `EqualsVertexTest`? Otherwise just test `elementWiseEquals` directly.

GeorgeNash: No - we're testing tensor logic here. But yeah - I've removed the unnecessary creation of the vertices and kept them as tensors

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -1045,9 +1045,14 @@ public BooleanTensor greaterThanOrEqual(DoubleTensor value) {
 
     @Override
     public BooleanTensor elementwiseEquals(Tensor that) {
-
-        if (that instanceof Nd4jDoubleTensor) {
-            INDArray eq = tensor.eq(unsafeGetNd4J((Nd4jDoubleTensor) that));
+        INDArray eq;
```
migwellian: I'm not a fan of uninitialised variables. I think this is a scenario where the `? :` syntax is preferable:

        if (that instanceof DoubleTensor) {
            INDArray eq = that.isScalar() ?
                tensor.eq(((DoubleTensor) that).scalar()) : 
                tensor.eq(unsafeGetNd4J((Nd4jDoubleTensor) that));
            return fromMask(eq, getShape());

migwellian: likewise in `Nd4jIntegerTensor.java`

PR #319, opened 2018-10-31
Feature/save load gpb

christophernorth:
Not a "true" PR - just here to demonstrate the proposed approach.  Only works for Gaussian + Constant Double at the moment and  few other limitations:

- Tensors don't work at the moment when saving values
- Functions that probably should be abstract, aren't.

I've decided to take this approach to avoid having a giant factory/horrible parsing/casting in the load code, but let me know if you don't like the new constructors and we can discuss.


icantsayrural:
Cool! Without looking too much into it, the approach looks good to me so far.

Can we just make these protobuf related constructors package private since everything is saved from and loaded to a network?

build.gradle
```diff
@@ -10,6 +10,7 @@ allprojects {
     version gitVersion()
 
     apply plugin: "java"
+    apply plugin: "idea"
```
migwellian: how come you've introduced this here?

I think it would be a good thing if we started using `gradlew idea` to generate the IntelliJ dependencies instead of relying on the IntelliJ plugin which has caused people a lot of grief.

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
```diff
@@ -249,4 +305,32 @@ public String toString() {
...
     }
+
+    public final KeanuSavedBayesNet.Vertex toProtoBuf() {
+        KeanuSavedBayesNet.Vertex.Builder vertexBuilder = KeanuSavedBayesNet.Vertex.newBuilder();
+
+        if (label != null) {
+            vertexBuilder = vertexBuilder.setLabel(label.toString());
+        }
+
+        if (isConstantVertex(this)) {
```
migwellian: The fact that you need to check the type here feels like a code smell. Can you instead put the `toProtoBuf` logic into a separate class, and use polymorphism (maybe the visitor pattern) to decide whether to `setConstantValue`?

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
```diff
@@ -80,6 +87,12 @@ public int getIndentation() {
         return idValues.length;
     }
 
+    public KeanuSavedBayesNet.VertexID toProtoBuf() {
```
migwellian: `toProtoBuf<T>` looks like a good candidate for an interface

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
```diff
@@ -103,6 +155,11 @@ public void setValue(T value) {
         }
     }
 
+    public void setValue(KeanuSavedBayesNet.VertexValue value) {
+        throw new UnsupportedOperationException("This Vertex Type doesn't support loading from GPB");
```
migwellian: If instead you create a new interface for these methods then for now you can have just `GaussianVertex` implement it, and you can move the interface higher up the `Vertex` class hierarchy as you implement it in more places.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/GaussianVertex.java
```diff
@@ -126,4 +133,19 @@ public PartialDerivatives forwardModeAutoDifferentiation(Map<Vertex, PartialDeri
             return PartialDerivatives.withRespectToSelf(this.getId(), this.getShape());
         }
     }
+
+    @Override
+    public List<KeanuSavedBayesNet.NamedParent> getNamedParents() {
```
migwellian: It feels wrong that `getParents` and `getNamedParents` are logically independent of each other.

build.gradle
christophernorth: Without this - IntelliJ can't find the protoc generated class files so won't autocomplete on them etc.

This plugin does seem to work really well!

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
christophernorth: Agreed - it was a small quick addition, but I didn't want to overcomplicate the API too much for what each Vertex.  Interested if you have an example of how you'd like to see this work.

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
christophernorth: 100% - good suggestion.  I'll get this updated.

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
christophernorth: Yeah - I agree with the sentiment - I guess I just did it this way for quick PoC purposes.  Eventually though I see this as just something *every* vertex will need to support, so thought the Vertex base class made sense (as otherwise we'll need to do a cast in the loading code anyway).

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/GaussianVertex.java
christophernorth: Ok - I'll try and think of a better name for this.

keanu-project/src/main/java/io/improbable/keanu/network/ProtobufWriter.java
```diff
@@ -0,0 +1,101 @@
...
+        bayesNetBuilder = KeanuSavedBayesNet.BayesianNetwork.newBuilder();
+
+        net.save(this);
+
+        if (saveValues) {
+            net.saveValues(this);
+        }
+
+        bayesNetBuilder.build().writeTo(output);
+        bayesNetBuilder = null;
```
gordoncaleb: Why does the builder need to be stored as an instance variable if it is created and destroyed within this method?

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
```diff
@@ -7,30 +7,57 @@ option java_outer_classname = "KeanuSavedBayesNet";
...
+}
+
+message BooleanTensor {
+    repeated int64 shape = 1;
+    repeated bool values = 2;
+}
+
+message GenericTensor {
+    repeated int64 shape = 1;
+    repeated bytes values = 2;
```
gordoncaleb: Are you going to infer the number of bytes in the individual value from the shape?

keanu-project/src/main/java/io/improbable/keanu/vertices/ConstantVertexFactory.java
```diff
@@ -10,9 +10,9 @@
 import io.improbable.keanu.vertices.generic.nonprobabilistic.ConstantGenericVertex;
 import io.improbable.keanu.vertices.intgr.nonprobabilistic.ConstantIntegerVertex;
 
-public class ConstantVertex {
+public class ConstantVertexFactory {
 
-    private ConstantVertex() {
+    private ConstantVertexFactory() {
```
gordoncaleb: This is a factory, fair, but do we have to put that in its name? It's very Enterprise Java (J2EE) to go for MyClassFactoryFactory as a name. Is there a reason or is it just to highlight the specific pattern?

keanu-project/src/main/java/io/improbable/keanu/vertices/ConstantVertexFactory.java
```diff
@@ -74,4 +74,11 @@ public static ConstantDoubleVertex of(DoubleTensor value) {
     public static <TENSOR extends Tensor> ConstantGenericVertex<TENSOR> of(TENSOR tensor) {
         return new ConstantGenericVertex<>(tensor);
     }
+
+    public static boolean isConstantVertex(Vertex v) {
```
gordoncaleb: How about defining a tag interface and putting it on these classes and then using something like assignableFrom (https://stackoverflow.com/questions/3949260/java-class-isinstance-vs-class-isassignablefrom)

keanu-project/src/main/java/io/improbable/keanu/network/ProtobufWriter.java
christophernorth: Because there are lots of callbacks in to this object to add each Vertex/Value as you go along.  We'd otherwise need to pass that through to BayesNet/each vertex - removing the benefit of removing all the Protobuf related stuff from the Vertex object?

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
christophernorth: TBH: I needed to discuss this with you - I'm not sure how GenericTensors really work at the moment.  I was going to put their implementation in a separate PR though once we're happy with the overall approach.  Certainly this structure mirrors what is stored inside the generic Tensor itself though.

keanu-project/src/main/java/io/improbable/keanu/vertices/ConstantVertexFactory.java
christophernorth: Can we discuss this further tomorrow?  I've created an interface, but I couldn't see a good place in the hierarchy to pull together all the constants under a single class umbrella.

keanu-project/src/main/java/io/improbable/keanu/network/ProtobufReader.java
```diff
@@ -0,0 +1,107 @@
...
+        } catch (NoSuchMethodException e) {
+            throw new
+                IllegalArgumentException("Vertex Type doesn't support loading from Proto: " + vertex.getVertexType(), e);
+        }
+
+        Vertex<T> newVertex;
+        Map<String, Vertex> parentsMap = getParentsMap(vertex, existingVertices);
+
+        try {
+            newVertex = (Vertex<T>)vertexConstructor.newInstance(parentsMap, vertex.getConstantValue());
```
migwellian: if you parametrise `Class` and `Constructor` on `<Vertex<T>>` then you don't need this cast here.

keanu-project/src/main/java/io/improbable/keanu/vertices/ConstantVertexFactory.java
migwellian: as a matter of fact you didn't need to rename `ConstantVertex` at all. Just make it an interface - you're allowed static methods on an interface.

keanu-project/src/main/java/io/improbable/keanu/vertices/ConstantVertexFactory.java
migwellian: @gordoncaleb - isn't `ConstantVertex` the tag interface you're asking for?

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
migwellian: OK fair enough. I'd suggest using `NotImplementedException` rather than `UnsupportedOperationException` since it (arguably) implies that it will be implemented in the future. To me, `UnsupportedOperationException` is a code smell that suggests the need for a redesign, whereas `NotImplementedException` isn't.

(alternatively - just add the word "yet" to your error message!)

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/ConstantDoubleVertex.java
```diff
@@ -44,7 +65,16 @@ public DoubleTensor sample(KeanuRandom random) {
     }
 
     @Override
+    public Map<String, Vertex> getParentsMap() {
+        return new HashMap<>();
```
migwellian: This behaviour is the same as that in the method you're overriding (in `Vertex`).

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
```diff
@@ -335,11 +338,25 @@ public PartialDerivatives forwardModeAutoDifferentiation(Map<Vertex, PartialDeri
         }
     }
 
+    @Override
+    public void setValue(KeanuSavedBayesNet.VertexValue valueBuf) {
+        if (valueBuf.getValueTypeCase() != KeanuSavedBayesNet.VertexValue.ValueTypeCase.DOUBLEVAL) {
+            throw new IllegalArgumentException("Non Double Value specified for Double Vertex");
```
migwellian: This pair of lines appears twice. If you encapsulate it as a static method somewhere, you'll be able to reuse it many times (covering all the different `ValueTypeCase`s)

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/GaussianVertex.java
```diff
@@ -25,6 +28,8 @@
 
     private final DoubleVertex mu;
     private final DoubleVertex sigma;
+    private final static String MU_NAME = "mu";
+    private final static String SIGMA_NAME = "sigma";
```
migwellian: can you make use of the static final `ParameterName`s in `Diffs.java`? (We should move them to another class since this is nothing to do with diffs).

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
```diff
@@ -7,30 +7,57 @@ option java_outer_classname = "KeanuSavedBayesNet";
 
 message BayesianNetwork {
     repeated Vertex vertices = 1;
-    repeated VertexValue defaultState = 2;
+    repeated StoredValue defaultState = 2;
+}
+
+message VertexID {
+    string id = 1;
```
migwellian: how come you've changed from `repeated int64` to `string`?

keanu-project/src/test/java/io/improbable/keanu/network/BayesianNetworkTest.java
```diff
@@ -90,4 +97,17 @@ public void labelErrorsDetected() {
...
+    public void youCanSaveAndLoadANetworkWithValues() throws IOException {
+        DoubleVertex gaussianVertex = new GaussianVertex(0.0, 1.0);
+        BayesianNetwork net = new BayesianNetwork(gaussianVertex.getConnectedGraph());
+        ByteArrayOutputStream output = new ByteArrayOutputStream();
+
+        ProtobufWriter protobufWriter = new ProtobufWriter(net);
+        protobufWriter.save(output, true);
+        assertThat(output.size(), greaterThan(0));
+        ByteArrayInputStream input = new ByteArrayInputStream(output.toByteArray());
+        net = ProtobufReader.loadNetwork(input);
```
migwellian: can you assert something about the `net`?

keanu-project/src/main/java/io/improbable/keanu/vertices/ConstantVertexFactory.java
christophernorth: Done

keanu-project/src/main/java/io/improbable/keanu/vertices/ConstantVertexFactory.java
christophernorth: Done

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -204,4 +206,15 @@ public void incrementIndentation() {
...
     }
 
+    public void save(NetworkWriter protobufWriter) throws IOException {
+        for (Vertex vertex : TopologicalSort.sort(vertices)) {
+            vertex.save(protobufWriter);
+        }
+    }
+
+    public void saveValues(ProtobufWriter protobufWriter) throws IOException {
+        for (Vertex vertex : vertices) {
```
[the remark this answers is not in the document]
christophernorth: Generally because it doesn't matter - the values are just hooked up at a later point.

keanu-project/src/main/java/io/improbable/keanu/network/NetworkWriter.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+public interface NetworkWriter {
+
+    void save(OutputStream output, boolean saveValues) throws IOException;
+    void save(Vertex vertex);
+
+    default void save(ConstantDoubleVertex vertex) {
```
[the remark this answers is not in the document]
christophernorth: 👍

keanu-project/src/main/java/io/improbable/keanu/network/ProtobufReader.java
```diff
@@ -0,0 +1,107 @@
...
+
+            if (value.hasId()) {
+                targetVertex = instantiatedVertices.get(value.getId());
+            }
+
+            if (value.getVertexLabel() != "") {
+                Vertex newTarget = bayesNet.getVertexByLabel(new VertexLabel(value.getVertexLabel()));
+
+                if (targetVertex != null && newTarget != targetVertex) {
+                    throw new IllegalArgumentException("Label and VertexID don't refer to same Vertex");
```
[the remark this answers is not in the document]
christophernorth: Done

keanu-project/src/main/java/io/improbable/keanu/network/ProtobufReader.java
```diff
@@ -0,0 +1,107 @@
...
+                } else {
+                    targetVertex = newTarget;
+                }
+            }
+
+            if (targetVertex == null) {
+                throw new IllegalArgumentException("Value specified for unknown Vertex");
```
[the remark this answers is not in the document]
christophernorth: Done

keanu-project/src/main/java/io/improbable/keanu/network/ProtobufReader.java
christophernorth: The compiler really doesn't like that!  Complains about captures etc?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/ConstantDoubleVertex.java
christophernorth: Yeah - the parent version will become abstract once this is implemented everywhere so I don't want to rely on the base class implementation at the mo.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
christophernorth: Refactored in a slightly different way.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/GaussianVertex.java
christophernorth: Yeah, I'll do this as part of the more general change to move over every vertex.

keanu-project/src/main/proto/io/improbable/keanu/BayesianNetwork.proto
christophernorth: Let's chat - but basically IDs are actually pretty bad and meaningless once saved (ie we can't use them for loading anyway).  They're just there to hook things together, so I didn't want to impose our internal structure (ie the long array) on to the save format - other generators can then use what they like.

keanu-project/src/test/java/io/improbable/keanu/network/BayesianNetworkTest.java
christophernorth: Done now.

keanu-project/src/main/java/io/improbable/keanu/network/ProtobufReader.java
```diff
@@ -0,0 +1,236 @@
...
+
+    private final Map<Vertex, KeanuSavedBayesNet.VertexValue> savedValues;
+
+    public ProtobufReader() {
+        savedValues = new HashMap<>();
+    }
+
+    @Override
+    public void loadValue(Vertex vertex) {
+        throw new IllegalArgumentException("Cannot Load value for Untyped Vertex");
```
gordoncaleb: Can we ever load this? We could check for the vertex type first right?

keanu-project/src/main/java/io/improbable/keanu/network/NetworkReader.java
```diff
@@ -0,0 +1,14 @@
+package io.improbable.keanu.network;
+
+import io.improbable.keanu.vertices.Vertex;
+import io.improbable.keanu.vertices.bool.BoolVertex;
+import io.improbable.keanu.vertices.dbl.DoubleVertex;
+import io.improbable.keanu.vertices.intgr.IntegerVertex;
+
+public interface NetworkReader {
```
gordoncaleb: Given the interface methods, should it be NetworkLoader? Alternatively, the methods could be `readValue(...)`

keanu-project/src/main/java/io/improbable/keanu/network/NetworkWriter.java
```diff
@@ -0,0 +1,26 @@
...
+import io.improbable.keanu.vertices.Vertex;
+import io.improbable.keanu.vertices.bool.BoolVertex;
+import io.improbable.keanu.vertices.dbl.DoubleVertex;
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.ConstantDoubleVertex;
+import io.improbable.keanu.vertices.intgr.IntegerVertex;
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+public interface NetworkWriter {
```
gordoncaleb: NetworkSaver or methods -> `write(...)`

keanu-project/src/main/java/io/improbable/keanu/network/ProtobufReader.java
```diff
@@ -0,0 +1,236 @@
...
+                                   Map<KeanuSavedBayesNet.VertexID, Vertex> instantiatedVertices,
+                                   BayesianNetwork bayesNet) {
+        for (KeanuSavedBayesNet.StoredValue value : parsedNet.getDefaultStateList()) {
+            Vertex targetVertex = null;
+
+            if (value.hasId()) {
+                targetVertex = instantiatedVertices.get(value.getId());
+            }
+
+            if (value.getVertexLabel() != "") {
```
gordoncaleb: Either `!value.getVertexLabel().isEmpty()` or `value.getVertexLabel().equals("")`. Generally avoid != unless you're actually going for object equality. In Java's case Strings are interned, which is why this usually works.

christophernorth: Done

keanu-project/src/main/java/io/improbable/keanu/network/NetworkWriter.java
christophernorth: Can we fix this up once Ieva's stuff is out?  She's implementing the same interface so it would be better to clean once her stuff is out?

keanu-project/src/main/java/io/improbable/keanu/network/NetworkReader.java
christophernorth: As below

keanu-project/src/main/java/io/improbable/keanu/network/ProtobufReader.java
christophernorth: Yeah, we do check for other "loadable" types, and this is just the fallback and we wouldn't know what to do with it (we could just avoid saving anything, but that feels dirty).

keanu-project/src/main/java/io/improbable/keanu/network/ProtobufReader.java
```diff
@@ -0,0 +1,236 @@
...
+        }
+    }
+
+    private Constructor getAnnotatedConstructor(Class vertexClass) {
+        Constructor[] constructors = vertexClass.getConstructors();
+
+        for (Constructor constructor : constructors) {
+            Parameter[] parameters = constructor.getParameters();
+
+            if (parameters.length > 0 &&
```
gordoncaleb: Doesn't all of the parameters need to be annotated?

keanu-project/src/main/java/io/improbable/keanu/network/ProtobufReader.java
```diff
@@ -0,0 +1,236 @@
...
+        }
+    }
+
+    private void loadDefaultValues(KeanuSavedBayesNet.BayesianNetwork parsedNet,
+                                   Map<KeanuSavedBayesNet.VertexID, Vertex> instantiatedVertices,
+                                   BayesianNetwork bayesNet) {
+        for (KeanuSavedBayesNet.StoredValue value : parsedNet.getDefaultStateList()) {
+            Vertex targetVertex = null;
+
+            if (value.hasId()) {
```
gordoncaleb: Can some of these ifs be broken out into smaller methods?

keanu-project/src/main/java/io/improbable/keanu/vertices/ConstantVertex.java
```diff
@@ -74,4 +71,8 @@ public static ConstantDoubleVertex of(DoubleTensor value) {
     public static <TENSOR extends Tensor> ConstantGenericVertex<TENSOR> of(TENSOR tensor) {
         return new ConstantGenericVertex<>(tensor);
     }
+
+    public static boolean isConstantVertex(Vertex v) {
```
gordoncaleb: Is this needed?

keanu-project/src/test/java/io/improbable/keanu/network/ProtobufTest.java
```diff
@@ -0,0 +1,46 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.closeTo;
+import static org.hamcrest.Matchers.greaterThan;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.is;
+
+public class ProtobufTest {
+
+    @Test
+    public void youCanSaveAndLoadANetworkWithValues() throws IOException {
```
gordoncaleb: This is a good basic integration test but we could do a lot more here. Creating nested classes that represent problematic vertex annotations would be a good one. Also, making sure that the BayesianNetwork object can be saved using it's helper methods.

keanu-project/src/main/java/io/improbable/keanu/network/NetworkWriter.java
gordoncaleb: 👍

keanu-project/src/main/java/io/improbable/keanu/network/NetworkWriter.java
```diff
@@ -0,0 +1,26 @@
+package io.improbable.keanu.network;
+
+import io.improbable.keanu.vertices.ConstantVertex;
+import io.improbable.keanu.vertices.Vertex;
+import io.improbable.keanu.vertices.bool.BoolVertex;
+import io.improbable.keanu.vertices.dbl.DoubleVertex;
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.ConstantDoubleVertex;
+import io.improbable.keanu.vertices.intgr.IntegerVertex;
+
+import java.io.IOException;
```
gordoncaleb: Might want to give these files a quick auto format in Intellij too.

keanu-project/src/main/java/io/improbable/keanu/vertices/ConstantVertex.java
```diff
@@ -10,10 +10,7 @@
 import io.improbable.keanu.vertices.generic.nonprobabilistic.ConstantGenericVertex;
 import io.improbable.keanu.vertices.intgr.nonprobabilistic.ConstantIntegerVertex;
 
-public class ConstantVertex {
-
-    private ConstantVertex() {
-    }
+public interface ConstantVertex {
```
gordoncaleb: The methods no longer need to be explicitly called public now that it's an interface.

keanu-project/src/main/java/io/improbable/keanu/network/ProtobufReader.java
christophernorth: Yeah - I do verify that in my full test - I didn't know whether to redo that here though - all I wanted to do was find the one annotated constructor that we'll use to create the object.

keanu-project/src/main/java/io/improbable/keanu/vertices/ConstantVertex.java
christophernorth: No, I'll get rid.

keanu-project/src/main/java/io/improbable/keanu/vertices/ConstantVertex.java
christophernorth: Fixed

keanu-project/src/main/java/io/improbable/keanu/network/NetworkWriter.java
christophernorth: This doesn't seem to change if I autoformat in IntelliJ and Spotless doesn't touch it either?

keanu-project/src/test/java/io/improbable/keanu/network/ProtobufTest.java
christophernorth: Our real vertices provide some decent test cases already it turns out!  Would be good to chat briefly to see exactly what test cases you're looking for - stuff like negative test cases (like hand crafted negative Protobufs?)

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
christophernorth: Just for reference - this will get cleaned up once all vertices are saveable.

PR #320, opened 2018-10-31
Add parameter names to python constructors

icantsayrural:
Rename lambda to rate in exponential vertex, because lambda is a keyword in Python.


codegen/src/main/java/io/improbable/keanu/codegen/python/VertexProcessor.java
```diff
@@ -1,5 +1,6 @@
 package io.improbable.keanu.codegen.python;
 
+import com.google.common.base.CaseFormat;
```
migwellian: sweeet!

keanu-project/build.gradle
```diff
@@ -40,6 +40,10 @@ dependencies {
 
 }
 
+compileJava {
+    options.compilerArgs += ['-parameters']
```
migwellian: sweeeeeet!

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/ExponentialVertex.java
```diff
@@ -21,45 +21,45 @@
 
 public class ExponentialVertex extends DoubleVertex implements ProbabilisticDouble, SamplableWithManyScalars<DoubleTensor> {
 
-    private final DoubleVertex lambda;
+    private final DoubleVertex rate;
```
migwellian: why are you changing the name of this field?

keanu-python/keanu/generated/vertex.py
```diff
@@ -4,146 +4,146 @@
...
+java_import(context.jvm_view(), "io.improbable.keanu.vertices.dbl.probabilistic.UniformVertex")
+java_import(context.jvm_view(), "io.improbable.keanu.vertices.intgr.nonprobabilistic.ConstantIntegerVertex")
+java_import(context.jvm_view(), "io.improbable.keanu.vertices.intgr.nonprobabilistic.operators.binary.IntegerDivisionVertex")
+java_import(context.jvm_view(), "io.improbable.keanu.vertices.intgr.probabilistic.PoissonVertex")
+java_import(context.jvm_view(), "io.improbable.keanu.vertices.intgr.probabilistic.UniformIntVertex")
 
 
-def ConstantBool(*args) -> k.jvm_view().ConstantBoolVertex:
-    return Vertex(k.jvm_view().ConstantBoolVertex, args)
+def ConstantBool(constant) -> context.jvm_view().ConstantBoolVertex:
```
migwellian: do you still intend to add type annotations to the arguments?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/ExponentialVertex.java
icantsayrural: `lambda` is a keyword in Python so the easy fix was to change the parameter name in the constructor. I just changed the field name too for consistency

keanu-python/keanu/generated/vertex.py
icantsayrural: Yeah that's easy to do, maybe in another PR because it's a different jira task?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/ExponentialVertex.java
christophernorth: Given the domain Keanu operates in, are we likely to come across a few instances of "lambda"s as we move forward?  Do we have an obvious error message if the lambda thing pops up?  Perhaps we want to add one to the code generator if not?

icantsayrural: Yeah, that might be a good idea... The only reason I caught it was because one of the python tests used Exponential. Otherwise it could have gone unnoticed!

codegen/src/main/java/io/improbable/keanu/codegen/python/VertexProcessor.java
```diff
@@ -36,14 +38,18 @@ static void process(String generatedDir) {
...
         List<Import> imports = new ArrayList<>();
         List<PythonConstructor> pythonConstructors = new ArrayList<>();
+
         root.put("imports", imports);
         root.put("constructors", pythonConstructors);
 
         for (Constructor constructor : constructors) {
             String javaClass = constructor.getDeclaringClass().getSimpleName();
 
+            String[] pythonParameters = Arrays.stream(constructor.getParameters()).map(
```
gordoncaleb: Nice!

keanu-project/build.gradle
gordoncaleb: Note: this makes the build 20k bigger but I think it's worth it if anyone is using IntelliJ to decompile and view the source.

icantsayrural: @CharlieCrisp found a way to not include this parameter argument!

PR #324, opened 2018-11-01
Put vertex related modules into vertex and sampling, optimizer into algorithm

icantsayrural:
- `keanu.vertex` gives you all the generated vertices and `Const`.
- `keanu.algorithm` gives you the optimisers and `sample`.

I chose not to use these namespaces in our modules (i.e. explicitly use the directory namespace).

### DIFF
--- a/keanu-python/examples/coal_mining.py
+++ b/keanu-python/examples/coal_mining.py
@@ -1,7 +1,7 @@
 import pandas as pd
-import keanu as kn
 import numpy as np
-
+from keanu.vertex import UniformInt, Exponential, DoubleIf, Poisson
+from keanu import Model
 
 class CoalMining():
     __fname = "data/coal-mining-disaster-data.csv"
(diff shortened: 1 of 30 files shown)


keanu-python/examples/coal_mining.py
```diff
@@ -12,15 +12,15 @@ def __init__(self):
...
 
-            m.early_rate = kn.Exponential(1.0)
-            m.late_rate = kn.Exponential(1.0)
+            m.early_rate = Exponential(1.0)
+            m.late_rate = Exponential(1.0)
 
             m.years = np.array(self._data.index)
-            m.rates = kn.DoubleIf([1, 1], m.switchpoint > m.years, m.early_rate, m.late_rate)
-            m.disasters = kn.Poisson(m.rates)
+            m.rates = DoubleIf([1, 1], m.switchpoint > m.years, m.early_rate, m.late_rate)
```
christophernorth: Nit: can we make the [1, 1] in to a tuple? ie (1, 1).  Lists should ideally only be used for things you want to be mutable.

keanu-python/examples/coal_mining.py
```diff
@@ -1,7 +1,7 @@
 import pandas as pd
-import keanu as kn
 import numpy as np
-
+from keanu.vertex import UniformInt, Exponential, DoubleIf, Poisson
+from keanu import Model
 
 class CoalMining():
```
christophernorth: Nit: Do we get much value having this as a class rather than just a function? (I know this is nothing to do with the change at hand, but this feels like a leaking Java-ism? :)

icantsayrural: It's a curse

keanu-python/examples/coal_mining.py
icantsayrural: Good call.

keanu-python/examples/coal_mining.py
christophernorth: Sounds good.

PR #325, opened 2018-11-01
Take pandas dataframe, series and numpy scalars as argument for vertices AND tensors

icantsayrural:
I put all acceptable types in `vartypes.py`.

### DIFF
--- a/keanu-python/Pipfile
+++ b/keanu-python/Pipfile
@@ -7,10 +7,10 @@ name = "pypi"
 "py4j" = "*"
 numpy = "*"
 keanu = {editable = true, path = "."}
+pandas = "*"
 
 [dev-packages]
 pytest = "*"
-pandas = "*"
 
 [requires]
 python_version = "3.6"
--- a/keanu-python/keanu/vartypes.py
+++ b/keanu-python/keanu/vartypes.py
@@ -1,0 +1,14 @@
+import numpy as np
+import pandas as pd
+
+int_types = (int, np.integer)
+
+float_types = (float, np.float)
+
+bool_types = (bool, np.bool_)
+
+primitive_types = int_types + float_types + bool_types
+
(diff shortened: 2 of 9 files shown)


keanu-python/keanu/vartypes.py
```diff
@@ -0,0 +1,14 @@
...
+
+float_types = (float, np.float)
+
+bool_types = (bool, np.bool_)
+
+primitive_types = int_types + float_types + bool_types
+
+pandas_types = (pd.Series, pd.DataFrame)
+
+const_arg_types = (np.ndarray, ) + pandas_types + primitive_types
```
migwellian: for consistency it's probably worth calling out `numpy_types = (np.ndarray, )` explicitly. I expect in time it will contain more than one item anyway - for example there's a type called `np.mat`.

keanu-python/keanu/vertex/base.py
```diff
@@ -152,11 +152,11 @@ def __parse_args(args):
 
     @staticmethod
     def __parse_arg(arg):
-        if isinstance(arg, np.ndarray) or isinstance(arg, numbers.Number):
+        if isinstance(arg, const_arg_types):
```
migwellian: is `numbers.Number` the same as `(int, float, bool)`? I guess we don't need to talk about `numbers.Number` any more.

keanu-python/Pipfile
```diff
@@ -7,10 +7,10 @@ name = "pypi"
 "py4j" = "*"
 numpy = "*"
 keanu = {editable = true, path = "."}
+pandas = "*"
```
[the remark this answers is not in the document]
icantsayrural: Yeah, that should be pretty easy to do (don't quote me on this). I'll make a python m2 jira task for optional dependencies.

PR #326, opened 2018-11-02
Generate python constructor docs from JavaDoc strings

CharlieCrisp:
This PR introduces a feature to allow python constructors to pull in JavaDoc strings from their respective Java constructors. 
This is included as a gradle task that is run whenever `codegen:codegen` is run which is the existing way of generating python code.

What is not covered: 
 - Python docs in hand written code
 - Actually creation of python docs using e.g. Sphinx
 - Any sort of generated python code that isn't a constructor

### DIFF
--- a/codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
+++ b/codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
@@ -5,23 +5,31 @@ import java.util.Map;
 class DocString {
     private String comment;
     private Map<String, String> params;
-    private String methodName;
 
-    DocString(String comment, Map<String, String> params, String methodName) {
+    DocString(String comment, Map<String, String> params) {
         this.comment = comment;
         this.params = params;
-        this.methodName = methodName;
     }
 
-    String getComment() {
-        return comment;
+    private boolean isEmpty() {
+        return comment.isEmpty() && params.size() == 0;
     }
 
-    Map<String, String> getParams() {
-        return params;
-    }
-
-    String getMethodName() {
-        return methodName;
+    String getAsString() {
+        if (isEmpty()) {
+            return "";
+        }
+        StringBuilder stringBuilder = new StringBuilder();
+        stringBuilder.append("\"\"\"\n    ");
+        stringBuilder.append(comment.replaceAll("\n ", "\n    "));
+        for (String param : params.keySet()) {
+            stringBuilder.append("\n    ");
+            stringBuilder.append(":param ");
+            stringBuilder.append(param);
+            stringBuilder.append(": ");
+            stringBuilder.append(params.get(param));
+        }
+        stringBuilder.append("\n    \"\"\"\n    ");
+        return stringBuilder.toString();
     }
 }
--- a/codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
+++ b/codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
@@ -29,7 +29,7 @@ class DocString {
         stringBuilder.append(THREE_QUOTES);
         stringBuilder.append(NEW_LINE_TAB);
         stringBuilder.append(comment.replaceAll("\n ", NEW_LINE_TAB));
-        stringBuilder.append("\n");
+        stringBuilder.append(NEW_LINE_TAB);
         for (String param : params.keySet()) {
             stringBuilder.append(NEW_LINE_TAB);
             stringBuilder.append(":param ");
@@ -39,7 +39,6 @@ class DocString {
         }
         stringBuilder.append(NEW_LINE_TAB);
         stringBuilder.append(THREE_QUOTES);
-        stringBuilder.append("\n");
         return stringBuilder.toString();
     }
 }
--- a/codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
+++ b/codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
@@ -39,7 +39,7 @@ class DocString {
         }
         stringBuilder.append(NEW_LINE_TAB);
         stringBuilder.append(THREE_QUOTES);
-        stringBuilder.append(NEW_LINE_TAB);
+        stringBuilder.append("\n");
         return stringBuilder.toString();
     }
 }
--- a/codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
+++ b/codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
@@ -19,9 +19,13 @@ class DocString {
         if (isEmpty()) {
             return "";
         }
+        if (params.size() == 0) {
+            return "\"\"\"\n" + comment + "\n\"\"\"\n";
+        }
         StringBuilder stringBuilder = new StringBuilder();
         stringBuilder.append("\"\"\"\n    ");
         stringBuilder.append(comment.replaceAll("\n ", "\n    "));
+        stringBuilder.append("\n");
         for (String param : params.keySet()) {
             stringBuilder.append("\n    ");
             stringBuilder.append(":param ");
--- a/codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
+++ b/codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
@@ -1,0 +1,27 @@
+package io.improbable.keanu.codegen.python;
+
+import java.util.Map;
+
+class DocString {
+    private String comment;
+    private Map<String, String> params;
+    private String methodName;
+
+    DocString(String comment, Map<String, String> params, String methodName) {
+        this.comment = comment;
+        this.params = params;
+        this.methodName = methodName;
+    }
+
+    String getComment() {
+        return comment;
+    }
+
+    Map<String, String> getParams() {
+        return params;
+    }
+
+    String getMethodName() {
+        return methodName;
+    }
+}
--- a/codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
+++ b/codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
@@ -1,41 +1,43 @@
 package io.improbable.keanu.codegen.python;
 
+import org.apache.commons.lang3.StringUtils;
+
 import java.util.Map;
 
 class DocString {
     private static final String THREE_QUOTES = "\"\"\"";
     private static final String NEW_LINE_TAB = "\n    ";
 
-    private String comment;
-    private Map<String, String> params;
+    private final String methodDescription;
+    private final Map<String, String> parameterNameToDescriptionMap;
 
-    DocString(String comment, Map<String, String> params) {
-        this.comment = comment;
-        this.params = params;
+    DocString(String methodDescription, Map<String, String> parameterNameToDescriptionMap) {
+        this.methodDescription = methodDescription;
+        this.parameterNameToDescriptionMap = parameterNameToDescriptionMap;
     }
 
     private boolean isEmpty() {
-        return comment.isEmpty() && params.size() == 0;
+        return StringUtils.isEmpty(methodDescription) && parameterNameToDescriptionMap.isEmpty();
     }
 
     String getAsString() {
         if (isEmpty()) {
             return "";
         }
-        if (params.size() == 0) {
-            return THREE_QUOTES + "\n" + comment + THREE_QUOTES + "\n\n";
+        if (parameterNameToDescriptionMap.size() == 0) {
+            return THREE_QUOTES + "\n" + methodDescription + THREE_QUOTES + "\n\n";
         }
         StringBuilder stringBuilder = new StringBuilder();
         stringBuilder.append(THREE_QUOTES);
         stringBuilder.append(NEW_LINE_TAB);
-        stringBuilder.append(comment.replaceAll("\n ", NEW_LINE_TAB));
+        stringBuilder.append(methodDescription.replaceAll("\n ", NEW_LINE_TAB));
         stringBuilder.append(NEW_LINE_TAB);
-        for (String param : params.keySet()) {
+        for (String param : parameterNameToDescriptionMap.keySet()) {
             stringBuilder.append(NEW_LINE_TAB);
             stringBuilder.append(":param ");
             stringBuilder.append(param);
             stringBuilder.append(": ");
-            stringBuilder.append(params.get(param));
+            stringBuilder.append(parameterNameToDescriptionMap.get(param));
         }
         stringBuilder.append(NEW_LINE_TAB);
         stringBuilder.append(THREE_QUOTES);
--- a/codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
+++ b/codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
@@ -41,6 +41,7 @@ class DocString {
         }
         stringBuilder.append(NEW_LINE_TAB);
         stringBuilder.append(THREE_QUOTES);
+        stringBuilder.append(NEW_LINE_TAB);
         return stringBuilder.toString();
     }
 }
--- a/codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
+++ b/codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
@@ -3,6 +3,9 @@ package io.improbable.keanu.codegen.python;
 import java.util.Map;
 
 class DocString {
+    private static final String THREE_QUOTES = "\"\"\"";
+    private static final String NEW_LINE_TAB = "\n    ";
+
     private String comment;
     private Map<String, String> params;
 
@@ -20,20 +23,23 @@ class DocString {
             return "";
         }
         if (params.size() == 0) {
-            return "\"\"\"\n" + comment + "\n\"\"\"\n";
+            return THREE_QUOTES + "\n" + comment + THREE_QUOTES + "\n\n";
         }
         StringBuilder stringBuilder = new StringBuilder();
-        stringBuilder.append("\"\"\"\n    ");
-        stringBuilder.append(comment.replaceAll("\n ", "\n    "));
+        stringBuilder.append(THREE_QUOTES);
+        stringBuilder.append(NEW_LINE_TAB);
+        stringBuilder.append(comment.replaceAll("\n ", NEW_LINE_TAB));
         stringBuilder.append("\n");
         for (String param : params.keySet()) {
-            stringBuilder.append("\n    ");
+            stringBuilder.append(NEW_LINE_TAB);
             stringBuilder.append(":param ");
             stringBuilder.append(param);
             stringBuilder.append(": ");
             stringBuilder.append(params.get(param));
         }
-        stringBuilder.append("\n    \"\"\"\n    ");
+        stringBuilder.append(NEW_LINE_TAB);
+        stringBuilder.append(THREE_QUOTES);
+        stringBuilder.append(NEW_LINE_TAB);
         return stringBuilder.toString();
     }
 }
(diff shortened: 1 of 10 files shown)


codegen/src/main/java/io/improbable/keanu/codegen/python/KeanuProjectDoclet.java
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+
+    static Map<String, DocString> getDocStringsFromFile() {
+        try {
+            Type listType = new TypeToken<Map<String, DocString>>(){}.getType();
+            Gson gson = new Gson();
+            Reader reader = new FileReader(READ_DESTINATION + DESTINATION_FILE_NAME);
+            return gson.fromJson(reader, listType);
+        } catch (IOException e) {
```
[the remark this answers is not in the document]
CharlieCrisp: Unfortunately the interface that we need to implement in order to use the `javadoc` tool with this Doclet means we can't throw any errors. If you think we should be handling this error differently then I'm open to suggestions :)

codegen/src/main/java/io/improbable/keanu/codegen/python/KeanuProjectDoclet.java
```diff
@@ -0,0 +1,74 @@
...
+            Gson gson = new Gson();
+            String json = gson.toJson(docString);
+            File outputFile = new File(WRITE_DESTINATION + DESTINATION_FILE_NAME);
+            outputFile.getParentFile().mkdirs();
+            outputFile.createNewFile(); // if file already exists will do nothing
+            OutputStream outputStream = new FileOutputStream(WRITE_DESTINATION + DESTINATION_FILE_NAME, false);
+        System.out.println("Writing docstrings");
+            outputStream.write(json.getBytes());
+            System.out.println("Finished writing docstrings");
+        } catch (IOException e) {
```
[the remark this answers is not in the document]
CharlieCrisp: Unfortunately the interface that we need to implement in order to use the javadoc tool with this Doclet means we can't throw any errors. If you think we should be handling this error differently then I'm open to suggestions :)

migwellian:
Can you make the workflow automatic as part of `:codegen:build`?

codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
```diff
@@ -0,0 +1,39 @@
...
+
+    String getAsString() {
+        if (isEmpty()) {
+            return "";
+        }
+        if (params.size() == 0) {
+            return "\"\"\"\n" + comment + "\n\"\"\"\n";
+        }
+        StringBuilder stringBuilder = new StringBuilder();
+        stringBuilder.append("\"\"\"\n    ");
```
migwellian: likewise, you can make `"\n    "` a static final 

(that's supposed to say `\n` followed by 4 spaces but github has formatted it with one space)

codegen/src/main/java/io/improbable/keanu/codegen/python/KeanuProjectDoclet.java
```diff
@@ -0,0 +1,75 @@
+package io.improbable.keanu.codegen.python;
+
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
+import com.sun.javadoc.*;
```
migwellian: please configure your IDE to get rid of the `*` imports - I can show you how.

codegen/src/main/java/io/improbable/keanu/codegen/python/KeanuProjectDoclet.java
```diff
@@ -0,0 +1,75 @@
...
+                }
+            }
+        }
+        writeDocStringsToFile(docStrings);
+        return true;
+    }
+
+    private static boolean isConstructorAnnotated(ConstructorDoc constructorDoc) {
+        for (AnnotationDesc an: constructorDoc.annotations()) {
+            String exportVertexAnnotationName = "@" + ExportVertexToPythonBindings.class.getName();
```
migwellian: you could make this a static final

codegen/src/main/java/io/improbable/keanu/codegen/python/KeanuProjectDoclet.java
```diff
@@ -0,0 +1,75 @@
...
+                    }
+                }
+            }
+        }
+        writeDocStringsToFile(docStrings);
+        return true;
+    }
+
+    private static boolean isConstructorAnnotated(ConstructorDoc constructorDoc) {
+        for (AnnotationDesc an: constructorDoc.annotations()) {
```
migwellian: return Arrays.stream(constructorDoc.annotations())
            .anyMatch(an -> an.toString().equals(exportVertexAnnotationName));

codegen/src/main/java/io/improbable/keanu/codegen/python/KeanuProjectDoclet.java
```diff
@@ -0,0 +1,75 @@
...
+            if (an.toString().equals(exportVertexAnnotationName)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private static void writeDocStringsToFile(Map<String, DocString> docString) {
+        try {
+            Gson gson = new Gson();
```
migwellian: you can inline this variable

codegen/src/main/java/io/improbable/keanu/codegen/python/KeanuProjectDoclet.java
```diff
@@ -0,0 +1,75 @@
...
+            String json = gson.toJson(docString);
+            File outputFile = new File(WRITE_DESTINATION + DESTINATION_FILE_NAME);
+            outputFile.getParentFile().mkdirs();
+            outputFile.createNewFile(); // if file already exists will do nothing
+            OutputStream outputStream = new FileOutputStream(WRITE_DESTINATION + DESTINATION_FILE_NAME, false);
```
migwellian: use a try-with-resources here so that the stream gets closed

codegen/src/main/java/io/improbable/keanu/codegen/python/KeanuProjectDoclet.java
```diff
@@ -0,0 +1,75 @@
...
+            System.out.println("Writing docstrings");
```
migwellian: please use `slf4j` instead of `println` - I can show you how.

codegen/src/main/java/io/improbable/keanu/codegen/python/KeanuProjectDoclet.java
```diff
@@ -0,0 +1,75 @@
...
+            outputStream.write(json.getBytes());
+            System.out.println("Finished writing docstrings");
```
migwellian: do you actually need these `println`s? You could just `tail` your output file to see the progress.

codegen/src/main/java/io/improbable/keanu/codegen/python/KeanuProjectDoclet.java
```diff
@@ -0,0 +1,75 @@
...
+        } catch (IOException e) {
+            e.printStackTrace();
+            System.out.println("Could not write to file while processing JavaDoc strings");
```
migwellian: better to use slf4j: `log.error("message", e);`

migwellian: or would it be better in this scenario to wrap and throw `e`?

codegen/src/main/java/io/improbable/keanu/codegen/python/KeanuProjectDoclet.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    static Map<String, DocString> getDocStringsFromFile() throws IOException {
+        try {
+            Type listType = new TypeToken<Map<String, DocString>>(){}.getType();
+            Gson gson = new Gson();
+            Reader reader = new FileReader(READ_DESTINATION + DESTINATION_FILE_NAME);
+            return gson.fromJson(reader, listType);
+        } catch (IOException e) {
+            throw new IOException("Could not read JavaDoc strings from file at " + System.getProperty("user.dir") + READ_DESTINATION + DESTINATION_FILE_NAME);
```
migwellian:
```suggestion
            throw new IOException("Could not read JavaDoc strings from file at " + System.getProperty("user.dir") + READ_DESTINATION + DESTINATION_FILE_NAME, e);
```

codegen/src/main/java/io/improbable/keanu/codegen/python/ParamStringProcessor.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.Map;
+
+import static com.google.common.base.CaseFormat.LOWER_UNDERSCORE;
+import static com.google.common.base.CaseFormat.UPPER_CAMEL;
+
+class ParamStringProcessor {
+    static Map<String, String> getNameToCommentMapping(ConstructorDoc constructorDoc) {
+        String rawComment = constructorDoc.getRawCommentText();
+        String[] rawCommentLines = rawComment.split("\\r?\\n");
+        Map<String, String> nameToCommentMapping = new HashMap<>();
```
migwellian: better to use ImmutableMap.builder()

codegen/src/main/java/io/improbable/keanu/codegen/python/ParamStringProcessor.java
```diff
@@ -0,0 +1,28 @@
...
+        for (String commentLine : rawCommentLines) {
+            if (!commentLine.contains("@param")) {
```
migwellian: It looks like `constructorDoc.tags("@param")` gives you what you want in a cleaner way.

codegen/src/main/java/io/improbable/keanu/codegen/python/ParamStringProcessor.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.sun.javadoc.ConstructorDoc;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static com.google.common.base.CaseFormat.LOWER_UNDERSCORE;
+import static com.google.common.base.CaseFormat.UPPER_CAMEL;
+
+class ParamStringProcessor {
```
migwellian: I think this class would benefit from some tests - especially considering various edge cases around how the javadoc might be formatted.

codegen/src/main/java/io/improbable/keanu/codegen/python/KeanuProjectDoclet.java
CharlieCrisp: Unfortunately it would break the interface of `Standard` that we are implementing :(

codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
```diff
@@ -0,0 +1,45 @@
...
+        for (String param : params.keySet()) {
+            stringBuilder.append(NEW_LINE_TAB);
+            stringBuilder.append(":param ");
+            stringBuilder.append(param);
+            stringBuilder.append(": ");
+            stringBuilder.append(params.get(param));
+        }
+        stringBuilder.append(NEW_LINE_TAB);
+        stringBuilder.append(THREE_QUOTES);
+        stringBuilder.append(NEW_LINE_TAB);
```
migwellian: trailing whitespace?

```suggestion
        stringBuilder.append("\n");
```

codegen/src/main/java/io/improbable/keanu/codegen/python/KeanuProjectDoclet.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    private static void writeDocStringsToFile(Map<String, DocString> docString) {
+        String json = (new Gson()).toJson(docString);
+        createFilesIfNecessary();
+        try (OutputStream outputStream =
+                 new FileOutputStream(DESTINATION + DESTINATION_FILE_NAME, false)) {
+            outputStream.write(json.getBytes());
+        } catch (IOException e) {
+            Logger logger = LoggerFactory.getLogger(KeanuProjectDoclet.class);
```
migwellian: standard practice is to make `logger` a private static final member variable

codegen/src/main/java/io/improbable/keanu/codegen/python/KeanuProjectDoclet.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+
+    private static void createFilesIfNecessary() {
+        File outputFile = new File(DESTINATION + DESTINATION_FILE_NAME);
+        outputFile.getParentFile().mkdirs();
+        try {
+            outputFile.createNewFile();
+        } catch (IOException e) {
+            Logger logger = LoggerFactory.getLogger(KeanuProjectDoclet.class);
```
migwellian: see my comment above about static final

codegen/src/main/java/io/improbable/keanu/codegen/python/KeanuProjectDoclet.java
```diff
@@ -0,0 +1,87 @@
...
+            logger.error("Could not write to file while processing JavaDoc strings", e);
+        }
+    }
+
+    static Map<String, DocString> getDocStringsFromFile() throws IOException {
+        try {
+            Type listType = new TypeToken<Map<String, DocString>>(){}.getType();
+            Gson gson = new Gson();
```
migwellian: you can inline this `gson` too

codegen/src/test/java/io/improbable/keanu/codegen/python/ParamStringProcessorTest.java
```diff
@@ -0,0 +1,68 @@
...
+    private static final String SECOND_PARAM_TEXT = SECOND_PARAM_NAME_CAMEL_CASE + "   " + SECOND_PARAM_COMMENT;
+    private static final String RETURN_TAG_TEXT = "this";
+
+    private static final String PARAM_TAG_NAME = "@param";
+    private static final String RETURN_TAG_NAME = "@return";
+
+    @Mock ConstructorDoc constructorDoc;
+    @Mock Tag tag1;
+    @Mock Tag tag2;
+    @Mock Tag tag3;
```
migwellian: `tag3` is never used?

codegen/src/test/java/io/improbable/keanu/codegen/python/ParamStringProcessorTest.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Test
+    public void testParamStringMapping() {
+        Map<String, String> nameToCommentMap = getNameToCommentMapping(constructorDoc);
+        Set<String> keySet = nameToCommentMap.keySet();
+        assertThat(keySet, hasItems(FIRST_PARAM_NAME_SNAKE_CASE, SECOND_PARAM_NAME_SNAKE_CASE));
+        assertThat(nameToCommentMap.get(FIRST_PARAM_NAME_SNAKE_CASE), equalTo(FIRST_PARAM_COMMENT));
+        assertThat(nameToCommentMap.get(SECOND_PARAM_NAME_SNAKE_CASE), equalTo(SECOND_PARAM_COMMENT));
+        assertThat(keySet, hasSize(2));
```
migwellian: You can do this with `hasEntry`:

        assertThat(nameToCommentMap, hasEntry(equalTo(FIRST_PARAM_NAME_SNAKE_CASE), equalTo(FIRST_PARAM_COMMENT)));
        assertThat(nameToCommentMap, hasEntry(equalTo(SECOND_PARAM_NAME_SNAKE_CASE), equalTo(SECOND_PARAM_COMMENT)));

codegen/src/test/java/io/improbable/keanu/codegen/python/ParamStringProcessorTest.java
CharlieCrisp: Yeah so I guess I was trying to do use this indirectly by checking it doesn't cause another entry to appear in the map. Hence the line `AssertThat(keySet, hasSize(2));`. If you think this is not worth having in the test then I'm happy to remove

codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
```diff
@@ -0,0 +1,44 @@
+package io.improbable.keanu.codegen.python;
+
+import java.util.Map;
+
+class DocString {
+    private static final String THREE_QUOTES = "\"\"\"";
+    private static final String NEW_LINE_TAB = "\n    ";
+
+    private String comment;
+    private Map<String, String> params;
```
GeorgeNash: Make `comment` and `params` final

GeorgeNash: I don't really like the naming either - can we go for `description` or `documentation`? And what's `params`? Guessing it stores the name and the description - `parameterNameAndDescription`?

codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
```diff
@@ -0,0 +1,44 @@
...
+
+    DocString(String comment, Map<String, String> params) {
+        this.comment = comment;
+        this.params = params;
+    }
+
+    private boolean isEmpty() {
+        return comment.isEmpty() && params.size() == 0;
```
GeorgeNash: Use `StringUtils.isEmpty(comment)` instead

GeorgeNash: For the map use `isEmpty()`

codegen/src/main/java/io/improbable/keanu/codegen/python/ParamStringProcessor.java
```diff
@@ -0,0 +1,24 @@
...
+    static Map<String, String> getNameToCommentMapping(ConstructorDoc constructorDoc) {
+        ImmutableMap.Builder<String, String> nameToCommentMapping = ImmutableMap.builder();
+        Tag[] params = constructorDoc.tags("@param");
+        for (Tag param: params) {
+            String[] text = param.text().split(" ", 2);
+            String snakeCaseParamName = UPPER_CAMEL.to(LOWER_UNDERSCORE, text[0]);
+            String paramComment = text[1].trim();
```
GeorgeNash: Can you pull out the `text[1]` into a variable so it's a bit clearer? Is text[0] the name and text[1] the description?

keanu-python/keanu/vertex/generated.py
```diff
@@ -38,112 +38,209 @@
...
     return Vertex(context.jvm_view().ConstantDoubleVertex, constant)
 
 
 def DoubleIf(shape, predicate, thn, els) -> context.jvm_view().DoubleIfVertex:
+    
     return Vertex(context.jvm_view().DoubleIfVertex, shape, predicate, thn, els)
 
 
 def Addition(left, right) -> context.jvm_view().AdditionVertex:
+    """
```
GeorgeNash: Awesome!!

codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
CharlieCrisp: So the reason for using `comment` is because that is what it is called in the standard Doclet API but I agree that we can do better. Will update.

codegen/src/main/java/io/improbable/keanu/codegen/python/DocString.java
CharlieCrisp: This is actually important because of the way that the template is structured. Current template:
```
def ${constructor.pythonClass}(${constructor.pythonParameters}) -> context.jvm_view().${constructor.javaClass}:
    ${constructor.docString}return Vertex(context.jvm_view().${constructor.javaClass}, ${constructor.pythonParameters})
```
This means if `constructor.docString` is `""` then the return statement is on the line after the declaration i.e. without a line between. This consequently means we have to put in trailing whitespace for the return statement

PR #327, opened 2018-11-02
Feature/graph to dot file

ieva-improbable:
This PR adds DotWriter utility, which is an initial attempt at tooling for network visualisation. DotWriter allows you to output your network or a part of it into a [DOT](https://en.wikipedia.org/wiki/DOT_(graph_description_language)) file, which can then be used by various graph visualisation tools (e.g., [GraphViz](http://www.webgraphviz.com/)) to visualise the network.

Usage:
 * To output network to a DOT file: `WriteDot.outputDot(fileName, network)`
 * To output vertex and it's connections up to a degree n: `WriteDot.outputDot(fileName, vertex, n)`

To make the resulting visual easier to understand, vertex classes can be annotated with `DotAnnotation` which allows you to specify a display name for the class (e.g., "+" instead of "AdditionVertex") and whether there are any hyperparameters that should be visualised.

Note: currently hyperparameter handling isn't great - I manually check whether a vertex is a Gaussian vertex and which vertices represent its mu and sigma. I'm sure there is a better and more generic way to do that. Let me know if you think of anything :)


keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+     * Outputs the given graph to a DOT file which can be used by various graph visualisers to generate a visual representation of the graph.
+     * Read more about DOT format here: https://en.wikipedia.org/wiki/DOT_(graph_description_language)
+     *
+     * @param fileName name of the output file.
+     * @param net network to be written out to DOT format.
+     */
+    public static void outputDot(String fileName, BayesianNetwork net) {
+
+        // Get any vertex in the network and set the degree to infinity to print out the entire network.
+        if (net.getAllVertices().size() == 0) return;
```
christophernorth: My personal preference is to always avoid single line if statements - I find them a bit of a maintenance headache.

I'd much prefer the full:
if (stuff) {
    return;
}

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+     *
+     * @param fileName name of the output file
+     * @param vertex vertex around which the graph will be visualised
+     * @param degree degree of connections to be visualised; for instance, if the degree is 1,
+     *               only connections between the vertex amd it's parents and children will be written out to the DOT file.
+     */
+    public static void outputDot(String fileName, Vertex vertex, int degree) {
+        try {
+            // Reset variables
+            idsToLabelStrings = new HashMap<>();
```
christophernorth: I find using 'globals' to pass stuff around to be a bit confusing and error prone tbh.  Ideally I'd prefer a bit more of a functional style to this code - ie have your function pass back the completed, filled in, structures rather than filling in the globals.  Means that you don't have to deal with problems like these where we leave stale data around.

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+            verticeIdsToConnectionInfo = new HashMap<>();
+
+            File outputFile = getOutputFile(fileName);
+
+            writer = new BufferedWriter(new FileWriter(outputFile));
+            writer.write("digraph BayesianNetwork {\n");
```
christophernorth: I think the code here would be a bit more self describing if we used either:
a) A separate "DotOutputStream" kind of class (that would wrap our real output stream object and, for instance, stick this header on automatically).
b) Some "final" variables describing what they are (ie like public static final String DOT_HEADER = "digr...

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+
+            // Iterate over vertices and obtain the necessary label and conneciton info.
+            obtainGraphInfo(vertex, degree);
```
christophernorth: As above - I'm always in favour of avoiding functions with side effects where possible/sensible and I think it is here.

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+
+            // Write out labels and connections.
+            outputConnections(verticeIdsToConnectionInfo.values());
```
christophernorth: Similarly - I don't like using globals as "hidden parameters" where you can avoid it (ie, writer here).

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+        File outputFile = new File(fileName);
+
+        // Make sure the output directory exists.
+        if (outputFile.getParentFile() != null && !outputFile.getParentFile().exists()){
+            outputFile.getParentFile().mkdirs();
+        }
+
+        // Add an index to the filename if a file with the specified name already exists.
+        if (outputFile.exists()) {
+            String baseName = FilenameUtils.getBaseName(fileName);
```
christophernorth: Pulling this code out in to a separate function would make things a little cleaner IMO

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+            String extension = FilenameUtils.getExtension(fileName);
+            int counter = 1;
+            while(outputFile.exists())
```
christophernorth: nit: space between 'while' and '(' and { on same line.
```suggestion
            while (outputFile.exists()) {
```

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+            {
```
christophernorth:
```suggestion
```

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+                outputFile = new File(outputFile.getParent(), baseName + "_" + (counter++) + "." + extension);
```
christophernorth: Do we want to do 1000s of Loops if lots of these files exist?  Should we put an upper limit on it?

Also Nit: I'm not a fan of expressions with side effects within a single line of code (I think it's worth being explicit about every change you make to program state).

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+            }
+        }
+        return outputFile;
+    }
+
```
christophernorth:
```suggestion
```

christophernorth: We only have a single blank line between methods.

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+    /**
+     * Iterates over the vertices of the graph, storing information about vertex labels to display and connections between vertices.
+     *
+     * @param vertex vertex around which the graph will be visualised
+     * @param degree degree of connections to be visualised; for instance, if the degree is 1,
+     *               only connections between the vertex amd it's parents and children will be written out to the DOT file.
+     */
+    private static void obtainGraphInfo(Vertex vertex, int degree) {
+
+        HashSet<VertexId> processedVertices = new HashSet<>();
```
christophernorth: Do you need to explicitly store this as a HashSet?  We usually just store variables as their "most generic" type (to not expose unnecessary implementation details to other bits of the code - ie over the long term you end up with something more maintainable).

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+        HashSet<Vertex> verticesToProcessNow = new HashSet<>();
+        verticesToProcessNow.add(vertex);
+
+        // Iterate over the graph till the specified degree or till there are no more unprocessed vertices left.
+        int iterationIndex = 0;
+        while(iterationIndex < degree && !verticesToProcessNow.isEmpty()) {
```
christophernorth: would you be better with a for loop here?

christophernorth:
```suggestion
        while (iterationIndex < degree && !verticesToProcessNow.isEmpty()) {
```

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+            HashSet<Vertex> verticesToProcessNext = new HashSet<>();
+
+            for (Vertex v : verticesToProcessNow) {
+
+                // Use hashed IDs as the unique identifier for each vertex and store the other information in labels.
+                VertexId vId = v.getId();
+                setLabel(v);
+
+                // Iterate over all vertices connected to this one.
+                Stream<Vertex> connectedVertices = Stream.concat(v.getParents().stream(), v.getChildren().stream());
```
christophernorth: There are quite a few implementations of graph walkers in the codebase already - perhaps we should make a single implementation that all the code can share?

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+        }
+    }
+
+
+    // Utility function for creating a unique identifier for a connection between two vertices.
+    private static String concatenateVertexIds(VertexId v1, VertexId v2) {
+        if (v1.compareTo(v2) < 0) {
+            return v1.toString() + "_" + v2.toString();
+        }
+        else {
```
christophernorth: else on same line as '}' I think is our style

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+
+    /**
+     * Creates and stores a label for the given vertex.
+     * Label contains the information that will be displayed for this vertex, such as vertex class or display name and value.
+     *
+     * @param v a vertex
+     */
+    private static void setLabel(Vertex v){
```
christophernorth: This name could be confusing given it's shared with something on the Vertex class itself?

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+        int uniqueHiddenID = v.getId().hashCode();
+        String fullName = uniqueHiddenID + "[label=\"" + vertexLabel + "\"]";
+        idsToLabelStrings.put(uniqueHiddenID, fullName);
+    }
+
+
+    // Get value of constant vertices that always have their value set.
+    private static String getValueAsString(Vertex v) {
+        if (v.getClass().equals(ConstantDoubleVertex.class)) {
+            return "" + ((ConstantDoubleVertex) v).getValue().getValue(0);
```
christophernorth: I'm never a fan of having code that special cases types with no way to tie it back to the original (what if we add a new ConstantXVertex?).  Also, I have a similar problem with the GPB save/load stuff - we should probably chat to get a consistent solution?

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+
+
+    // Apply hyperparameters as a label to connection edges.
+    // Note - this is currently only done for Gaussian vertices. Can be extended for any others.
+    private static void applyHyperparameterInfo(Vertex v) {
+        VertexId vId = v.getId();
+
+        if (v.getClass().getSimpleName().equals(GaussianVertex.class.getSimpleName())) {
+            VertexId meanVertexId = ((GaussianVertex) v).getMu().getId();
+            verticeIdsToConnectionInfo.put(concatenateVertexIds(vId, meanVertexId), verticeIdsToConnectionInfo.get(concatenateVertexIds(vId, meanVertexId)) + " [label=mu]");
```
gordoncaleb: `verticeIdsToConnectionInfo.computeIfPresent(concatenateVertexIds(vId, meanVertexId), (ids, previousValue) -> previousValue + " [label=mu]");`

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+
+        int uniqueHiddenID = v.getId().hashCode();
+        String fullName = uniqueHiddenID + "[label=\"" + vertexLabel + "\"]";
+        idsToLabelStrings.put(uniqueHiddenID, fullName);
+    }
+
+
+    // Get value of constant vertices that always have their value set.
+    private static String getValueAsString(Vertex v) {
+        if (v.getClass().equals(ConstantDoubleVertex.class)) {
```
gordoncaleb: `v instanceof ConstantDoubleVertex`

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
gordoncaleb: We could have all the constant classes implement an interface and then we could check that v is assignable to that interface. That should make this more generalizable.

keanu-project/src/test/java/io/improbable/keanu/util/dot/WriteDotTest.java
```diff
@@ -0,0 +1,166 @@
...
+        double sigma = 1;
+        gaussianV = new GaussianVertex(mu, sigma);
+        doubleV = new ConstantDoubleVertex(5.43);
+        DoubleVertex result = gaussianV.multiply(doubleV);
+        gaussianNet = new BayesianNetwork(result.getConnectedGraph());
+        WriteDot.outputDot(gaussianDotFileName, gaussianNet);
+        gaussianDotOutput = new String(Files.readAllBytes(Paths.get(gaussianDotFileName)), Charset.defaultCharset());
+
+
+        ConstantBoolVertex boolV = new ConstantBoolVertex(true);
```
gordoncaleb: I'd go for some more descriptive names here. Like trueConstant...

keanu-project/src/test/java/io/improbable/keanu/util/dot/WriteDotTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+        WriteDot.outputDot(vertexDegree1DotFileName, gaussianV, 1);
+        vertexDegree1DotOutput = new String(Files.readAllBytes(Paths.get(vertexDegree1DotFileName)), Charset.defaultCharset());
+        WriteDot.outputDot(vertexDegree2DotFileName, gaussianV, 2);
+        vertexDegree2DotOutput = new String(Files.readAllBytes(Paths.get(vertexDegree2DotFileName)), Charset.defaultCharset());
+    }
+
+    @Test
+    public void filesGetWrittenAndAreNotOverwritten() throws IOException {
+        String fileName = "dottestfolder/test_file.dot";
```
gordoncaleb: If you want to test all the way down to the file system then you should use Java's create temp file (`File.createTempFile(...)`). You can find an example of it in `WriteCsvTest`

gordoncaleb: Even better would be to add the ability to write dot to an output stream that you could use to write to a test string instead of a file.

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -57,6 +52,13 @@ public Vertex getVertexByLabel(VertexLabel label) {
         return vertices;
     }
 
+    /**
+     * @return A list of all vertices in the network.
+     */
+    public List<Vertex> getAllVertices() {
+        return Collections.unmodifiableList(vertices);
```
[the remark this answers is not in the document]
ieva-improbable: What sort of method are you thinking of? And what would be the issue with getting a list of vertices like this?

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+
+
+    /**
+     * Outputs the given graph to a DOT file which can be used by various graph visualisers to generate a visual representation of the graph.
+     * Read more about DOT format here: https://en.wikipedia.org/wiki/DOT_(graph_description_language)
+     *
+     * @param fileName name of the output file.
+     * @param net network to be written out to DOT format.
+     */
+    public static void outputDot(String fileName, BayesianNetwork net) {
```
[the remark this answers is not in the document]
ieva-improbable: Good idea, will add that.

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -9,12 +9,7 @@
 import io.improbable.keanu.vertices.VertexLabel;
 import io.improbable.keanu.vertices.dbl.KeanuRandom;
 
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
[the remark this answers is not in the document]
migwellian: Actually `spotlessApply` doesn't cover this case, since it would only know how to detect it but not how to fix it.

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
ieva-improbable: Thanks Caleb, will use that. And agreed about the one line conditionals, they can be tricky.

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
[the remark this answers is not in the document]
migwellian: Agreed - in fact the only way you are using it at present is to "Get any vertex in the network" (WriteDot.java:58) - so is it actually needed?

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+ * Utility class for outputting a network to a DOT file that can then be used by a range of graph visualisers.
+ * Read more about DOT format here: https://en.wikipedia.org/wiki/DOT_(graph_description_language)
+ *
+ * Usage:
+ * To output network to a DOT file: WriteDot.outputDot(fileName, network)
+ * To output vertex and it's connections up to degree n: WriteDot.outputDot(fileName, vertex, n)
+ */
+public class WriteDot {
+
+    private static BufferedWriter writer;
```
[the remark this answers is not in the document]
migwellian: I favour option 2

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
migwellian: if you format the file in IntelliJ (Cmd-Shift-L) it will catch this and a bunch of other whitespace things.

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
ieva-improbable: Do we expect the user to have created and be storing thousands of such files? I'd be reluctant to put an upper limit to it, as I think it's never nice to overwrite a file without an explicit permission from the user.

Alternatively, we could timestamp them, but that looks a bit ugly. 

If you strongly think we should have an upper limit, I'll change it.

ieva-improbable: Re making it cleaner - would you prefer something like the following?

``` java
String newFileName = baseName + "_" + (counter++) + "." + extension;
outputFile = new File(outputFile.getParent(), newFileName);```

File: keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java (hunk 10 above)
Comment (ieva-improbable) [reply keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java]: ok

File: keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java (hunk 6 above)
Comment (ieva-improbable) [reply keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java]: Pulled it into a separate method, not sure it looks cleaner, but hopefully so :)

File: keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java (hunk 23 above)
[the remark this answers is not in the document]
Comment (ieva-improbable) [reply keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java]: Changed the imports, but would be good to fix the IntelliJ settings.

File: keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java (hunk 2 above)
Comment (ieva-improbable) [reply keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java]: Ok ,removed the globals.

File: keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java (hunk 3 above)
Comment (ieva-improbable) [reply keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java]: Makes sense, went for b)

File: keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java (hunk 12 above)
Comment (ieva-improbable) [reply keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java]: I'd still have to check whether I have any vertices left to process to avoid unnecessary iterations (as in, check for `!verticesToProcessNow.isEmpty()`). Or should I bring it in the loop?

File: keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java (hunk 13 above)
Comment (ieva-improbable) [reply keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java]: Can we chat about this in person? You guys definitely know the code base better than I do, but I couldn't find anything quite like what I needed.

File: keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java (hunk 14 above)
Comment (ieva-improbable) [reply keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java]: Oki

Comment (ieva-improbable) [reply keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java]: Is it fine to have it in new line if it is preceded by a comment? Or should I just be more sparing with my comments, as you guys don't seem to be big fans of them?

File: keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java (hunk 15 above)
Comment (ieva-improbable) [reply keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java]: It's a private method though, not intended to be used outside of this class. But I'll rename it to `setDotLabel`, as I guess it could be confusing if anyone wants to do more with this class.

File: keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java (hunk 16 above)
Comment (ieva-improbable) [reply keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java]: Yeah, this currently is really ugly, but I didn't know how to make it nicer. Let's chat offline.

File: keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java [hunk 25]
```diff
@@ -0,0 +1,279 @@
...
+    private static HashMap<Integer, String> idsToLabelStrings = new HashMap<>();
+    private static HashMap<String, String> verticeIdsToConnectionInfo = new HashMap<>();
+
+
+    /**
+     * Annotation used for vertex classes to specify how they should be exported to a DOT file.
+     */
+    @Retention(RetentionPolicy.RUNTIME)
+    public @interface DotAnnotation {
```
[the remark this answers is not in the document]
ieva-improbable: Good point, I'm an annotation noob. Will rename to `Display`.

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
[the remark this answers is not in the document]
ieva-improbable: Changed it to get rid of the global variables.

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+ * Read more about DOT format here: https://en.wikipedia.org/wiki/DOT_(graph_description_language)
+ *
+ * Usage:
+ * To output network to a DOT file: WriteDot.outputDot(fileName, network)
+ * To output vertex and it's connections up to degree n: WriteDot.outputDot(fileName, vertex, n)
+ */
+public class WriteDot {
+
+    private static BufferedWriter writer;
+    private static HashMap<Integer, String> idsToLabelStrings = new HashMap<>();
```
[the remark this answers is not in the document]
ieva-improbable: Oki

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
ieva-improbable: Oh, do you mean `Set<VertexId> processedVertices = new HashSet<>();`?

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+     * Outputs the given graph to a DOT file which can be used by various graph visualisers to generate a visual representation of the graph.
+     * Read more about DOT format here: https://en.wikipedia.org/wiki/DOT_(graph_description_language)
+     *
+     * @param fileName name of the output file
+     * @param vertex vertex around which the graph will be visualised
+     * @param degree degree of connections to be visualised; for instance, if the degree is 1,
+     *               only connections between the vertex amd it's parents and children will be written out to the DOT file.
+     */
+    public static void outputDot(String fileName, Vertex vertex, int degree) {
+        try {
```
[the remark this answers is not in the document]
ieva-improbable: Ok

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+
+        // If vertex is a constant vertex, display only it's value.
+        if ((vertexLabel = getValueAsString(v)).equals("")) {
+            // Else use vertex label as DOT label for it.
+            DotAnnotation vertexAnnotation = v.getClass().getAnnotation(DotAnnotation.class);
+            if (v.getLabel() != null) {
+                vertexLabel = v.getLabel().getUnqualifiedName();
+            }
+            // If label isn't set, and if DOT annotation is added to this class, use display name for label.
+            else if (vertexAnnotation != null && !vertexAnnotation.displayName().equals("")) {
```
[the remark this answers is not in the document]
ieva-improbable: Thanks!

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+        return "";
+    }
+
+
+    // Apply hyperparameters as a label to connection edges.
+    // Note - this is currently only done for Gaussian vertices. Can be extended for any others.
+    private static void applyHyperparameterInfo(Vertex v) {
+        VertexId vId = v.getId();
+
+        if (v.getClass().getSimpleName().equals(GaussianVertex.class.getSimpleName())) {
```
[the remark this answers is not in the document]
ieva-improbable: Much nicer :)

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
ieva-improbable: Oh, that's cool!

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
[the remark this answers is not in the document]
ieva-improbable: I need a way to traverse all the vertices in a network given a network. This can be done by having a vertex in a network, but my understanding is that currently we only have methods for retrieving latent and observed vertices, while it's possible that a network doesn't have either of these. 

What do you guys think is the best approach to achieving the above?

keanu-project/src/test/java/io/improbable/keanu/util/dot/WriteDotTest.java
```diff
@@ -0,0 +1,166 @@
...
+    private static String gaussianDotOutput;
+    private static String boolDotOutput;
+    private static String vertexDegree1DotOutput;
+    private static String vertexDegree2DotOutput;
+    private static String boolVertexLabel = "/BoolResult";
+    private static GaussianVertex gaussianV;
+    private static DoubleVertex doubleV;
+
+    @BeforeClass
+    public static void setup() throws IOException {
```
migwellian: It looks like you're constructing 3 independent graphs in this method, and their order of construction is not important. Can you split it into 3 separate `@Before` methods, with more descriptive method names?

keanu-project/src/test/java/io/improbable/keanu/util/dot/WriteDotTest.java
```diff
@@ -0,0 +1,166 @@
...
+    public void filesGetWrittenAndAreNotOverwritten() throws IOException {
+        String fileName = "dottestfolder/test_file.dot";
+        File outputFile = new File(fileName);
+        FileUtils.deleteDirectory(outputFile.getParentFile());
+
+        WriteDot.outputDot(fileName, gaussianNet);
+        assertTrue(outputFile.exists());
+
+        WriteDot.outputDot(fileName, booleanNet);
+        assertTrue(outputFile.getParentFile().listFiles().length == 2);
```
migwellian: you generally get more expressive error messages if you use hamcrest matchers, e.g.

        assertThat(outputFile.getParentFile().listFiles(), arrayWithSize(2));

keanu-project/src/test/java/io/improbable/keanu/util/dot/WriteDotTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+    @Test
+    public void outputContainsHyperparameters() throws IOException {
+        assertTrue(gaussianDotOutput.contains("[label=mu]"));
+        assertTrue(gaussianDotOutput.contains("[label=sigma]"));
+    }
+
+    @Test
+    public void outputContainsVertexLabels() {
+        assertTrue(boolDotOutput.contains(boolVertexLabel));
```
migwellian: You're testing a wide range of functionality, which is good, but some of them (like this one) are not very precise: you're checking that 11 chars exist anywhere in a string of almost 200 chars. It would be more natural to test your output against a pre-saved expected file which you can keep in `src/test/resources/`.

You should be able to check that the files are an exact match, as long as you reset `io.improbable.keanu.vertices.VertexId#ID_GENERATOR` to zero in a `@Before` method.

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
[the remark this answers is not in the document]
ieva-improbable: Makes sense, turns out we even have an annotation package, so I moved it there.

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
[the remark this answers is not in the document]
ieva-improbable: Very true, thanks for pointing it out, Changed this to a set.

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+                }
+
+                processedVertices.add(vId);
+            }
+
+            verticesToProcessNow = verticesToProcessNext;
+            iterationIndex++;
+        }
+
+        // Set labels for vertices that have not yet been processed.
```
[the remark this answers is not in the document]
ieva-improbable: Then I'd still have to separately set the label for the first vertex that I'm processing. Do you think that'd be better?

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+     * Label contains the information that will be displayed for this vertex, such as vertex class or display name and value.
+     *
+     * @param v a vertex
+     */
+    private static void setLabel(Vertex v){
+
+        String vertexLabel;
+
+        // If vertex is a constant vertex, display only it's value.
+        if ((vertexLabel = getValueAsString(v)).equals("")) {
```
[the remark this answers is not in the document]
ieva-improbable: Ok, pulled the assignment out.

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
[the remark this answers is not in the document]
ieva-improbable: I was under impression that Constant...Vertices store scalar values, but I might have misunderstood. If it's a non-scalar, we probably don't want to display it in a vertex in a graph, as it would get messy, so I should add a check for it. What do you think?

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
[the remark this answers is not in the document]
ieva-improbable: Yes, I'm aware of that. Any suggestions?

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+    // Output information about edges.
+    private static void outputConnections(Collection<String> connections) throws IOException {
+        for (String connection: connections) {
+            writer.write(connection + "\n");
+        }
+    }
+
+
+    // Output the stored labels.
+    private static void outputLabels() throws IOException {
```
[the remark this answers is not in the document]
ieva-improbable: Ok, merged them in one.

keanu-project/src/test/java/io/improbable/keanu/util/dot/WriteDotTest.java
```diff
@@ -0,0 +1,166 @@
...
+    private static String vertexDegree1DotOutput;
+    private static String vertexDegree2DotOutput;
+    private static String boolVertexLabel = "/BoolResult";
+    private static GaussianVertex gaussianV;
+    private static DoubleVertex doubleV;
+
+    @BeforeClass
+    public static void setup() throws IOException {
+
+        String gaussianDotFileName = testFileDir + "/gaussianDot.dot";
```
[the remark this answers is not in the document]
ieva-improbable: I am using Windows and it works for me.

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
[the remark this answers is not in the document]
ieva-improbable: Fixed them now.

keanu-project/src/test/java/io/improbable/keanu/util/dot/WriteDotTest.java
ieva-improbable: Oki, changed.

keanu-project/src/test/java/io/improbable/keanu/util/dot/WriteDotTest.java
```diff
@@ -0,0 +1,166 @@
...
+        gaussianNet = new BayesianNetwork(result.getConnectedGraph());
+        WriteDot.outputDot(gaussianDotFileName, gaussianNet);
+        gaussianDotOutput = new String(Files.readAllBytes(Paths.get(gaussianDotFileName)), Charset.defaultCharset());
+
+
+        ConstantBoolVertex boolV = new ConstantBoolVertex(true);
+        ConstantIntegerVertex intV = new ConstantIntegerVertex(5);
+        DoubleVertex doubleV2 = new ConstantDoubleVertex(5.43);
+        BoolVertex comparisonV = intV.greaterThan(doubleV2);
+        BoolVertex result2 = boolV.or(comparisonV);
```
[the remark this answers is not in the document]
ieva-improbable: Not entirely sure it's more readable from my perspective, but probably because I haven't worked with Keanu enough, Will change it to your suggestion.

keanu-project/src/test/java/io/improbable/keanu/util/dot/WriteDotTest.java
ieva-improbable: Ok, will use that.

keanu-project/src/test/java/io/improbable/keanu/util/dot/WriteDotTest.java
ieva-improbable: Is that the recommended way to unit test functionality with writing outputs? It seems really fragile to any changes. I agree that the current tests could be more exhaustive, but maybe I should improve them instead of going for full output comparison?

Probably better to discuss in person.

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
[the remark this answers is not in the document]
migwellian: I think that's a fair point.

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+ *
+ * Usage:
+ * To output network to a DOT file: WriteDot.outputDot(fileName, network)
+ * To output vertex and it's connections up to degree n: WriteDot.outputDot(fileName, vertex, n)
+ */
+public class WriteDot {
+
+    private static BufferedWriter writer;
+    private static HashMap<Integer, String> idsToLabelStrings = new HashMap<>();
+    private static HashMap<String, String> verticeIdsToConnectionInfo = new HashMap<>();
```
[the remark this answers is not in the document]
ieva-improbable: Done :)

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
```diff
@@ -0,0 +1,279 @@
...
+     * @param degree degree of connections to be visualised; for instance, if the degree is 1,
+     *               only connections between the vertex amd it's parents and children will be written out to the DOT file.
+     */
+    public static void outputDot(String fileName, Vertex vertex, int degree) {
+        try {
+            // Reset variables
+            idsToLabelStrings = new HashMap<>();
+            verticeIdsToConnectionInfo = new HashMap<>();
+
+            File outputFile = getOutputFile(fileName);
```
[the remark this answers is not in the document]
ieva-improbable: Also done.

keanu-project/src/main/java/io/improbable/keanu/util/dot/WriteDot.java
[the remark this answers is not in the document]
ieva-improbable: Now checking whether the value stored is a scalar and only displaying it if it is,

keanu-project/src/test/java/io/improbable/keanu/util/dot/WriteDotTest.java
ieva-improbable: Done

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -214,15 +222,49 @@ public void incrementIndentation() {
...
+        verticesToProcessNow.add(vertex);
+
+        int iterationIndex = 0;
+        while (iterationIndex <= degree && !verticesToProcessNow.isEmpty()) {
+
+            Set<Vertex> verticesToProcessNext = new HashSet<>();
+            subgraphVertices.addAll(verticesToProcessNow);
+
+            for (Vertex v : verticesToProcessNow) {
+                Stream<Vertex> connectedVertices = Stream.concat(v.getParents().stream(), v.getChildren().stream());
```
christophernorth: You don't need to create an actual variable for this I don't think - just call your stream processing methods directly on Stream.concat()

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -214,15 +222,49 @@ public void incrementIndentation() {
...
+                connectedVertices.forEach(connectedVertex -> {
+                    if (!subgraphVertices.contains(connectedVertex)) {
```
christophernorth: Streams have a .filter() method which would simplify the look of this code a bit (ie you wouldn't need the "if")

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
```diff
@@ -262,4 +266,30 @@ public void saveValue(NetworkWriter netWriter) {
...
+    /**
+     * @return a set of graph edges containing information about connected vertices.
+     */
+    public Set<GraphEdge> getParentEdgesInDotFormat() {
+        Set<GraphEdge> edges = new HashSet<>();
+        for (Vertex v : this.getParents()) {
+            edges.add(new GraphEdge(v, this));
+        }
+        return edges;
+    }
```
christophernorth: Can this code live in the DotWriter class?  This would probably make more sense given the layout as that keeps all of the Dot specific code in one place and away from the Vertex class (which is what was requested for the protobuf stuff - obviously the only exception to that is any necessary annotations).

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/ConstantDoubleVertex.java
```diff
@@ -53,6 +53,6 @@ public DoubleTensor calculate() {
 
     @Override
     public void save(NetworkWriter netWriter) {
-        netWriter.save(this);
+        netWriter.save((ConstantVertex)this);
```
christophernorth:
```suggestion
        netWriter.save(this);
```

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/GaussianVertex.java
```diff
@@ -135,4 +136,20 @@ public PartialDerivatives forwardModeAutoDifferentiation(Map<Vertex, PartialDeri
...
+        Set<GraphEdge> edges = super.getParentEdgesInDotFormat();
+
+        GraphEdge muEdge = new GraphEdge(this, mu);
+        edges.stream().filter(muEdge::equals).findFirst().get().appendToDotLabel("mu");
+
+        GraphEdge sigmaEdge = new GraphEdge(this, sigma);
+        edges.stream().filter(sigmaEdge::equals).findFirst().get().appendToDotLabel("sigma");
+
+        return edges;
+    }
```
christophernorth: You might be able to use the same annotations I added for the Protobuf stuff if you want parent names (will save you having to annotate all the vertices for a second time?).  We should also then be able to pull this code out in to the DotWriter stuff explicitly.  May require a bit of refactoring to pull out some of the code to read annotations, but I'm happy to help with that.

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -214,15 +222,49 @@ public void incrementIndentation() {
...
     }
+
+    /**
+     * Method for traversing a graph and returning a subgraph of vertices within the given degree of the specified vertex.
+     *
+     * @param vertex vertex that the subgraph will be centered around
+     * @param degree degree of connections from the vertex to be included in the subgraph
+     * @return a set of vertices within the specified degree from the given vertex
+     */
+    public Set<Vertex> getSubgraph(Vertex vertex, int degree) {
```
christophernorth: Would be good to write an explicit test for this functionality in the BayesianNetworkTest I think.

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -214,15 +222,49 @@ public void incrementIndentation() {
...
+
+        Set<Vertex> subgraphVertices = new HashSet<>();
+        Set<Vertex> verticesToProcessNow = new HashSet<>();
```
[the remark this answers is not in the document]
ieva-improbable: For `subgraphVertices` I believe hashset is better, as I'm doing finds against it (`if (!subgraphVertices.contains(connectedVertex))`). For `verticesToProcessNow` - if i have `verticesToProcessNext` as a hashset I would have to convert that to a list before iterating over it. Not sure about the complexity of this conversion, but I'm not convinced it would be faster. What did you have in mind?

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
ieva-improbable: Ok, changed it to do this.

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
ieva-improbable: Changed it to use filter.

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
ieva-improbable: Moved it to DotWriter now, I had to leave it in vertex previously to deal with hyperparameters (when they weren't annotated properly yet).

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/GaussianVertex.java
ieva-improbable: Done it now :)

keanu-project/src/main/java/io/improbable/keanu/network/NetworkWriter.java
```diff
@@ -14,10 +14,15 @@
 
     void save(OutputStream output, boolean saveValues) throws IOException;
     void save(Vertex vertex);
+
+    default void save(OutputStream output, Vertex vertex, int degree, boolean saveValues) throws IOException {
+        save(output, saveValues);
```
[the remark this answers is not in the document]
ieva-improbable: Yeah, true that. Don't actually know why I did it the way I did. Removed from the interface now.

keanu-project/src/main/java/io/improbable/keanu/util/dot/DotWriter.java
```diff
@@ -0,0 +1,156 @@
...
+    /**
+     * Outputs the network to a DOT file which can be used by various graph visualisers to generate a visual representation of the graph.
+     * Read more about DOT format here: https://en.wikipedia.org/wiki/DOT_(graph_description_language)
+     *
+     * @param output output stream to use for writing
+     * @param saveValues specify whether you want to output values of non-constant scalar vertices
+     */
+    @Override
+    public void save(OutputStream output, boolean saveValues) throws IOException {
+        // Get any vertex in the network and set the degree to infinity to print out the entire network.
```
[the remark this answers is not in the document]
ieva-improbable: Ok, changed it to what you suggested and got rid of the comment.

keanu-project/src/test/java/io/improbable/keanu/util/dot/DotWriterTest.java
```diff
@@ -0,0 +1,162 @@
...
+        if (output1Lines.size() != output2Lines.size()) {
+            return false;
+        }
+
+        // Don't care about line endings.
+        output1Lines = output1Lines.stream().map(line -> line.replace("\r", "")).collect(Collectors.toList());
+        output2Lines = output2Lines.stream().map(line -> line.replace("\r", "")).collect(Collectors.toList());
+
+        for (String line : output1Lines) {
+            if (!output2Lines.contains(line)) {
```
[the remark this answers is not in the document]
ieva-improbable: No, the visualisers that I looked at which take in DOT files don't seem to care about the order in which information is passed in.

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
```diff
@@ -262,4 +266,30 @@ public void saveValue(NetworkWriter netWriter) {
...
+        if (vertexAnnotation != null && !vertexAnnotation.displayName().isEmpty()) {
+            vertexDotLabel.setDotLabel(VertexDotLabel.VertexDotLabelType.ANNOTATION_LABEL, vertexAnnotation.displayName());
+        }
+        return vertexDotLabel;
+    }
+
+    /**
+     * @return a set of graph edges containing information about connected vertices.
+     */
+    public Set<GraphEdge> getParentEdgesInDotFormat() {
```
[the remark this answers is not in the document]
ieva-improbable: Ok, renamed.

keanu-project/src/main/java/io/improbable/keanu/util/dot/DotWriterUtils.java
```diff
@@ -0,0 +1,40 @@
+package io.improbable.keanu.util.dot;
+
+import org.apache.commons.io.FilenameUtils;
+
+import java.io.File;
+
+/**
+ * Utility class for creating a unique file in the given location. Used by Dot output writer.
+ */
+public class DotWriterUtils {
```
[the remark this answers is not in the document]
ieva-improbable: Left it in in case I ended up refactoring the DotWriter to be able to create files again, but I guess that won't happen.

keanu-project/src/main/java/io/improbable/keanu/util/dot/GraphEdge.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Class representing a graph edge.
+ */
+public class GraphEdge {
+
+    private Vertex parentVertex;
+    private Vertex childVertex;
+    private Set<String> dotLabels = new HashSet<>();
+    private static final String DOT_LABEL_OPENING = " [label=";
```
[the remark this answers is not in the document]
ieva-improbable: Ok

keanu-project/src/main/java/io/improbable/keanu/util/dot/GraphEdge.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public void appendToDotLabel(Set<String> dotLabel) {
+        dotLabels.addAll(dotLabel);
+    }
+
+    public void appendToDotLabel(String dotLabel) {
+        dotLabels.add(dotLabel);
+    }
+
+    public Set<String> getDotLabels() { return dotLabels;}
```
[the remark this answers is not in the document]
ieva-improbable: Ok

keanu-project/src/main/java/io/improbable/keanu/util/dot/VertexDotLabel.java
```diff
@@ -0,0 +1,66 @@
...
+        switch (labelType) {
+            case VALUE:
+                value = label;
+                break;
+            case VERTEX_LABEL:
+                vertexLabel = label;
+                break;
+            case ANNOTATION_LABEL:
+                annotation = label;
+        }
```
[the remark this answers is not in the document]
ieva-improbable: That's because it's never intended to be used to set class name (removed it from the enum), and it should not be possible to pass anything else than the options covered in. Should I add a default that throws an exception in that case?

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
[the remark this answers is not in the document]
christophernorth: I think if we just re-ordered how things are done - we could leave the "toProcess" things as Lists and keep the subgraph vertex as a set.  If we stick things in the subgraphVertices set before we process them we can simply check against that?

keanu-project/src/main/java/io/improbable/keanu/util/dot/VertexDotLabel.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public VertexDotLabel(Vertex vertex) {
+        this.vertex = vertex;
+    }
+
+    public enum VertexDotLabelType{
+        VALUE, CLASS_NAME, ANNOTATION_LABEL, VERTEX_LABEL
+    }
+
+    public void setDotLabel(VertexDotLabelType labelType, String label) {
```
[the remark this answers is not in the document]
ieva-improbable: I think I had intended to use it more, but ended up not doing it. I agree that it would be more readable to set value etc explicitly, but why is it simpler and safer?

keanu-project/src/test/java/io/improbable/keanu/util/dot/DotWriterTest.java
```diff
@@ -0,0 +1,162 @@
...
+    // Need to compare the outputs line by line, as the labels and edges are not written out in a fixed order.
+    private boolean dotFilesAreEqual(String output1, String output2) {
+        List<String> output1Lines = Arrays.asList(output1.split("\n"));
+        List<String> output2Lines = Arrays.asList(output2.split("\n"));
+
+        if (output1Lines.size() != output2Lines.size()) {
+            return false;
+        }
+
+        // Don't care about line endings.
```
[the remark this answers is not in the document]
ieva-improbable: True, changed the comment.

keanu-project/src/test/java/io/improbable/keanu/util/dot/DotWriterTest.java
```diff
@@ -0,0 +1,162 @@
...
+    private static final String COMPLEX_OUTPUT_FILENAME = resourcesFolder + "/ComplexNetDotOutput.dot";
+    private static final String VERTEX_DEGREE1__OUTPUT_FILENAME = resourcesFolder + "/VertexDegree1Output.dot";
+    private static final String VERTEX_DEGREE2__OUTPUT_FILENAME = resourcesFolder + "/VertexDegree2Output.dot";
+    private static final String OUTPUT_WITH_VALUES_FILENAME = resourcesFolder + "/OutputValuesSetToTrueOutput.dot";
+
+
+    @BeforeClass
+    public static void setUpComplexNet() {
+        complexResultVertex = (new ConstantBoolVertex(true)).or((new GaussianVertex(0, 1)).plus(new ConstantDoubleVertex(5)).equalTo(new ConstantDoubleVertex(10)));
+        complexNet = new BayesianNetwork(complexResultVertex.getConnectedGraph());
```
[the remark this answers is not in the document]
ieva-improbable: I don't think I understand what you mean by this, Having a class variable `DotWriter complexNetDotWriter` instead of a class variable `BayesianNetwork complexNet`?

keanu-project/src/test/java/io/improbable/keanu/util/dot/DotWriterTest.java
```diff
@@ -0,0 +1,162 @@
...
+        DotWriter dotWriter = new DotWriter(gaussianNet);
+        dotWriter.save(outputWriter, false);
+
+        String expectedGaussianNodeOutput = new String(Files.readAllBytes(Paths.get(GAUSSIAN_OUTPUT_FILENAME)), Charset.defaultCharset());
+        assertTrue(dotFilesAreEqual(outputWriter.toString(), expectedGaussianNodeOutput));
+    }
+
+    @Test
+    public void outputtingComplexNet() throws IOException {
+        ByteArrayOutputStream outputWriter = new ByteArrayOutputStream();
```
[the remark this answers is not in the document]
ieva-improbable: 👍

keanu-project/src/test/java/io/improbable/keanu/util/dot/DotWriterTest.java
```diff
@@ -0,0 +1,162 @@
...
+    public void outputtingVertexDegree2Surroundings() throws IOException {
+        ByteArrayOutputStream outputWriter = new ByteArrayOutputStream();
+        DotWriter dotWriter = new DotWriter(new BayesianNetwork(complexResultVertex.getConnectedGraph()));
+        dotWriter.save(outputWriter, complexResultVertex, 2, false);
+        String expectedVertexDegree2Output = new String(Files.readAllBytes(Paths.get(VERTEX_DEGREE2__OUTPUT_FILENAME)), Charset.defaultCharset());
+        assertTrue(dotFilesAreEqual(outputWriter.toString(), expectedVertexDegree2Output));
+    }
+
+    @Test
+    public void dotVertexLabelsAreSetCorrectly() throws IOException {
```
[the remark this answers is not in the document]
ieva-improbable: Not really, because it relies on the order in which the label and value are set for the specific vertex.

keanu-project/src/main/java/io/improbable/keanu/util/dot/GraphEdge.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public Vertex getParentVertex() { return parentVertex;}
+
+    public Vertex getChildVertex() { return childVertex;}
+
+    // Returns a string representing this edge in a DOT format.
+    public String inDotFormat() {
```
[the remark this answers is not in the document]
ieva-improbable: Ok

keanu-project/src/main/java/io/improbable/keanu/util/dot/DotWriter.java
```diff
@@ -0,0 +1,156 @@
...
+
+/**
+ * Utility class for outputting a network to a DOT file that can then be used by a range of graph visualisers.
+ * Read more about DOT format here: https://en.wikipedia.org/wiki/DOT_(graph_description_language)
+ *
+ * Usage:
+ * To output network to a DOT file: DotWriter.outputDot(fileName, network)
+ * To output vertex and it's connections up to degree n: DotWriter.outputDot(fileName, vertex, n)
+ */
+public class DotWriter implements NetworkWriter{
```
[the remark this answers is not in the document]
ieva-improbable: What kind of formatting? Tabs vs spaces?

keanu-project/src/main/java/io/improbable/keanu/util/dot/GraphEdge.java
```diff
@@ -0,0 +1,64 @@
...
+            dotOutput += DOT_LABEL_OPENING;
+            dotOutput += StringUtils.join(dotLabels, ", ");
+            dotOutput += DOT_LABEL_CLOSING;
+        }
+
+        return dotOutput;
+    }
+
+    public boolean equals(Object o) {
+        return o.hashCode() == this.hashCode();
```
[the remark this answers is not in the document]
ieva-improbable: Oh sweet, didn't know I could do that. All the magic of IDEs :)

keanu-project/src/main/java/io/improbable/keanu/util/dot/VertexDotLabel.java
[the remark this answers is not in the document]
ieva-improbable: (removed the method and enums)

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
ieva-improbable: Done :)

keanu-project/src/test/java/io/improbable/keanu/network/BayesianNetworkGetSubgraphTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+    @Test
+    public void aSubgraphOfDegreeOneContainsDegreeOneConnections() {
+        Set<Vertex> expectedDegree1Subgraph = ImmutableSet.of(v, parent, child);
+        Set<Vertex> degree1SubGraph = network.getSubgraph(v, 1);
+        assertThat(degree1SubGraph, containsInAnyOrder(expectedDegree1Subgraph.toArray()));
+    }
+
+    @Test
+    public void aSubgraphOfDegreeTwoContainsDegreeOneConnections() {
```
migwellian: Typo?

```suggestion
    public void aSubgraphOfDegreeTwoContainsDegreeOneAndTwoConnections() {
```

PR #329, opened 2018-11-05
Python streaming samples

GeorgeNash:
- Refactor Hamiltonian to use the `Sampler` interface. 
- Allows users to create a generator for reading samples from a java stream

Works by turning the java stream into an iterator and then calling `next` on that each time the user calls `next` on the python generator

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
@@ -103,22 +103,10 @@ public class Hamiltonian implements PosteriorSamplingAlgorithm {
 
         Map<VertexId, DoubleTensor> position = new HashMap<>();
         cachePosition(latentVertices, position);
-        Map<VertexId, DoubleTensor> positionBeforeLeapfrog = new HashMap<>();
-
         Map<VertexId, DoubleTensor> gradient = logProbGradientCalculator.getJointLogProbGradientWrtLatents();
-        Map<VertexId, DoubleTensor> gradientBeforeLeapfrog = new HashMap<>();
 
-        final Map<VertexId, DoubleTensor> momentum = new HashMap<>();
-        final Map<VertexId, DoubleTensor> momentumBeforeLeapfrog = new HashMap<>();
-
-        double logOfMasterPBeforeLeapfrog = bayesNet.getLogOfMasterP();
-        final List<Double> logOfMasterPForEachSample = new ArrayList<>();
-        logOfMasterPForEachSample.add(logOfMasterPBeforeLeapfrog);
-
-        final Map<VertexId, ?> sampleBeforeLeapfrog = new HashMap<>();
-
-        HamiltonianSampler.VertexState before = new HamiltonianSampler.VertexState(position, gradient, momentum);
-        HamiltonianSampler.VertexState after = new HamiltonianSampler.VertexState(positionBeforeLeapfrog, gradientBeforeLeapfrog, momentumBeforeLeapfrog);
+        HamiltonianSampler.VertexState before = new HamiltonianSampler.VertexState(position, gradient);
+        HamiltonianSampler.VertexState after = new HamiltonianSampler.VertexState(new HashMap<>(), new HashMap<>());
 
         return new HamiltonianSampler(
             latentVertices,
@@ -129,9 +117,7 @@ public class Hamiltonian implements PosteriorSamplingAlgorithm {
             bayesNet,
             before,
             after,
-            logProbGradientCalculator,
-            sampleBeforeLeapfrog,
-            logOfMasterPBeforeLeapfrog
+            logProbGradientCalculator
         );
     }
 
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
@@ -80,21 +80,20 @@ public class Hamiltonian implements PosteriorSamplingAlgorithm {
     }
 
     public NetworkSamples getPosteriorSamples(BayesianNetwork bayesianNetwork,
-                                              Vertex vertexToSampleFrom,
+                                              Vertex fromVertices,
                                               int sampleCount) {
-        return generatePosteriorSamples(bayesianNetwork, Collections.singletonList(vertexToSampleFrom))
+        return generatePosteriorSamples(bayesianNetwork, Collections.singletonList(fromVertices))
             .generate(sampleCount);
     }
 
     public NetworkSamplesGenerator generatePosteriorSamples(final BayesianNetwork bayesianNetwork,
-                                                            final List<? extends Vertex> verticesToSampleFrom) {
+                                                            final List<? extends Vertex> fromVertices) {
 
-        return new NetworkSamplesGenerator(setupSampler(bayesianNetwork, verticesToSampleFrom), ProgressBar::new);
+        return new NetworkSamplesGenerator(setupSampler(bayesianNetwork, fromVertices), ProgressBar::new);
     }
 
     private SamplingAlgorithm setupSampler(final BayesianNetwork bayesNet,
                                            final List<? extends Vertex> fromVertices) {
-
         bayesNet.cascadeObservations();
 
         final List<Vertex<DoubleTensor>> latentVertices = bayesNet.getContinuousLatentVertices();
@@ -119,7 +118,10 @@ public class Hamiltonian implements PosteriorSamplingAlgorithm {
 
         final Map<VertexId, ?> sampleBeforeLeapfrog = new HashMap<>();
 
-        return new Sampler(latentVertices, position, gradient, momentum, positionBeforeLeapfrog, gradientBeforeLeapfrog, momentumBeforeLeapfrog, random, fromVertices, leapFrogCount, stepSize, logProbGradientCalculator, sampleBeforeLeapfrog, bayesNet, logOfMasterPBeforeLeapfrog);
+        return new Sampler(latentVertices, position, gradient, momentum, positionBeforeLeapfrog,
+            gradientBeforeLeapfrog, momentumBeforeLeapfrog, random, fromVertices, leapFrogCount,
+            stepSize, logProbGradientCalculator, sampleBeforeLeapfrog, bayesNet, logOfMasterPBeforeLeapfrog
+        );
     }
 
     public static class Sampler implements SamplingAlgorithm {
@@ -159,7 +161,7 @@ public class Hamiltonian implements PosteriorSamplingAlgorithm {
             this.latentVertices = latentVertices;
             this.position = position;
             this.gradient = gradient;
-            this.momentum= momentum;
+            this.momentum = momentum;
             this.positionBeforeLeapfrog = positionBeforeLeapfrog;
             this.momentumBeforeLeapfrog = momentumBeforeLeapfrog;
             this.gradientBeforeLeapfrog = gradientBeforeLeapfrog;
@@ -198,18 +200,17 @@ public class Hamiltonian implements PosteriorSamplingAlgorithm {
         @Override
         public void sample(Map<VertexId, List<?>> samples, List<Double> logOfMasterPForEachSample) {
             step();
-            sampleLogic(samples);
+            addSampleFromCache(samples, acceptOrReject());
             logOfMasterPForEachSample.add(logOfMasterPBeforeLeapfrog);
         }
 
         @Override
         public NetworkState sample() {
             step();
-            sampleLogic(new HashMap<>());
-            return new SimpleNetworkState(takeSample(fromVertices));
+            return new SimpleNetworkState(acceptOrReject());
         }
 
-        private Map<VertexId, List<?>> sampleLogic(Map<VertexId, List<?>> samples) {
+        private Map<VertexId, ?> acceptOrReject() {
             final double logOfMasterPAfterLeapfrog = bayesNet.getLogOfMasterP();
 
             final double likelihoodOfLeapfrog = getLikelihoodOfLeapfrog(
@@ -230,21 +231,15 @@ public class Hamiltonian implements PosteriorSamplingAlgorithm {
                 gradient = gradientBeforeLeapfrog;
                 gradientBeforeLeapfrog = tempSwap;
 
-                addSampleFromCache(samples, sampleBeforeLeapfrog);
+                return sampleBeforeLeapfrog;
             } else {
-                addSampleFromVertices(samples, fromVertices);
                 logOfMasterPBeforeLeapfrog = logOfMasterPAfterLeapfrog;
+                Map<VertexId, ?> sample = new HashMap<>();
+                putSample(sample, fromVertices);
+                return sample;
             }
-            return samples;
         }
 
-        private static Map<VertexId, ?> takeSample(List<? extends Vertex> fromVertices) {
-            Map<VertexId, Object> sample = new HashMap<>();
-            for (Vertex v : fromVertices) {
-                sample.put(v.getId(), v.getValue());
-            }
-            return sample;
-        }
     }
 
     private static void cachePosition(List<Vertex<DoubleTensor>> latentVertices, Map<VertexId, DoubleTensor> position) {
@@ -348,6 +343,13 @@ public class Hamiltonian implements PosteriorSamplingAlgorithm {
         return dotProduct;
     }
 
+    /**
+     * This is meant to be used for caching a pre-leapfrog sample. This sample
+     * will be used if the leapfrog is rejected.
+     *
+     * @param sample
+     * @param fromVertices
+     */
     private static void putSample(Map<VertexId, ?> sample, List<? extends Vertex> fromVertices) {
         for (Vertex<?> vertex : fromVertices) {
             putValue(vertex, sample);
@@ -375,7 +377,6 @@ public class Hamiltonian implements PosteriorSamplingAlgorithm {
      * This is used when a leapfrog is accepted. At that point the vertices are in a
      * post leapfrog state.
      *
-     * @param samples
      * @param fromVertices vertices from which to create and save new sample.
      */
     private static void addSampleFromVertices(Map<VertexId, List<?>> samples, List<? extends Vertex> fromVertices) {
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
@@ -13,7 +13,6 @@ import lombok.Builder;
 import lombok.Getter;
 import lombok.Setter;
 
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
@@ -118,13 +118,13 @@ public class Hamiltonian implements PosteriorSamplingAlgorithm {
 
         final Map<VertexId, ?> sampleBeforeLeapfrog = new HashMap<>();
 
-        return new Sampler(latentVertices, position, gradient, momentum, positionBeforeLeapfrog,
+        return new HamiltonianSampler(latentVertices, position, gradient, momentum, positionBeforeLeapfrog,
             gradientBeforeLeapfrog, momentumBeforeLeapfrog, random, fromVertices, leapFrogCount,
             stepSize, logProbGradientCalculator, sampleBeforeLeapfrog, bayesNet, logOfMasterPBeforeLeapfrog
         );
     }
 
-    public static class Sampler implements SamplingAlgorithm {
+    public static class HamiltonianSampler implements SamplingAlgorithm {
 
         private List<Vertex<DoubleTensor>> latentVertices;
         private Map<VertexId, DoubleTensor> position;
@@ -142,7 +142,7 @@ public class Hamiltonian implements PosteriorSamplingAlgorithm {
         private BayesianNetwork bayesNet;
         private double logOfMasterPBeforeLeapfrog;
 
-        public Sampler(List<Vertex<DoubleTensor>> latentVertices,
+        public HamiltonianSampler(List<Vertex<DoubleTensor>> latentVertices,
                        Map<VertexId, DoubleTensor> position,
                        Map<VertexId, DoubleTensor> gradient,
                        Map<VertexId, DoubleTensor> momentum,
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
@@ -1,8 +1,5 @@
 package io.improbable.keanu.algorithms.mcmc;
 
-import static io.improbable.keanu.algorithms.mcmc.HamiltonianSampler.addSampleFromVertices;
-import static io.improbable.keanu.algorithms.mcmc.HamiltonianSampler.cachePosition;
-
 import io.improbable.keanu.algorithms.NetworkSamples;
 import io.improbable.keanu.algorithms.PosteriorSamplingAlgorithm;
 import io.improbable.keanu.network.BayesianNetwork;
@@ -21,6 +18,9 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+
+import static io.improbable.keanu.algorithms.mcmc.HamiltonianSampler.addSampleFromVertices;
+import static io.improbable.keanu.algorithms.mcmc.HamiltonianSampler.cachePosition;
 
 /**
  * Hamiltonian Monte Carlo is a method for obtaining samples from a probability
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
@@ -205,10 +205,11 @@ public class Hamiltonian implements PosteriorSamplingAlgorithm {
         @Override
         public NetworkState sample() {
             step();
-            return new SimpleNetworkState(sampleLogic(new HashMap<>()));
+            sampleLogic(new HashMap<>());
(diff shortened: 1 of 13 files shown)


keanu-python/keanu/algorithm/sampling.py
```diff
@@ -20,3 +20,17 @@ def sample(net, sample_from, algo='metropolis', draws=500, drop=0, down_sample_i
...
 
     return vertex_samples
+
+def iterate_samples(net, sample_from, algo='metropolis', draws=500, drop=0, down_sample_interval=1):
+	vertices_unwrapped = k.to_java_object_list(sample_from)
+
+	network_samples = algorithms[algo].withDefaultConfig().generatePosteriorSamples(net.unwrap(), vertices_unwrapped).dropCount(drop).downSampleInterval(down_sample_interval).stream().iterator()
+	
+	return iterator_samples(network_samples, draws, vertices_unwrapped)
+
```
christophernorth:
```suggestion
```

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -20,3 +20,17 @@ def sample(net, sample_from, algo='metropolis', draws=500, drop=0, down_sample_i
...
+
+def iterator_samples(network_samples, draws, vertices_unwrapped):
```
christophernorth: Do we need to limit the number of draws?  Given we're streaming, I don't see a need to put an upper bound on it - instead, just allow them to keep streaming until they're done (ie they could use itertools.islice() etc).

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -20,3 +20,17 @@ def sample(net, sample_from, algo='metropolis', draws=500, drop=0, down_sample_i
     vertex_samples = {Vertex._get_python_id(vertex_unwrapped): list(map(Tensor._to_ndarray, network_samples.get(vertex_unwrapped).asList())) for vertex_unwrapped in vertices_unwrapped}
 
     return vertex_samples
+
+def iterate_samples(net, sample_from, algo='metropolis', draws=500, drop=0, down_sample_interval=1):
+	vertices_unwrapped = k.to_java_object_list(sample_from)
+
+	network_samples = algorithms[algo].withDefaultConfig().generatePosteriorSamples(net.unwrap(), vertices_unwrapped).dropCount(drop).downSampleInterval(down_sample_interval).stream().iterator()
```
christophernorth: This is actually an "iterator", right?  As such, perhaps we can give it a slightly more descriptive name?  This isn't really a set of samples?

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -20,3 +20,17 @@ def sample(net, sample_from, algo='metropolis', draws=500, drop=0, down_sample_i
...
+	
+	return iterator_samples(network_samples, draws, vertices_unwrapped)
+
+
+def iterator_samples(network_samples, draws, vertices_unwrapped):
+	for i in range(draws):
```
christophernorth: Given you don't actually need the result you can just do:
for _ in range(draws):

This will just tell Python you're not bothered about the return value.

keanu-python/tests/test_sampling.py
```diff
@@ -53,3 +56,45 @@ def test_down_sample_interval(net):
...
+
+
+@pytest.mark.parametrize("algo", [
+    ("metropolis"),
+    ("hamiltonian")
+])
+def test_iter_is_correct_size(algo, net):
+    draws = 10
+    samples = kn.iterate_samples(net=net, sample_from=net.get_latent_vertices(), algo=algo, draws=draws, down_sample_interval=1)
+    assert sum(1 for i in samples) == 10
```
christophernorth: Similarly here - you don't actually need the sample, so using "_" will (marginally) speed up the code.

keanu-python/tests/test_sampling.py
```diff
@@ -53,3 +56,45 @@ def test_down_sample_interval(net):
...
+    set_starting_state(model)
+    samples = sample(net=net, sample_from=net.get_latent_vertices(), algo=algo, draws=draws)
+    set_starting_state(model)
+    iter_samples = iterate_samples(net=net, sample_from=net.get_latent_vertices(), algo=algo, draws=draws)
+    iter_samples_dict = {}
+
+    for sample in iter_samples:
+        for vertex_id, vertex_sample in sample.items():
+            if vertex_id in iter_samples_dict:    
+                iter_samples_dict[vertex_id] += vertex_sample
```
christophernorth: Perhaps consider a "defaultdict" for this.

keanu-python/keanu/algorithm/sampling.py
GeorgeNash: nice!!! thanks

keanu-python/tests/test_sampling.py
icantsayrural: For readability, I think we should just make it a list and check its length?

keanu-python/tests/test_sampling.py
```diff
@@ -53,3 +56,45 @@ def test_down_sample_interval(net):
...
+    for next_sample in iter_samples:
+        for vertex_id, vertex_sample in next_sample.items():
+            if vertex_id in iter_samples_dict:    
+                iter_samples_dict[vertex_id] += vertex_sample
+            else:
+                iter_samples_dict[vertex_id] = vertex_sample
+
+    for vertex_id, summed_vertex_sample in iter_samples_dict.items():
+        average = summed_vertex_sample / draws
+        np.testing.assert_almost_equal(average, np.average(samples[vertex_id]))
```
icantsayrural: 🎉

keanu-python/tests/test_sampling.py
```diff
@@ -53,3 +56,45 @@ def test_down_sample_interval(net):
...
+
+def set_starting_state(model):
+    KeanuRandom().set_default_random_seed(1)
```
icantsayrural: `KeanuRandom.set_default_random_seed(1)`. Also make sure to make that method static.

keanu-python/tests/test_sampling.py
christophernorth: I do wish they had a length function for this!  I think just wrapping the sum thing in a function would be better - this pretty much is the idiomatic way to do it after all...

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
```diff
@@ -96,15 +118,72 @@ public NetworkSamples getPosteriorSamples(final BayesianNetwork bayesNet,
...
+        private KeanuRandom random;
+        private List<? extends Vertex> fromVertices;
+        private int leapFrogCount;
+        private double stepSize;
+        private LogProbGradientCalculator logProbGradientCalculator;
+        private Map<VertexId, ?> sampleBeforeLeapfrog;
+        private BayesianNetwork bayesNet;
+        private double logOfMasterPBeforeLeapfrog;
+
+        public Sampler(List<Vertex<DoubleTensor>> latentVertices,
```
CharlieCrisp: Perhaps we should be using Lombok's `@AllArgsConstructor` for this class?

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
```diff
@@ -96,15 +118,72 @@ public NetworkSamples getPosteriorSamples(final BayesianNetwork bayesNet,
...
         final Map<VertexId, ?> sampleBeforeLeapfrog = new HashMap<>();
 
-        for (int sampleNum = 1; sampleNum < sampleCount; sampleNum++) {
+        return new Sampler(latentVertices, position, gradient, momentum, positionBeforeLeapfrog,
+            gradientBeforeLeapfrog, momentumBeforeLeapfrog, random, fromVertices, leapFrogCount,
+            stepSize, logProbGradientCalculator, sampleBeforeLeapfrog, bayesNet, logOfMasterPBeforeLeapfrog
+        );
+    }
+
+    public static class Sampler implements SamplingAlgorithm {
```
CharlieCrisp: Is there anything more descriptive that this can be called other than Sampler? It seems odd that the class name here is less descriptive than the interface it implements.

keanu-python/keanu/algorithm/sampling.py
GeorgeNash: Sorry I've changed this to a private method so it's clearer - this function creates the generator and is returned inside the `iterate_samples` function. I need the number of draws so I know how many times I'm able to call `.next()` on my Java iterator before I run out of samples!

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
```diff
@@ -96,15 +118,72 @@ public NetworkSamples getPosteriorSamples(final BayesianNetwork bayesNet,
...
+                       BayesianNetwork bayesNet,
+                       double logOfMasterPBeforeLeapfrog) {
+
+            this.latentVertices = latentVertices;
+            this.position = position;
+            this.gradient = gradient;
+            this.momentum = momentum;
+            this.positionBeforeLeapfrog = positionBeforeLeapfrog;
+            this.momentumBeforeLeapfrog = momentumBeforeLeapfrog;
+            this.gradientBeforeLeapfrog = gradientBeforeLeapfrog;
```
migwellian: it looks like (position, gradient, momentum) belong together - put them into a new class?

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
```diff
@@ -96,15 +118,72 @@ public NetworkSamples getPosteriorSamples(final BayesianNetwork bayesNet,
...
+        private Map<VertexId, DoubleTensor> gradientBeforeLeapfrog;
+        private Map<VertexId, DoubleTensor> momentumBeforeLeapfrog;
+        private KeanuRandom random;
+        private List<? extends Vertex> fromVertices;
+        private int leapFrogCount;
+        private double stepSize;
+        private LogProbGradientCalculator logProbGradientCalculator;
+        private Map<VertexId, ?> sampleBeforeLeapfrog;
+        private BayesianNetwork bayesNet;
+        private double logOfMasterPBeforeLeapfrog;
```
migwellian: some of these can be final

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
```diff
@@ -106,10 +106,10 @@ private SamplingAlgorithm setupSampler(final BayesianNetwork bayesianNetwork,
 
         double logProbabilityBeforeStep = bayesianNetwork.getLogOfMasterP();
 
-        return new Sampler(latentVertices, verticesToSampleFrom, mhStep, variableSelector, logProbabilityBeforeStep);
+        return new MetropolisHastingSampler(latentVertices, verticesToSampleFrom, mhStep, variableSelector, logProbabilityBeforeStep);
     }
 
-    public static class Sampler implements SamplingAlgorithm {
+    public static class MetropolisHastingSampler implements SamplingAlgorithm {
```
migwellian: nit: MetropolisHasting**s**Sampler

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/HamiltonianTest.java
```diff
@@ -112,4 +121,30 @@ public void canDefaultToSettingsInBuilderAndIsConfigurableAfterBuilding() {
...
+    @Test
+    public void canStreamSamples() {
+
+        Hamiltonian hmc = Hamiltonian.builder()
+            .leapFrogCount(10)
+            .stepSize(0.4)
+            .build();
+
+        int sampleCount = 7500;
+        int downSampleInterval = 1;
```
migwellian: you can inline this variable - naming it doesn't add any useful information

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/HamiltonianTest.java
```diff
@@ -112,4 +121,30 @@ public void canDefaultToSettingsInBuilderAndIsConfigurableAfterBuilding() {
...
+        double mu = 0;
+        double sigma = 1;
+        GaussianVertex A = new GaussianVertex(0, 1);
```
migwellian:
```suggestion
        GaussianVertex A = new GaussianVertex(mu, sigma);
```

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/HamiltonianTest.java
```diff
@@ -112,4 +121,30 @@ public void canDefaultToSettingsInBuilderAndIsConfigurableAfterBuilding() {
...
+
+    @Test
+    public void canStreamSamples() {
+
+        Hamiltonian hmc = Hamiltonian.builder()
+            .leapFrogCount(10)
+            .stepSize(0.4)
+            .build();
+
+        int sampleCount = 7500;
```
migwellian: It seems unnecessary to introduce another slow test for this:  can you make `sampleCount` small? You might need to change what you're asserting, but what's under test in this method is the ability to stream, not the accuracy of the sampling.

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -20,3 +20,16 @@ def sample(net, sample_from, algo='metropolis', draws=500, drop=0, down_sample_i
     vertex_samples = {Vertex._get_python_id(vertex_unwrapped): list(map(Tensor._to_ndarray, network_samples.get(vertex_unwrapped).asList())) for vertex_unwrapped in vertices_unwrapped}
 
     return vertex_samples
+
+def iterate_samples(net, sample_from, algo='metropolis', drop=0, down_sample_interval=1):
+	vertices_unwrapped = k.to_java_object_list(sample_from)
+
+	sample_iterator = algorithms[algo].withDefaultConfig().generatePosteriorSamples(net.unwrap(), vertices_unwrapped).dropCount(drop).downSampleInterval(down_sample_interval).stream().iterator()
```
migwellian: put this on >1 line?

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -20,3 +20,16 @@ def sample(net, sample_from, algo='metropolis', draws=500, drop=0, down_sample_i
...
+	
+	return _iterator_samples(sample_iterator, vertices_unwrapped)
+
+def _iterator_samples(sample_iterator, vertices_unwrapped):
+	while (True):
```
migwellian: What happens when you reach the end of the iterator? Don't you need to loop `while (sample_iterator.hasNext())`?

keanu-python/tests/test_sampling.py
```diff
@@ -53,3 +58,46 @@ def test_down_sample_interval(net):
 
     expected_num_samples = draws / down_sample_interval
     assert all(len(vertex_samples) == expected_num_samples for vertex_id, vertex_samples in samples.items())
+
+
+@pytest.mark.parametrize("algo", [
+    ("metropolis"),
+    ("hamiltonian")
+])
```
migwellian: nice

keanu-python/tests/test_sampling.py
```diff
@@ -53,3 +58,46 @@ def test_down_sample_interval(net):
...
+    draws = 100
+    model = Thermometer.model()
+    net = BayesNet(model.temperature.get_connected_graph())
+    set_starting_state(model)
+    samples = sample(net=net, sample_from=net.get_latent_vertices(), algo=algo, draws=draws)
+    set_starting_state(model)
+    iter_samples = iterate_samples(net=net, sample_from=net.get_latent_vertices(), algo=algo)
+    iter_samples_dict = defaultdict(float)
+
+    for next_sample in islice(iter_samples, draws):
```
migwellian: pandas will have some out-the-box functionality to get the average over a data frame of samples

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
```diff
@@ -68,10 +72,28 @@ public static Hamiltonian withDefaultConfig(KeanuRandom random) {
      * @return Samples taken with Hamiltonian Monte Carlo
      */
     @Override
-    public NetworkSamples getPosteriorSamples(final BayesianNetwork bayesNet,
-                                              final List<? extends Vertex> fromVertices,
-                                              final int sampleCount) {
+    public NetworkSamples getPosteriorSamples(BayesianNetwork bayesNet,
+                                              List<? extends Vertex> fromVertices,
+                                              int sampleCount) {
```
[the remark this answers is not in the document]
GeorgeNash: Oops - was a mistake! Have changed it back so it's consistent in this class at least.

keanu-python/keanu/algorithm/sampling.py
GeorgeNash: There's no end to the iterator - that's the beauty of it! It will keep sampling for as long as the user wants.

keanu-python/tests/test_sampling.py
GeorgeNash: any suggestions? It's iterating over data frames so I can't think of a better way than this

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -20,3 +20,18 @@ def sample(net, sample_from, algo='metropolis', draws=500, drop=0, down_sample_i
     vertex_samples = {Vertex._get_python_id(vertex_unwrapped): list(map(Tensor._to_ndarray, network_samples.get(vertex_unwrapped).asList())) for vertex_unwrapped in vertices_unwrapped}
 
     return vertex_samples
+
+def iterate_samples(net, sample_from, algo='metropolis', drop=0, down_sample_interval=1):
```
christophernorth:
```suggestion
def generate_samples(net, sample_from, algo='metropolis', drop=0, down_sample_interval=1):
```

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -20,3 +20,18 @@ def sample(net, sample_from, algo='metropolis', draws=500, drop=0, down_sample_i
...
+	vertices_unwrapped = k.to_java_object_list(sample_from)
+
+	sample_iterator = algorithms[algo].withDefaultConfig().generatePosteriorSamples(net.unwrap(), vertices_unwrapped)
+	sample_iterator = sample_iterator.dropCount(drop).downSampleInterval(down_sample_interval)
+	sample_iterator = sample_iterator.stream().iterator()
+	
+	return _iterator_samples(sample_iterator, vertices_unwrapped)
```
christophernorth:
```suggestion
	return _samples_generator(sample_iterator, vertices_unwrapped)
```

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/HamiltonianSampler.java
```diff
@@ -0,0 +1,280 @@
...
+    private static void cacheThings(VertexState from, VertexState to) {
+        for (Map.Entry<VertexId, DoubleTensor> entry : from.position.entrySet()) {
+            to.position.put(entry.getKey(), entry.getValue());
+        }
+        for (Map.Entry<VertexId, DoubleTensor> entry : from.momentum.entrySet()) {
+            to.momentum.put(entry.getKey(), entry.getValue());
+        }
+        for (Map.Entry<VertexId, DoubleTensor> entry : from.gradient.entrySet()) {
+            to.gradient.put(entry.getKey(), entry.getValue());
+        }
```
migwellian: to.position.putAll(from.position);
        to.momentum.putAll(from.momentum);
        to.gradient.putAll(from.gradient);

migwellian: This feels like something that should be encapsulated as a method in `VertexState`

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/HamiltonianSampler.java
```diff
@@ -0,0 +1,280 @@
...
+    private static boolean shouldReject(double likelihood, KeanuRandom random) {
+        return likelihood < random.nextDouble();
+    }
+
+    private static double dotProduct(Map<VertexId, DoubleTensor> momentums) {
+        double dotProduct = 0.0;
+        for (DoubleTensor momentum : momentums.values()) {
+            dotProduct += momentum.pow(2).sum();
+        }
+        return dotProduct;
```
migwellian: return momentums.values().stream()
            .mapToDouble(m -> m.pow(2).sum())
            .sum();

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/HamiltonianSampler.java
```diff
@@ -0,0 +1,280 @@
...
+     * This is meant to be used for caching a pre-leapfrog sample. This sample
+     * will be used if the leapfrog is rejected.
+     *
+     * @param sample
+     * @param fromVertices
+     */
+    private static void putSample(Map<VertexId, ?> sample, List<? extends Vertex> fromVertices) {
+        for (Vertex<?> vertex : fromVertices) {
+            putValue(vertex, sample);
+        }
```
migwellian: You can do this instead, although you also end up having to remove some `?`s elsewhere in this file to make the compiler happy.

    private static <T> Map<VertexId, Object> putSample(List<? extends Vertex> fromVertices) {
        return fromVertices.stream().collect(Collectors.toMap(Vertex::getId, Vertex::getValue));
    }

migwellian: It's arguably preferable to having that gnarly cast in `putValue`

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
```diff
@@ -96,204 +117,22 @@ public NetworkSamples getPosteriorSamples(final BayesianNetwork bayesNet,
...
-            momentums.put(currentVertex.getId(), random.nextGaussian(currentVertex.getShape()));
-        }
-    }
-
-    private static void cache(Map<VertexId, DoubleTensor> from, Map<VertexId, DoubleTensor> to) {
-        for (Map.Entry<VertexId, DoubleTensor> entry : from.entrySet()) {
-            to.put(entry.getKey(), entry.getValue());
-        }
+        HamiltonianSampler.VertexState before = new HamiltonianSampler.VertexState(position, gradient, momentum);
+        HamiltonianSampler.VertexState after = new HamiltonianSampler.VertexState(positionBeforeLeapfrog, gradientBeforeLeapfrog, momentumBeforeLeapfrog);
```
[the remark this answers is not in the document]
GeorgeNash: Nice!! Haven't bothered with the `static factory` as I think it's a bit overkill and probably clearer with the empty maps.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/HamiltonianSampler.java
GeorgeNash: Nice!

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/HamiltonianSampler.java
GeorgeNash: Nice - having the returned map looks much clearer too

keanu-python/tests/test_sampling.py
migwellian: Basically, dump all your vertex_samples into a dataframe and then call `mean()` on it.

	>>> import pandas as pd
	>>> df = pd.DataFrame([[1, 2], [10, 12]])
	>>> df.mean()
	0    5.5
	1    7.0
	dtype: float64

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/HamiltonianSampler.java
```diff
@@ -0,0 +1,263 @@
...
+        } else {
+            logOfMasterPBeforeLeapfrog = logOfMasterPAfterLeapfrog;
+            return putSample(fromVertices);
+        }
+    }
+
+    private static void initializeMomentumForEachVertex(List<Vertex<DoubleTensor>> vertexes,
+                                                        Map<VertexId, DoubleTensor> momentums,
+                                                        KeanuRandom random) {
+        for (Vertex<DoubleTensor> currentVertex : vertexes) {
```
CharlieCrisp: nit: `vertices`

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/HamiltonianSampler.java
```diff
@@ -0,0 +1,263 @@
...
+                                                        final double stepSize,
+                                                        final LogProbGradientCalculator logProbGradient) {
+
+        final double halfTimeStep = stepSize / 2.0;
+
+        Map<VertexId, DoubleTensor> momentumsAtHalfTimeStep = new HashMap<>();
+
+        //Set `r = r + (eps/2)dTL(T)
+        for (Map.Entry<VertexId, DoubleTensor> vertexMomentum : during.momentum.entrySet()) {
+            final DoubleTensor updatedMomentum = during.gradient.get(vertexMomentum.getKey()).times(halfTimeStep).plusInPlace(vertexMomentum.getValue());
```
CharlieCrisp: I think we could probably make this much more readable by extracting out some variables...

CharlieCrisp: Actually this is true for all of the for loops here

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Hamiltonian.java
```diff
@@ -71,7 +74,24 @@ public static Hamiltonian withDefaultConfig(KeanuRandom random) {
...
+            .generate(sampleCount);
+    }
+
+    public NetworkSamples getPosteriorSamples(final BayesianNetwork bayesNet,
+                                              final Vertex fromVertex,
+                                              final int sampleCount) {
+        return getPosteriorSamples(bayesNet, Collections.singletonList(fromVertex), sampleCount);
+    }
+
+    public NetworkSamplesGenerator generatePosteriorSamples(final BayesianNetwork bayesNet,
```
CharlieCrisp: Maybe we should be calling this something like `getPosteriorSamplesGenerator`. I don't think I like the `generatePosteriorSamples().generate()` chain

PR #333, opened 2018-11-07
Feature/linear regression with sampling

migwellian:
This PR introduces a new class, `SamplingModelFitter`, and the option of using it in linear regression.


keanu-project/src/main/java/io/improbable/keanu/model/SamplingModelFitter.java
```diff
@@ -0,0 +1,41 @@
+package io.improbable.keanu.model;
+
+import io.improbable.keanu.algorithms.NetworkSamples;
+import io.improbable.keanu.algorithms.PosteriorSamplingAlgorithm;
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+import io.improbable.keanu.vertices.Vertex;
+
+public class SamplingModelFitter<INPUT, OUTPUT> implements ModelFitter<INPUT, OUTPUT> {
```
CharlieCrisp: May be worth giving a short description of how this works and maybe how it is different to the ML/MAP estimation fitters?

keanu-project/src/main/java/io/improbable/keanu/model/SamplingModelFitter.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public SamplingModelFitter(ModelGraph<INPUT, OUTPUT> modelGraph, PosteriorSamplingAlgorithm samplingAlgorithm, int sampleCount, int dropCount) {
+        this.modelGraph = modelGraph;
+        this.samplingAlgorithm = samplingAlgorithm;
+        this.sampleCount = sampleCount;
+        this.dropCount = dropCount;
+    }
+
+    /**
+     * Uses the Metropolis Hastings sampling algorithm to fit the model graph to a given set of input and output data.
```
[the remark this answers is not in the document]
migwellian: 👍

keanu-project/src/main/java/io/improbable/keanu/model/SamplingModelFitter.java
```diff
@@ -0,0 +1,52 @@
...
+    private final ModelGraph<INPUT, OUTPUT> modelGraph;
+    private final PosteriorSamplingAlgorithm samplingAlgorithm;
+    private final int sampleCount;
+    private final int dropCount;
+
+    /**
+     * This fitter uses a {@link PosteriorSamplingAlgorithm}, in contrast to the {@link MAPModelFitter} and {@link MaximumLikelihoodModelFitter}, which use gradient methods.
+     *
+     * The model's latent vertices will have their values set to the average over the samples.
+     * A down-sample of 2 is used (not configurable).
```
CharlieCrisp: Just curious - why is is we use a down-sampling rate of 2?

keanu-project/src/main/java/io/improbable/keanu/model/SamplingModelFitter.java
CharlieCrisp: Javadoc is great - my only concern now is that because this will often be used via the `RegressionModelBuilder.withSampling()` method that we should maybe put some documentation on that too?

keanu-project/src/main/java/io/improbable/keanu/model/SamplingModelFitter.java
migwellian: Down-sampling is to avoid cross-correlation in the samples. I simply chose 2 because it's what some of the other tests do. 

We could make this configurable but I didn't want to clutter the API. Any thoughts @gordoncaleb ? I'm also planning to discuss it with @johannespetrat when he's got some time next week.

keanu-project/src/main/java/io/improbable/keanu/model/SamplingModelFitter.java
migwellian: good call - done.

keanu-project/src/main/java/io/improbable/keanu/model/SamplingModelFitter.java
```diff
@@ -0,0 +1,52 @@
...
+     * This will mutate the graph which can then be used to construct a graph-backed model like, for instance, a
+     * {@link io.improbable.keanu.model.regression.RegressionModel RegressionModel}
+     *
+     * @param input  The input data to your model graph
+     * @param output The output data to your model graph
+     */
+    @Override
+    public void fit(INPUT input, OUTPUT output) {
+        modelGraph.observeValues(input, output);
+        NetworkSamples posteriorSamples = samplingAlgorithm
```
gordoncaleb: The reason you would want to use sampling for something like Linear Regression instead of gradient optimisation would be so that you can preserve the uncertainty around the weight values. There's more information in these samples than just the average (e.g. std deviation). The average doesn't even always represent MAP as I think we are using it here as. For example, if the posterior was a the donut shape then the average would be the middle of the donut, which would probably be a very low probability point. If you want to use sampling for linear regression then it would be to use samples to create some type of empirical distribution (e.g. GaussianKDE) that would be used as the weight vertex.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
```diff
@@ -29,6 +31,7 @@
  */
 @Builder
 public class MetropolisHastings implements PosteriorSamplingAlgorithm {
+    private static final Logger log = LoggerFactory.getLogger(MetropolisHastings.class);
```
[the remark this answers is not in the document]
migwellian: I was not aware of that!

keanu-project/src/main/java/io/improbable/keanu/model/SamplingModelFitter.java
migwellian: This comment is now obsolete - I will remove it. I'm not doing any downsampling inside the fitter any more.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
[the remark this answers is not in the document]
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastingsStep.java
```diff
@@ -18,6 +20,8 @@
 
 class MetropolisHastingsStep {
 
+    private static final Logger log = LoggerFactory.getLogger(MetropolisHastingsStep.class);
```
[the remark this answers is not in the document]
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/ScalarDoubleTensor.java
```diff
@@ -860,10 +860,7 @@ public int hashCode() {
 
     @Override
     public String toString() {
-        return "{\n" +
-            "data = [" + value + "]" +
-            "\nshape = " + Arrays.toString(shape) +
-            "\n}";
+        return String.format("[[%.4f]]", value);
```
GeorgeNash: Why have we taken the shape out of this?

keanu-project/src/main/java/io/improbable/keanu/model/SamplingModelFitter.java
migwellian: this is obsolete now - don't do the down-sampling as part of the fitting any more

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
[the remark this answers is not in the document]
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/tensor/intgr/ScalarIntegerTensor.java
```diff
@@ -519,9 +519,7 @@ public BooleanTensor elementwiseEquals(Integer value) {
 
     @Override
     public String toString() {
-        return "ScalarIntegerTensor{" +
-            "value=" + value +
-            '}';
+        return String.format("[[%d]]", value);
```
gordoncaleb: The number of [s will depend on the rank of the scalar. 1x1x1 would be [[[value]]].

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/ScalarDoubleTensor.java
gordoncaleb: The number of [s depends on the rank.

PR #334, opened 2018-11-08
Make Differentiable interface reliable

ieva-improbable:
Remove `Differentiable` interface from the `DoubleVertex` and make vertices that are differentiable implement Differentiable directly. Otherwise all vertices extending `DoubleVertex` seem to be differentiable, which is misleading, as `RoundVertex`, `CeilingVertex`, `FloorVertex` etc are not differentiable. The default implementation of `forwardModeAutoDifferentiation` and `reverseModeAutoDifferentiation` has been moved form `DoubleVertex` to `Differentiable` as a default method. 

Note: I'm not 100% taking the implementation of `forwardModeAutoDifferentiation` out of `DoubleUnaryOpVertex` is ok.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/CastDoubleVertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/CastDoubleVertex.java
@@ -12,7 +12,7 @@ import io.improbable.keanu.vertices.dbl.nonprobabilistic.diff.PartialDerivatives
 
 import java.util.Map;
 
-public class CastDoubleVertex extends DoubleVertex  implements Differentiable, NonProbabilistic<DoubleTensor> {
+public class CastDoubleVertex extends DoubleVertex  implements NonProbabilistic<DoubleTensor> {
 
     private final Vertex<? extends NumberTensor> inputVertex;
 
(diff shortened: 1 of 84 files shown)


ieva-improbable:
Nope, a few of the operator vertices aren't (like ceiling, floor,
rounding). Will add this to the description.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/binary/DoubleBinaryOpVertex.java
```diff
@@ -67,7 +68,7 @@ public PartialDerivatives forwardModeAutoDifferentiation(Map<Vertex, PartialDeri
         try {
             return forwardModeAutoDifferentiation(derivativeOfParentsWithRespectToInputs.get(left), derivativeOfParentsWithRespectToInputs.get(right));
         } catch (UnsupportedOperationException e) {
-            return super.forwardModeAutoDifferentiation(derivativeOfParentsWithRespectToInputs);
+            return Differentiable.super.forwardModeAutoDifferentiation(derivativeOfParentsWithRespectToInputs);
```
gordoncaleb: Differentiable.super? I think this is another interface DifferentiableBinaryOp.

ieva-improbable: So the super used to lead to the `DoubleVertex`'s implementation of `forwardModeAutoDifferentiation`, which I moved to `Differentiable` as a default. Why would we need another interface for it?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/CastDoubleVertex.java
```diff
@@ -5,13 +5,14 @@
...
 import io.improbable.keanu.vertices.Vertex;
+import io.improbable.keanu.vertices.dbl.Differentiable;
 import io.improbable.keanu.vertices.dbl.DoubleVertex;
 import io.improbable.keanu.vertices.dbl.KeanuRandom;
 import io.improbable.keanu.vertices.dbl.nonprobabilistic.diff.PartialDerivatives;
 
 import java.util.Map;
 
-public class CastDoubleVertex extends DoubleVertex implements NonProbabilistic<DoubleTensor> {
+public class CastDoubleVertex extends DoubleVertex  implements Differentiable, NonProbabilistic<DoubleTensor> {
```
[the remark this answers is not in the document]
ieva-improbable: Good spot, thank you!

PR #347, opened 2018-11-16
Feature/save load gpb dbl annotations

christophernorth:
Annotate the Double Vertices so they can be saved.

Also adds a few Python annotations too to export more of the Double Vertices.


keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/BetaVertex.java
```diff
@@ -21,10 +25,12 @@
 import static io.improbable.keanu.tensor.TensorShapeValidation.checkHasSingleNonScalarShapeOrAllScalar;
 import static io.improbable.keanu.tensor.TensorShapeValidation.checkTensorsMatchNonScalarShapeOrAreScalar;
 
-public class BetaVertex extends DoubleVertex implements Differentiable, ProbabilisticDouble, SamplableWithManyScalars<DoubleTensor> {
+public class BetaVertex extends DoubleVertex implements SaveableVertex, Differentiable, ProbabilisticDouble, SamplableWithManyScalars<DoubleTensor> {
 
     private final DoubleVertex alpha;
     private final DoubleVertex beta;
+    private static final String ALPHA_NAME = "alpha";
```
migwellian: you've introduced lots of `"alpha"`s etc. - these should be consolidated in one place, and preferably consolidated with the existing `ParameterName`s in `io.improbable.keanu.distributions.hyperparam.Diffs` (which could be moved somewhere else, since they have nothing to do with differentiation specifically)

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/DoubleIfVertex.java
```diff
@@ -13,24 +16,51 @@
...
+
+public class DoubleIfVertex extends DoubleVertex implements SaveableVertex, Differentiable, NonProbabilistic<DoubleTensor> {
 
     private final Vertex<? extends BooleanTensor> predicate;
-    private final Vertex<? extends DoubleTensor> thn;
-    private final Vertex<? extends DoubleTensor> els;
+    private final DoubleVertex thn;
+    private final DoubleVertex els;
+    protected static final String PREDICATE_NAME = "predicate";
+    protected static final String THN_NAME = "thn";
```
GeorgeNash: Can we just call these `THEN_NAME` and `ELSE_NAME` and do `getThen` etc. The `thn` and `els` are just to avoid the keyword

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/binary/ArcTan2Vertex.java
```diff
@@ -10,16 +12,31 @@
...
      * @param y y coordinate
      */
-    public ArcTan2Vertex(DoubleVertex x, DoubleVertex y) {
+    @ExportVertexToPythonBindings
+    public ArcTan2Vertex(@LoadParentVertex(X_NAME) DoubleVertex x,
+                         @LoadParentVertex(Y_NAME) DoubleVertex y) {
         super(x, y);
     }
 
+    // No SaveParentVertex annotation needed as DoubleBinaryOpVertex.getLeft() will be used when saving the vertex
```
GeorgeNash: I don't think the comments are needed here - a bit overkill

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/binary/ArcTan2Vertex.java
```diff
@@ -10,16 +12,31 @@
 
 public class ArcTan2Vertex extends DoubleBinaryOpVertex {
 
+    private static final String X_NAME = LEFT_NAME;
```
gordoncaleb: Nice

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/binary/ArcTan2Vertex.java
```diff
@@ -10,16 +12,31 @@
...
      */
-    public ArcTan2Vertex(DoubleVertex x, DoubleVertex y) {
+    @ExportVertexToPythonBindings
+    public ArcTan2Vertex(@LoadParentVertex(X_NAME) DoubleVertex x,
+                         @LoadParentVertex(Y_NAME) DoubleVertex y) {
         super(x, y);
     }
 
+    // No SaveParentVertex annotation needed as DoubleBinaryOpVertex.getLeft() will be used when saving the vertex
+    public DoubleVertex getX() {
```
GeorgeNash: Seems a bit tedious having to bang in these getters for `left` and `right` when it goes to the super anyway. I'm a bit confused on why we need the getters and the `X_NAME` and `Y_NAME` - can't we just use `LEFT` and `RIGHT` from super?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/BetaVertex.java
gordoncaleb: I think I've gone off the ParameterName approach. I think it unnecessarily tightly couples all of the distribution vertices. There might be two distribution that have a hyper param called alpha but I think the duplicated code that defines it's name (e.g. `private static final String ALPHA_NAME = "alpha";`) is worth it in order to avoid the tight coupling that removing the duplication would cause.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/DoubleIfVertex.java
GeorgeNash: or do they need to exactly match the params?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/binary/ArcTan2Vertex.java
GeorgeNash: Similar to how you've got no getters for the `unary` ops

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/binary/MaxVertex.java
```diff
@@ -11,7 +13,9 @@
      * @param left  one of the vertices to find the maximum of
      * @param right one of the vertices to find the maximum of
      */
-    public MaxVertex(DoubleVertex left, DoubleVertex right) {
+    @ExportVertexToPythonBindings
+    public MaxVertex(@LoadParentVertex(THN_NAME) DoubleVertex left,
```
gordoncaleb: THN_NAME?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/binary/ArcTan2Vertex.java
GeorgeNash: Seems like we've got this for a few of the vertices already - like `MaxVertex`

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/binary/MaxVertex.java
gordoncaleb: Do we need left and right getters on this?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/BetaVertex.java
migwellian: I'm not particularly attached to the `ParameterNames` but don't you think some consistency with the strings is still important? Not just for the protobufs but for text-based serialisation formats too e.g. for the DOT format.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/DoubleIfVertex.java
christophernorth: No - they don't need to match to work.  I'll rename.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/binary/ArcTan2Vertex.java
christophernorth: Cool - will remove

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/binary/ArcTan2Vertex.java
christophernorth: Aha - this is actually Jon's code.  I'll have a scan through and make it consistent where we only have getters on the parent I think.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/binary/MaxVertex.java
christophernorth: Sorted the naming.  We don't really need the getters here (and can't really put them on without adding extra duplication at the moment.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/binary/MaxVertex.java
```diff
@@ -11,7 +13,9 @@
      * @param left  one of the vertices to find the maximum of
      * @param right one of the vertices to find the maximum of
      */
-    public MaxVertex(DoubleVertex left, DoubleVertex right) {
+    @ExportVertexToPythonBindings
+    public MaxVertex(@LoadParentVertex(THEN_NAME) DoubleVertex left,
```
GeorgeNash: I know I've made this wrap the `IF` and that's why we have the `THEN` and the `ELSE` name but it doesn't  make much sense for `Min` and `Max`. Can we change these names to `LEFT` and `RIGHT` to match the double binary op's??

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/ReshapeVertex.java
```diff
@@ -13,6 +14,7 @@
 
 public class ReshapeVertex extends DoubleUnaryOpVertex implements Differentiable {
 
+    @ExportVertexToPythonBindings
```
GeorgeNash: Not saveable?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/SliceVertex.java
```diff
@@ -25,6 +26,7 @@
      * @param dimension   the dimension to extract along
      * @param index       the index of extraction
      */
+    @ExportVertexToPythonBindings
```
GeorgeNash: Not saveable?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/SumVertex.java
```diff
@@ -27,6 +28,7 @@
      * @param inputVertex    the vertex to have its values summed
      * @param overDimensions dimensions to sum over
      */
+    @ExportVertexToPythonBindings
```
GeorgeNash: Not saveable?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/TakeVertex.java
```diff
@@ -24,6 +25,7 @@
      * @param inputVertex the input vertex to extract from
      * @param index       the index to extract at
      */
+    @ExportVertexToPythonBindings
```
GeorgeNash: Not saveable?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/binary/MaxVertex.java
christophernorth: That won't work as the saves are annotated in the parent class.  I can put in a "pseudonym" final string - but it will still get saved as "then" and "else" due to the way the serialiser works in these cases.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/ReshapeVertex.java
christophernorth: Not at the moment - requires more save functionality

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/SliceVertex.java
christophernorth: As above

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/SumVertex.java
christophernorth: As above.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/unary/TakeVertex.java
christophernorth: As above.

PR #349, opened 2018-11-16
Python formatting

wesleyw72:
Added a build step to run [yapf](https://github.com/google/yapf) to enforce a python style. 

- Format check runs prior to tests so that builds fail as early as possible.
- Added task `formatApply` to automatically fix any formatting issues.

Have chosen to use the `google` style. However, this can be changed and customised to select a project preferred style.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,8 @@ For IntelliJ:
 
 We use [Spotless](https://github.com/diffplug/spotless/tree/master/plugin-gradle) to automatically enforce some basic code style checks. If your build fails due to a formatting issue, simply run `./gradlew spotlessApply` and commit the changes.
 
+We use [yapf](https://github.com/google/yapf) for formatting python code. If your build fails due to a python formatting issue, run './gradlew keanu-python:formatApply' and commit the changes.  
+
 
 #### Python Code Generation
 
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ For IntelliJ:
 
 We use [Spotless](https://github.com/diffplug/spotless/tree/master/plugin-gradle) to automatically enforce some basic code style checks. If your build fails due to a formatting issue, simply run `./gradlew spotlessApply` and commit the changes.
 
-We use [yapf](https://github.com/google/yapf) for formatting python code. If your build fails due to a python formatting issue, run './gradlew keanu-python:formatApply' and commit the changes.  
+We use [yapf](https://github.com/google/yapf) for formatting python code. If your build fails due to a python formatting issue, run `./gradlew keanu-python:formatApply` and commit the changes.  
 
 
 #### Python Code Generation
--- a/README.md
+++ b/README.md
@@ -74,10 +74,7 @@ For IntelliJ:
 
 #### Formatting
 
-We use [Spotless](https://github.com/diffplug/spotless/tree/master/plugin-gradle) to automatically enforce some basic code style checks. If your build fails due to a formatting issue, simply run `./gradlew spotlessApply` and commit the changes.
-
-We use [yapf](https://github.com/google/yapf) for formatting python code. If your build fails due to a python formatting issue, run `./gradlew keanu-python:formatApply` and commit the changes.  
-
+We use [Spotless](https://github.com/diffplug/spotless/tree/master/plugin-gradle) and [yapf](https://github.com/google/yapf) to automatically enforce some basic code style checks. If your build fails due to a formatting issue, simply run `./gradlew formatApply` and commit the changes.
 
 #### Python Code Generation
 
--- a/keanu-python/build.gradle
+++ b/keanu-python/build.gradle
@@ -103,7 +103,9 @@ task formatCheck {
     dependsOn("preparePythonEnvironment")
     doLast { 
         exec {
-            commandLine 'pipenv', 'run', 'yapf', '--style', 'google', '--diff', '--recursive', '--parallel',  '.'
+            // Exclude files that are generated
+            commandLine 'pipenv', 'run', 'yapf', '--style', 'google', '--diff', '--recursive', '--parallel', 
+            '--exclude', 'keanu/vertex/generated.py', '--exclude', 'keanu/vertex/__init__.py', '.'
         }
     }
 }
--- a/keanu-python/build.gradle
+++ b/keanu-python/build.gradle
@@ -1,4 +1,6 @@
-import static org.apache.tools.ant.taskdefs.condition.Os.*
+import static org.apache.tools.ant.taskdefs.condition.Os.isFamily;
+import static org.apache.tools.ant.taskdefs.condition.Os.FAMILY_WINDOWS;
+
 task installPipenv {
     doLast {
         exec {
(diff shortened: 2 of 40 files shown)


keanu-python/build.gradle
```diff
@@ -93,6 +93,37 @@ task preparePythonEnvironment {
...
+    }
+}
+
+task formatCheck {
+    dependsOn("preparePythonEnvironment")
+    doLast { 
+        exec {
+            // Exclude files that are generated
+            def keanuVetexPath = 'keanu/vertex/'
+            if(isFamily(FAMILY_WINDOWS)) {
```
gordoncaleb: Talked about this offline. We should add a task to make our generated code pass the linting.

keanu-python/examples/lorenz_model.py
```diff
@@ -17,10 +17,12 @@ def run_model(self, num_time_steps):
         for _ in range(num_time_steps):
             position = self.__get_next_position(position)
             yield position
-        
-    def __get_next_position(self, current : Coordinates) -> Coordinates :
-        nextX = current.x + self.time_step * (self.sigma * (current.y - current.x))
```
gordoncaleb: How long is the max line length? Is it possible to increase it?

icantsayrural: I'm wondering this too

wesleyw72: Google style is 80 by default. We can override this. Shall I try 100 and see what that looks like?

migwellian:
How does this workflow interact with an IDE like PyCharm or VSCode? I would like the experience to be as frictionless as possible, i.e. most of the time the IDE auto-formats for you and the gradle build passes without you having to run `formatApply`.

It looks like there's a yapf plugin for pycharm but not for VSCode. Have you tried it? I don't think anyone's too wedded to an IDE at this point.

keanu-python/examples/lorenz_model.py
migwellian: go *120*

wesleyw72:
@migwellian You can do the same with [VSCode](https://code.visualstudio.com/docs/python/editing) without a plugin.

keanu-python/build.gradle
migwellian: Gradle configs are just Groovy, which is a superset of Java, so you can solve this problem using Java's `Paths` class or `File.separator`

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -5,33 +5,61 @@
...
+           draws=500,
+           drop=0,
+           down_sample_interval=1):
     vertices_unwrapped = k.to_java_object_list(sample_from)
 
-    network_samples = algorithms[algo].withDefaultConfig().getPosteriorSamples(net.unwrap(), vertices_unwrapped, draws).drop(drop).downSample(down_sample_interval)
-    vertex_samples = {Vertex._get_python_id(vertex_unwrapped): list(map(Tensor._to_ndarray, network_samples.get(vertex_unwrapped).asList())) for vertex_unwrapped in vertices_unwrapped}
+    network_samples = algorithms[algo].withDefaultConfig().getPosteriorSamples(
+        net.unwrap(), vertices_unwrapped,
+        draws).drop(drop).downSample(down_sample_interval)
```
migwellian: This looks weird. What's it doing?
 
method name, open parentheses
2 args
1 arg, close parentheses, chained method calls

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -5,33 +5,61 @@
...
+def generate_samples(net,
+                     sample_from,
+                     algo='metropolis',
+                     drop=0,
+                     down_sample_interval=1):
+    vertices_unwrapped = k.to_java_object_list(sample_from)
+
+    sample_iterator = algorithms[algo].withDefaultConfig(
+    ).generatePosteriorSamples(net.unwrap(), vertices_unwrapped)
+    sample_iterator = sample_iterator.dropCount(drop).downSampleInterval(
```
migwellian: weird to split on empty parentheses. I guess this one goes away when we increase line length limit.

keanu-python/keanu/keanu_random.py
```diff
@@ -16,4 +17,4 @@ def __init__(self, seed=None):
 
     @staticmethod
     def set_default_random_seed(seed):
-    	k.jvm_view().KeanuRandom.setDefaultRandomSeed(seed)
+        k.jvm_view().KeanuRandom.setDefaultRandomSeed(seed)
```
migwellian: good catch yapf!

keanu-python/tests/test_lorenz.py
```diff
@@ -39,39 +38,43 @@ def test_lorenz():
...
         actual_at_post_t = observed[post_t]
 
-        error = math.sqrt(
-                    (actual_at_post_t.x - posterior[0]) ** 2 +
-                    (actual_at_post_t.y - posterior[1]) ** 2 +
-                    (actual_at_post_t.z - posterior[2]) ** 2
-                )
+        error = math.sqrt((actual_at_post_t.x - posterior[0])**2 +
+                          (actual_at_post_t.y - posterior[1])**2 +
+                          (actual_at_post_t.z - posterior[2])**2)
```
migwellian: this is odd, it's as if it doesn't know `**` is a mathematical operator

keanu-python/tests/test_vertex_operations.py
```diff
@@ -6,42 +6,58 @@
...
+         np.array([[True], [False]])),
+        (Const(np.array([1.])), 1., np.array([[True]])),
+        (Const(np.array([2.])), 1., np.array([[False]])),
+        (1., Const(np.array([1.])), np.array([[True]])),
+        (1., Const(np.array([-1.])), np.array([[False]])),
+        (Const(np.array([1., 2.])), Const(np.array([1.])),
+         np.array([[True], [False]])),
+        (Const(np.array([1.])), Const(np.array([1., 2.])),
+         np.array([[True], [False]])),
+    ])
```
migwellian: is it possible to mark specific blocks as "don't touch this formatting"?

wesleyw72: Yes. With '# yapf: disable' and '# yapf: enable'. I will wrap this block.

migwellian: great!

please do all the similar bits in this file, with lists of parameters

keanu-python/build.gradle
wesleyw72: I had tried `File.seperator` in [this](https://github.com/improbable-research/keanu/pull/349/commits/5c327b0776b2300ac939e2279f14b9b1e71ba571) commit. As shown in the Appveyor run for that commit, it also fails. It seems Windows will only exclude if the path is `.\keanu\vertex\file.py` and not `keanu\vertex\file.py`. 

Setting the path as `'.' + 'File.seperator + 'kenau' ...` won't work as yapf will error when the file starts `./` (see [here](https://github.com/google/yapf/blob/master/yapf/yapflib/file_resources.py#L117)) which would be the case for MacOS, Linux etc.

Have just tried Java's `Paths` and get the exact same issue.

keanu-python/examples/lorenz_model.py
wesleyw72: Increased to 120.

README.md
```diff
@@ -76,6 +76,8 @@ For IntelliJ:
 
 We use [Spotless](https://github.com/diffplug/spotless/tree/master/plugin-gradle) to automatically enforce some basic code style checks. If your build fails due to a formatting issue, simply run `./gradlew spotlessApply` and commit the changes.
 
+We use [yapf](https://github.com/google/yapf) for formatting python code. If your build fails due to a python formatting issue, run `./gradlew keanu-python:formatApply` and commit the changes.
```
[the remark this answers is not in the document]
wesleyw72: Good point. Done.

keanu-python/build.gradle
migwellian: Hmm, so it's yapf's fault. I don't see why they wouldn't allow `./keanu/vertex/` as an exclude path.

Have you tried it using absolute paths? You can use the special gradle variable `$rootDir`

wesleyw72: The file handling doesn't seem great currently. I think ideally as Caleb mentioned we will overcome this by making sure the generated files are formatted.

I have tried absolute paths, they don't work. I think this is down to the way the matching of paths is done when it checks for the exclusions. I attempted using the absolute path for the exclusions whilst also specifying the directory to format as an absolute path. No luck with this either.

PR #352, opened 2018-11-19
Feature/disable logging in tests sol 2786

ieva-improbable:
This PR makes sure that logging is done through `slf4j` logging facade instead of `System.out` where possible. It also disables logging for tests by adding an extra `log4j` logger configuration (configuration for tests specifies logging only error level logs, while the configuration for prod sets logging to debug level), as well disabling loading bar when tests are ran from gradle.

Up for debate:
* I didn't change logging in the code used for doc snippets as I thought it would be confusing for the readers. So these are still using `System.out.println` (for instance, _WetGrass.java_);
* Some tests reside under _keanu-examples_ directory. I did not add a specific logger config for those, but it could be easily done if we decided that we want to change logging settings for them.


keanu-examples/bayesianMethodsForHackers/src/test/java/com/examples/ChallengerDisasterTest.java
```diff
@@ -3,12 +3,16 @@
...
 import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.within;
 
 public class ChallengerDisasterTest {
 
+    private final Logger log = LoggerFactory.getLogger(ChallengerDisasterTest.class);
```
migwellian: The convention is always to make it `static`. However @gordoncaleb has just pointed out to me that you can annotate the class with Lombok's `@Slf4j` and then you can remove this line!

keanu-project/src/main/resources/log4j.properties
```diff
@@ -0,0 +1,10 @@
+# Root logger option
+log4j.rootLogger=INFO, stdout
+log4j.logger.io.improbable.keanu=INFO
+log4j.logger.org.nd4j=DEBUG
```
migwellian: shall we whack all of these up to WARN as the default?

keanu-project/src/test/resources/log4j.properties
```diff
@@ -1,7 +1,7 @@
 # Root logger option
-log4j.rootLogger=WARN, stdout
-log4j.logger.io.improbable.keanu=DEBUG
-log4j.logger.org.nd4j=DEBUG
+log4j.rootLogger=ERROR
```
migwellian: why did you remove `, stdout`?

keanu-project/src/main/resources/log4j.properties
ieva-improbable: Yeah, I think that would make sense, but wasn't sure what defaults others prefer.

ieva-improbable: We do have a bunch of logs that are on INFO level, these would not be displayed if we change it to WARN. Do you think that'd be fine?

keanu-project/src/test/resources/log4j.properties
ieva-improbable: By accident, thanks for spotting!

keanu-project/src/main/resources/log4j.properties
migwellian: I doubt anyone would miss them. Worth getting 1 or 2 more opinions though.

wesleyw72: I think WARN by default is fine. If anyone wants to particularly probe into whats happening they can override it.

wesleyw72:
LGTM

PR #353, opened 2018-11-20
Feature/save load nonsaveable interface

wesleyw72:
Vertices which can't be saved are now labelled as `NonSaveable` rather than labelling those that can be saved as `Saveable`. This implies that vertices default to being `Saveable` without inheriting an interface.


keanu-project/src/main/java/io/improbable/keanu/vertices/generic/nonprobabilistic/operators/binary/BinaryOpVertex.java
```diff
@@ -1,10 +1,11 @@
 package io.improbable.keanu.vertices.generic.nonprobabilistic.operators.binary;
 
 import io.improbable.keanu.vertices.NonProbabilistic;
+import io.improbable.keanu.vertices.NonSaveableVertex;
 import io.improbable.keanu.vertices.Vertex;
 import io.improbable.keanu.vertices.dbl.KeanuRandom;
 
-public abstract class BinaryOpVertex<A, B, C> extends Vertex<C> implements NonProbabilistic<C> {
+public abstract class BinaryOpVertex<A, B, C> extends Vertex<C> implements NonProbabilistic<C>, NonSaveableVertex {
```
christophernorth: Probably best not to put this on an abstract class and better to stick it on the concrete subclasses?

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -215,11 +216,20 @@ public void incrementIndentation() {
...
+            for (Vertex vertex : TopologicalSort.sort(vertices)) {
+                vertex.save(networkWriter);
+            }
+        }
+        else {
+            throw new IllegalArgumentException("Trying to save a BayesianNetwork that isn't Saveable");
         }
     }
 
+    private Boolean isSaveable() {
```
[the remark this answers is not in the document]
wesleyw72: Good spot, meant to use the primitive. Have fixed.

keanu-project/src/main/java/io/improbable/keanu/vertices/generic/nonprobabilistic/operators/binary/BinaryOpVertex.java
wesleyw72: Agreed, have fixed.

PR #362, opened 2018-11-26
Codegen javadoc to json classpath

wesleyw72:
This was causing the problem where `error: package com.google.protobuf does not exist` was reported from the `codegen` project.

## Cause
The task `saveJavadocStringsToFile` analysed the source. However, it would generate warnings whenever an unknown dependency was encountered. The dependencies on the `keanu-project` classpath were not being correctly included in this task. The warnings were capped to 100, therefore only showing the protobuf issues.

## Solution
- Stop processing generated and test sources. These are not used in the docs, so no need to process. Removes the need for including the compile classpath.
- Set the classpath for javadoc task to the runtime classpath of `keanu-project`.
- Include the compile class path of `keanu-project` in order to support Lombok.

### DIFF
--- a/codegen/build.gradle
+++ b/codegen/build.gradle
@@ -19,15 +19,14 @@ dependencies {
 }
 
 task saveJavadocStringsToFile(type: Javadoc) {
-    source = fileTree(dir:'../keanu-project', includes:["**/*.java"])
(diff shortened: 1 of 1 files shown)


gordoncaleb:
Nice work. Also, nice PR description!

PR #364, opened 2018-11-27
Heirarchical Linear Regression

GeorgeNash:
PYMC3 blogpost on this: https://docs.pymc.io/notebooks/GLM-hierarchical.html

- Priors on weights and intercepts can be vertices
- Observation of training data and fitting of model no longer happens in the `build` step
- There are now separate `observe` and `fit` functions
- Simple radon linear regression example with a few different levels of hierarchy
- The data is meant to be partitioned by `county`, with each `county` representing a separate linear regression. This is too many latents for the optimiser so I've simplified the problem to groups of `1`, `2`, `4`, and `10`. Hopefully NUTS will enable us to run all 200+
- I have to start the optimiser fairly near the answer for the parent probabilistic vertices to enable it to find the answer for the individual linear regression weights
- NUTS work is on going....


keanu-project/src/test/java/io/improbable/keanu/e2e/regression/DiabetesLinearRegression.java
```diff
@@ -57,6 +57,10 @@ public void doesLinearRegressionOnBMIAsModel() {
             .withRegularization(RegressionRegularization.RIDGE)
             .withPriorOnWeightsAndIntercept(0, 100)
             .build();
+
+        linearRegressionModel.observe();
```
wesleyw72: Since we do `observe()` then `fit()` so much, I wonder whether it would be worth creating an `observeAndFit()`.

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
```diff
@@ -0,0 +1,179 @@
...
+public class RadonHeirarchicalRegression {
+
+    @Rule
+    public DeterministicRule rule = new DeterministicRule();
+
+    private List<Data> radonData;
+
+    @Before
+    public void readRadonCSV() {
+        radonData = ReadCsv.fromResources("data/datasets/radon/radon.csv")
```
wesleyw72: Could we put this in `CsvDataResource` and load in the same way as `DiabetesLinearRegression`? Would be more consistent.

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
```diff
@@ -0,0 +1,179 @@
...
+            withPriorOnIntercept(0., 5.).
+            build();
+
+        model.observe();
+        model.fit();
+
+        return model;
+    }
+
+    private void buildHeirarchicalNetwork(List<Data> radonData, int numberOfModels) {
```
wesleyw72: This builds the network and runs an optimiser on it. The name should reflect that.

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
```diff
@@ -0,0 +1,179 @@
...
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class RadonHeirarchicalRegression {
```
wesleyw72: Spelling: hierarchical

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
```diff
@@ -0,0 +1,179 @@
...
+
+    @Test
+    public void canPerformSimpleLinearRegression() {
+        RegressionModel model = linearRegression(radonData);
+        Assert.assertTrue(model.getWeight(0) > -0.7 && model.getWeight(0) < -0.4);
+        Assert.assertTrue(model.getIntercept() > 1.2 && model.getIntercept() < 1.5);
+    }
+
+    @Test
+    public void canPerformRegressionWithOneHeirarchy() {
```
wesleyw72: Spelling: hierarchy

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
```diff
@@ -0,0 +1,179 @@
...
+        buildHeirarchicalNetwork(radonData, 1);
```
wesleyw72: Spelling: hierarchical

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
```diff
@@ -0,0 +1,179 @@
...
+        double[] allRadon = radonData.stream().mapToDouble(k -> k.log_radon).toArray();
+
+        List<RegressionModel> models = new ArrayList<>();
+
+        for (int i = 0; i < numberOfModels; i++) {
+            RegressionModel model = createSubModel(allFloor, allRadon, i, numPartitions, muBeta, muAlpha, sigmaBeta, sigmaAlpha);
+            models.add(model);
+        }
+
+        muAlpha.setValue(1.);
```
wesleyw72: A comment to why these value are set would be useful as it's a bit of an odd case.

keanu-project/src/main/java/io/improbable/keanu/model/regression/LinearRegressionGraph.java
```diff
@@ -59,6 +59,14 @@ public DoubleTensor getWeights() {
         return weightsVertex.getValue();
     }
 
+    public DoubleVertex getInterceptVertex() {
```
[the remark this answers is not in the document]
migwellian: I would prefer to keep the API smaller - have `getInterceptValue` instead of `getInterceptVertex`. Ditto for `weight`

keanu-project/src/main/java/io/improbable/keanu/model/regression/RegressionModelBuilder.java
```diff
@@ -56,16 +59,23 @@ public RegressionModelBuilder withRegularization(RegressionRegularization regula
...
-        RegressionWeights.checkArrayHasCorrectNumberOfFeatures(scaleParameters, getFeatureCount());
-
+    public RegressionModelBuilder withPriorOnWeights(DoubleVertex means, DoubleVertex scaleParameters) {
         this.priorOnWeightsMeans = means;
         this.priorOnWeightsScaleParameters = scaleParameters;
-
         return this;
     }
 
+    public RegressionModelBuilder withPriorOnWeights(DoubleTensor means, DoubleTensor scaleParameters) {
```
migwellian: for symmetry we ought to have a method ` withPriorOnIntercept(DoubleTensor means, DoubleTensor scaleParameters)`

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
```diff
@@ -0,0 +1,179 @@
...
+        radonData = ReadCsv.fromResources("data/datasets/radon/radon.csv")
+            .asRowsDefinedBy(Data.class)
+            .load(true);
+    }
+
+    @Test
+    public void canPerformSimpleLinearRegression() {
+        RegressionModel model = linearRegression(radonData);
+        Assert.assertTrue(model.getWeight(0) > -0.7 && model.getWeight(0) < -0.4);
+        Assert.assertTrue(model.getIntercept() > 1.2 && model.getIntercept() < 1.5);
```
migwellian: Hamcrest matchers will have better error reporting:

    assertThat(model.getIntercept(), both(greaterThan(1.2)).and(lessThan(1.5)));

keanu-project/src/main/java/io/improbable/keanu/model/regression/RegressionModelBuilder.java
```diff
@@ -118,8 +138,11 @@ public RegressionModelBuilder withSampling(SamplingModelFitting sampling) {
             getWeightsVertex()
         );
 
-        performDataFitting(regressionGraph, outputTrainingData);
```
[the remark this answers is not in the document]
migwellian: He did (line 148)

keanu-project/src/main/java/io/improbable/keanu/model/regression/RegressionModelBuilder.java
```diff
@@ -145,14 +162,6 @@ private DoubleVertex getWeightsVertex() {
         return this.regularization.getWeightsVertex(getFeatureCount(), priorOnWeightsMeans, priorOnWeightsScaleParameters);
     }
 
-    private void performDataFitting(LinearRegressionGraph<OUTPUT> regressionGraph, OUTPUT outputTrainingData) {
-        ModelFitter<DoubleTensor, OUTPUT> fitter = samplingAlgorithm == null ?
-            this.regularization.createFitterForGraph(regressionGraph) :
-            samplingAlgorithm.createFitterForGraph(regressionGraph);
-
-        fitter.fit(inputTrainingData, outputTrainingData);
```
migwellian: If you no longer do the fitting in the builder, you will need to update `SimpleLinearRegressionExample`

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
```diff
@@ -0,0 +1,179 @@
...
+        RegressionModel model = RegressionModel.
+            withTrainingData(x, y).
+            withRegularization(RegressionRegularization.RIDGE).
+            withPriorOnWeights(muGradient, sigmaGradient).
+            withPriorOnIntercept(muIntercept, sigmaIntercept).
+            build();
+
+        model.observe();
+
+        return model;
```
migwellian: so what's the deal with these models? They never get fitted?

keanu-project/src/main/java/io/improbable/keanu/model/regression/RegressionModelBuilder.java
```diff
@@ -14,17 +15,19 @@
...
     private static final double DEFAULT_MU = 0.0;
     private static final double DEFAULT_SCALE_PARAMETER = 1.0;
 
-    private RegressionRegularization regularization = RegressionRegularization.NONE;
-    private double[] priorOnWeightsScaleParameters;
-    private double[] priorOnWeightsMeans;
-    private Double priorOnInterceptScaleParameter;
-    private Double priorOnInterceptMean;
     private final DoubleTensor inputTrainingData;
     private final OUTPUT outputTrainingData;
```
migwellian: If you no longer do the fitting inside the builder, you don't need these fields or the `withTrainingData` method

keanu-project/src/main/java/io/improbable/keanu/model/regression/RegressionModel.java
```diff
@@ -23,11 +24,16 @@
...
 public class RegressionModel<OUTPUT> implements Model<DoubleTensor, OUTPUT> {
     private static final double DEFAULT_OBSERVATION_SIGMA = 1.0;
-
+    private final DoubleTensor inputTrainingData;
+    private final OUTPUT outputTrainingData;
+    private final ModelFitter<DoubleTensor, OUTPUT> fitter;
     private final LinearRegressionGraph<OUTPUT> modelGraph;
 
-    RegressionModel(LinearRegressionGraph<OUTPUT> modelGraph) {
+    RegressionModel(LinearRegressionGraph<OUTPUT> modelGraph, DoubleTensor inputTrainingData, OUTPUT outputTrainingData, ModelFitter<DoubleTensor, OUTPUT> fitter) {
```
migwellian: I would prefer it if the input/output data were passed as arguments to `fit` and `observe`, not stored as fields.

keanu-project/src/main/java/io/improbable/keanu/model/ModelFitter.java
```diff
@@ -2,4 +2,5 @@
 
 public interface ModelFitter<INPUT, OUTPUT> {
     void fit(INPUT input, OUTPUT output);
```
migwellian: `fit` no longer needs the `input` and `output` parameters, now that you've decoupled it from `observe`

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
christophernorth: It also does all the verification.

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
```diff
@@ -0,0 +1,179 @@
...
+        muAlpha.setValue(1.);
+        sigmaAlpha.setValue(0.5);
+
+        muBeta.setValue(-1.);
+        sigmaBeta.setValue(0.5);
+
+        optimise(new BayesianNetwork(muAlpha.getConnectedGraph()), models);
+    }
+
+    private RegressionModel createSubModel(double[] allFloor,
```
christophernorth: As above - this does more than just create a sub model

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
```diff
@@ -0,0 +1,179 @@
...
+            build();
+
+        model.observe();
+
+        return model;
+    }
+
+    private void optimise(BayesianNetwork bayesianNetwork, List<RegressionModel> models) {
+        bayesianNetwork.probeForNonZeroProbability(100);
+        GradientOptimizer optimizer = GradientOptimizer.builder()
```
christophernorth: I think it's quite confusing that for Non-hierarchical models we have a nice "fit" method to get the right results, but as soon as we use a hierarchical model we resort to building a BayesNet ourselves.  Perhaps would be better to ensure that you can just deal with the LinearModel APIs to do this cleanly?

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
```diff
@@ -0,0 +1,179 @@
...
+        int startIndex = size * i;
+        int endIndex = size * (i + 1);
+
+        double[] floorForSubModel = Arrays.copyOfRange(allFloor, startIndex, endIndex);
+        double[] radonForSubModel = Arrays.copyOfRange(allRadon, startIndex, endIndex);
+
+        DoubleTensor x = DoubleTensor.create(floorForSubModel, 1, floorForSubModel.length);
+        DoubleTensor y = DoubleTensor.create(radonForSubModel, 1, floorForSubModel.length);
+
+        RegressionModel model = RegressionModel.
```
christophernorth: At the moment we'll end up with large amount of vertices instead of "Tensorised" Vertices, right?  Should we explicitly represent these hierarchies in the API to allow us to reduce the number of vertices we'll create in the graph?

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
```diff
@@ -0,0 +1,179 @@
...
+        Assert.assertTrue(model.getIntercept() > 1.2 && model.getIntercept() < 1.5);
+    }
+
+    @Test
+    public void canPerformRegressionWithOneHeirarchy() {
+        buildHeirarchicalNetwork(radonData, 1);
+    }
+
+    @Test
+    public void canPerformRegressionWithTwoHeirarchies() {
```
christophernorth: Do we have any idea why the optimiser is struggling so much with the number of latents?  I wonder whether it's a case of it falling off in to the Wilderness of the HalfGaussian?  We've seen our gradient optimiser really struggle with those limited support distributions in the past?

keanu-project/src/main/java/io/improbable/keanu/model/regression/RegressionModelBuilder.java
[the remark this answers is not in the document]
wesleyw72: That's the method definition. It's not used anywhere as far as I can see.

keanu-project/src/main/java/io/improbable/keanu/model/regression/LinearRegressionGraph.java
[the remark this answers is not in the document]
GeorgeNash: `Vertex` is needed for when we have a linear regression with more than one weight

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/DiabetesLinearRegression.java
GeorgeNash: Reduce this to just `fit()` - `observe` happens as part of the build

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
GeorgeNash: Nope - these are the sub-models that are connected together through the parent `random variable`, we `fit` the parent model

keanu-project/src/main/java/io/improbable/keanu/model/regression/RegressionModel.java
GeorgeNash: The builder is where the user specifies what type of fitter they'd like. I personally don't like having the input/output data passed to the `fit` method, I'd rather have it in the builder as we require it for determining the shape of weight 
 / intercept vertices while we build the graphs. I've put the observe into the `build()` step

keanu-project/src/main/java/io/improbable/keanu/model/regression/RegressionModel.java
```diff
@@ -73,4 +87,12 @@ public VertexId getInterceptVertexId() {
     public VertexId getWeightsVertexId() {
         return modelGraph.getWeightsVertexId();
     }
+
+    public void fit() {
```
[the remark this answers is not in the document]
GeorgeNash: Mike's added the `SamplingModelFitter` lets you sample to find the most likely value + the variance

keanu-project/src/main/java/io/improbable/keanu/model/regression/RegressionModelBuilder.java
GeorgeNash: I prefer defining the complete data set / fitting inside the builder, or else we have to `build` and then continue passing more parameters which feels incorrect

keanu-project/src/main/java/io/improbable/keanu/model/regression/RegressionRegularization.java
```diff
@@ -4,21 +4,17 @@
...
 import io.improbable.keanu.vertices.dbl.probabilistic.LaplaceVertex;
 
 public enum RegressionRegularization {
     NONE {
-        public DoubleVertex getWeightsVertex(long featureCount, double[] priorOnWeightsMeans, double[] priorOnInterceptScaleParameter) {
-            return new GaussianVertex(new long[]{1, featureCount},
-                ConstantVertex.of(priorOnWeightsMeans, 1, featureCount),
-                ConstantVertex.of(priorOnInterceptScaleParameter, 1, featureCount)).setLabel("weights");
+        public DoubleVertex getWeightsVertex(long featureCount,  DoubleVertex priorOnWeightsMeans, DoubleVertex priorOnWeightsScaleParameter) {
+            return new GaussianVertex(new long[]{1, featureCount}, priorOnWeightsMeans, priorOnWeightsScaleParameter).setLabel("weights");
```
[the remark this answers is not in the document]
GeorgeNash: we need a rank 2 tensor here for the matrix multiplication performed later on

keanu-project/src/main/java/io/improbable/keanu/util/csv/pojo/CsvColumnDeserializer.java
```diff
@@ -38,7 +38,7 @@ static Object convertToAppropriateType(List<String> s, Class<?> fieldType) {
         }
 
         if (fieldType == DoubleTensor.class) {
-            return DoubleTensor.create(convertToDoubles(s));
+            return DoubleTensor.create(convertToDoubles(s), 1, s.size());
```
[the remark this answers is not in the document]
GeorgeNash: as above

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
GeorgeNash: good point - I want to read in rows (a list of `Data`) whereas `CsvDataResource` only supports vectorised columns at the minute (reading into a tensor). Let's add the support for reading in as rows as a separate PR

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
GeorgeNash: Nah it struggles with a uniform too. It's just not a very good optimiser and requires a lot of `assistance` to find the right answer as the dimensionality goes up

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    private RegressionModel linearRegression(List<Data> data) {
+        double[] radon = data.stream().mapToDouble(k -> k.log_radon).toArray();
+        double[] floor = data.stream().mapToDouble(k -> k.floor).toArray();
+        DoubleTensor y = DoubleTensor.create(radon, 1, radon.length);
+        DoubleTensor x = DoubleTensor.create(floor, 1, floor.length);
+
+        RegressionModel model = RegressionModel.
+            withTrainingData(x, y).
```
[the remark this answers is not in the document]
GeorgeNash: Our CSV reader lets you read straight into tensors

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
GeorgeNash: does it?

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
GeorgeNash: I don't think Hierarchical modelling is something that will be done enough to warrant it's own API. So long as people can compose these `sub-models` together I think it's fine to let people optimise and "fit" themselves

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
GeorgeNash: How will that work when we try to do observations on sub models if they do not have their own random variables?

GeorgeNash: It's an interesting idea but I don't think it's applicable here!

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHeirarchicalRegression.java
GeorgeNash: `splitDataAndCreateSubModel`?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Wed Jun 06 15:50:00 BST 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.1-bin.zip
```
migwellian: is this needed?

keanu-project/src/main/java/io/improbable/keanu/model/ModelFitter.java
```diff
@@ -1,5 +1,5 @@
 package io.improbable.keanu.model;
 
 public interface ModelFitter<INPUT, OUTPUT> {
```
migwellian: the parameters are no longer needed

keanu-project/src/main/java/io/improbable/keanu/model/regression/LinearRegressionGraph.java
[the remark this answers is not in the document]
migwellian: OK.

keanu-project/src/main/java/io/improbable/keanu/model/regression/LinearRegressionGraph.java
```diff
@@ -67,7 +75,7 @@ public VertexId getInterceptVertexId() {
         return interceptVertex.getId();
     }
 
-    public VertexId getWeightsVertexId() {
+    public VertexId getWeightVertexId() {
```
migwellian: Since we're now exposing `getInterceptVertex` and `getWeightVertex`, can you remove the other methods from this class and `RegressionModel`, as they will only cause confusion:
1. `DoubleTensor getWeights()`
1. `double getIntercept()`
1. `VertexId getInterceptVertexId()`
1. `VertexId getWeightVertexId()`

keanu-project/src/main/java/io/improbable/keanu/model/regression/RegressionModelBuilder.java
migwellian: Agreed - we are on the same page here.

keanu-project/src/main/java/io/improbable/keanu/model/regression/RegressionModelBuilder.java
```diff
@@ -118,8 +136,12 @@ public RegressionModelBuilder withSampling(SamplingModelFitting sampling) {
...
         );
 
-        performDataFitting(regressionGraph, outputTrainingData);
-        return new RegressionModel<>(regressionGraph);
+        ModelFitter<DoubleTensor, OUTPUT> fitter = samplingAlgorithm == null ?
+            this.regularization.createFitterForGraph(regressionGraph) :
+            samplingAlgorithm.createFitterForGraph(regressionGraph);
+
+        regressionGraph.observeValues(inputTrainingData, outputTrainingData);
+        return new RegressionModel<>(regressionGraph, fitter);
```
migwellian: The `fitter` contains the `regressionGraph`, so it doesn't feel quite right that you have to pass in both the `regressionGraph` and the `fitter` here.

I wonder if the `ModelFitter` API should actually be `fit(ModelGraph)` instead of `fit()`. That would make this cleaner and you wouldn't be storing the `regressionGraph` in 2 different places.

keanu-project/src/main/java/io/improbable/keanu/model/regression/RegressionModel.java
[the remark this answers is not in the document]
christophernorth: As discussed - this also allows you to get samples so isn't just limited to value + variance so provides the functionality I'm asking for.

keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHierarchicalRegression.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    @Test
+    public void canPerformRegressionWithFourHierarchies() {
+        buildAndRunHeirarchicalNetwork(radonData, 4);
+    }
+
+    @Test
+    public void canPerformRegressionWithTenHierarchies() {
+        buildAndRunHeirarchicalNetwork(radonData, 10);
```
christophernorth: You'd mentioned before that this was there until we had the sampling fitter.  Now we have it - can we use it instead of doing this arbitrary split?

christophernorth:
> I can see @christophernorth 's point here. The sub-models never have `fit()` called on them - instead you are doing it yourself with a `GradientOptimizer` that you create later.
> 
> Perhaps there is a solution where the sub-models become part of a larger composite model, and you pass your `GradientOptimizer` to it.
> 
> In any case, I think that the changes you've made look quite neat now and are a useful small increment that enables you to do what you need to, so we should merge it in.
> 
> Given that not calling `fit()` is the exception rather than the rule, we could roll the `fit()` back into the call to `build()`, and for this one case of Hierarchical Regression you could make it explicit by calling a new method `buildWithoutFitting()`. What do you think @christophernorth ?

I think that's a good idea for now and makes it explicit that you have to do something funky for Hierarchical models at the moment.  As you say - the current code can then be merged in and we can add a cleaner "hierarchical" API if necessary.

PR #365, opened 2018-11-27
Feature/graph metrics

icantsayrural:
Graph metrics, including:
- Vertex degree
- Vertex count
- Average degree

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
@@ -85,11 +85,11 @@ public class BayesianNetwork {
         return vertexCount;
     }
 
-    public int getVertexDegree(Vertex vertex) {
+    public Integer getVertexDegree(Vertex vertex) {
         return vertexDegrees.get(vertex.getId());
     }
 
-    public int getVertexDegree(VertexId id) {
+    public Integer getVertexDegree(VertexId id) {
         return vertexDegrees.get(id);
     }
 
--- a/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
@@ -19,6 +19,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.OptionalDouble;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -30,9 +31,26 @@ public class BayesianNetwork {
     private final int TOP_LEVEL_INDENTATION = 1;
     private int indentation = TOP_LEVEL_INDENTATION;
 
+    private final int vertexCount;
+    private final Map<VertexId, Integer> vertexDegrees;
+    private final double averageVertexDegree;
+
     public BayesianNetwork(Set<? extends Vertex> vertices) {
         this.vertices = ImmutableList.copyOf(vertices);
         this.vertexLabels = buildLabelMap(vertices);
+
+        this.vertexCount = vertices.size();
+        this.vertexDegrees = calculateVertexDegrees(vertices);
+        this.averageVertexDegree = calculateAverageVertexDegree(vertices);
+    }
+
+    private Map<VertexId,Integer> calculateVertexDegrees(Set<? extends Vertex> vertices) {
+        return vertices.stream().collect(Collectors.toMap(Vertex::getId, Vertex::getDegree));
+    }
+
+    private double calculateAverageVertexDegree(Set<? extends Vertex> vertices) {
+        OptionalDouble optionalDouble = vertices.stream().mapToInt(Vertex::getDegree).average();
+        return optionalDouble.isPresent() ? optionalDouble.getAsDouble() : 0;
     }
 
     public BayesianNetwork(Collection<? extends Vertex> vertices) {
@@ -61,6 +79,22 @@ public class BayesianNetwork {
 
     List<? extends Vertex> getVertices() {
         return vertices;
+    }
+
+    public int getVertexCount() {
+        return vertexCount;
+    }
+
+    public int getVertexDegree(Vertex vertex) {
+        return vertexDegrees.get(vertex.getId());
+    }
+
+    public int getVertexDegree(VertexId id) {
+        return vertexDegrees.get(id);
+    }
+
+    public double getAverageVertexDegree() {
+        return averageVertexDegree;
     }
 
     public void setState(NetworkState state) {
--- a/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
@@ -32,7 +32,7 @@ public class BayesianNetwork {
     private int indentation = TOP_LEVEL_INDENTATION;
 
     public BayesianNetwork(Set<? extends Vertex> vertices) {
-        Preconditions.checkArgument(!vertices.isEmpty(), "List of vertices in the network is empty");
+        Preconditions.checkArgument(!vertices.isEmpty(), "A bayesian network must contain at least one vertex");
         this.vertices = ImmutableList.copyOf(vertices);
         this.vertexLabels = buildLabelMap(vertices);
     }
--- a/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
@@ -50,7 +50,7 @@ public class BayesianNetwork {
 
     private double calculateAverageVertexDegree(Set<? extends Vertex> vertices) {
         OptionalDouble optionalDouble = vertices.stream().mapToInt(Vertex::getDegree).average();
-        return optionalDouble.isPresent() ? optionalDouble.getAsDouble() : 0;
+        return optionalDouble.isPresent() ? optionalDouble.getAsDouble() : 0.;
     }
 
     public BayesianNetwork(Collection<? extends Vertex> vertices) {
--- a/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
@@ -65,7 +65,7 @@ public class BayesianNetwork {
         return vertices;
     }
 
-    public int getNumVertices() {
+    public int getVertexCount() {
         return getVertices().size();
     }
 
--- a/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
@@ -1,5 +1,6 @@
 package io.improbable.keanu.network;
 
+import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import io.improbable.keanu.algorithms.graphtraversal.TopologicalSort;
 import io.improbable.keanu.algorithms.graphtraversal.VertexValuePropagation;
@@ -19,7 +20,6 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.OptionalDouble;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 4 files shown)


keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -30,9 +31,26 @@
...
+        this.averageVertexDegree = calculateAverageVertexDegree(vertices);
+    }
+
+    private Map<VertexId,Integer> calculateVertexDegrees(Set<? extends Vertex> vertices) {
+        return vertices.stream().collect(Collectors.toMap(Vertex::getId, Vertex::getDegree));
+    }
+
+    private double calculateAverageVertexDegree(Set<? extends Vertex> vertices) {
+        OptionalDouble optionalDouble = vertices.stream().mapToInt(Vertex::getDegree).average();
+        return optionalDouble.isPresent() ? optionalDouble.getAsDouble() : 0.;
```
ieva-improbable: I'd rather have a check in the constructor to see whether the set of vertices passed in isn't empty, this seems like the wrong place to check for it.

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -63,6 +81,22 @@ public Vertex getVertexByLabel(VertexLabel label) {
...
+    public int getVertexCount() {
+        return vertexCount;
+    }
+
+    public Integer getVertexDegree(Vertex vertex) {
+        return vertexDegrees.get(vertex.getId());
+    }
+
+    public Integer getVertexDegree(VertexId id) {
+        return vertexDegrees.get(id);
```
ieva-improbable: For this one and the previous method - would we want to return null or throw an exception if a vertex that is not a part of the network is passed in?

keanu-project/src/test/java/io/improbable/keanu/network/BayesianNetworkGetSubgraphTest.java
```diff
@@ -48,6 +49,12 @@ public static void setUp() {
 
         when(v.getConnectedGraph()).thenReturn(ImmutableSet.of(v, parent, child, grandChild, grandParent));
 
+        when(grandParent.getId()).thenReturn(new VertexId(0));
+        when(parent.getId()).thenReturn(new VertexId(1));
+        when(v.getId()).thenReturn(new VertexId(2));
+        when(child.getId()).thenReturn(new VertexId(3));
+        when(grandChild.getId()).thenReturn(new VertexId(4));
+
```
ieva-improbable: Just curious, how was the test breaking?

keanu-project/src/test/java/io/improbable/keanu/network/BayesianNetworkTest.java
```diff
@@ -102,4 +107,68 @@ public void networkWithNonSaveableVerticesThrowsExceptionOnSave() throws IOExcep
...
+            .apply((v) -> v.plus(ConstantVertex.of(1.)));
+         Vertex<?> output = loop.getOutput();
+
+         Set<Vertex> connectedGraphInLoop = output.getConnectedGraph();
+         BayesianNetwork net = new BayesianNetwork(connectedGraphInLoop);
+
+         assertThat(net.getVertexCount(), equalTo(1 + 9 * Loop.DEFAULT_MAX_COUNT + 2));
+         assertThat(net.getVertexDegree(output), equalTo(3));
+         assertThat(net.getAverageVertexDegree(), equalTo(connectedGraphInLoop.stream().mapToInt(Vertex::getDegree).average().getAsDouble()));
+    }
```
ieva-improbable: nit: might be nice to pull all the metrics tests out into a separate class, as there are quite a few of them.

keanu-project/src/test/java/io/improbable/keanu/network/BayesianNetworkTest.java
```diff
@@ -102,4 +107,68 @@ public void networkWithNonSaveableVerticesThrowsExceptionOnSave() throws IOExcep
...
     }
+
+    @Test
+    public void youCanGetVertexCountOfInitialBayesNet() {
+        ConstantDoubleVertex probTrue = ConstantVertex.of(0.5);
+        BoolVertex a = new BernoulliVertex(probTrue);
+        Set<Vertex> connectedGraph = a.getConnectedGraph();
+        BayesianNetwork net = new BayesianNetwork(connectedGraph);
+
+        BoolVertex b = new BernoulliVertex(0.5);
```
wesleyw72: This seems to be testing that adding vertices after a Bayes net has been created doesn't change the vertex count. I think this is worthy of it's own test, would be much clearer as to what's going on and the expected behaviour.

keanu-project/src/test/java/io/improbable/keanu/network/BayesianNetworkTest.java
```diff
@@ -102,4 +107,68 @@ public void networkWithNonSaveableVerticesThrowsExceptionOnSave() throws IOExcep
...
+    public void youCanGetVertexDegreeOfInitialBayesNet() {
+        ConstantDoubleVertex probTrue = ConstantVertex.of(0.5);
+        BoolVertex a = new BernoulliVertex(probTrue);
+        BayesianNetwork net = new BayesianNetwork(a.getConnectedGraph());
+
+        BoolVertex b = new BernoulliVertex(0.5);
+        BoolVertex ored = a.or(b);
+
+        assertThat(net.getVertexDegree(a.getId()), equalTo(1));
+        assertThat(net.getVertexDegree(a), equalTo(1));
```
wesleyw72: These test two different methods, should they be two different tests?

keanu-project/src/test/java/io/improbable/keanu/network/BayesianNetworkTest.java
```diff
@@ -102,4 +107,68 @@ public void networkWithNonSaveableVerticesThrowsExceptionOnSave() throws IOExcep
...
+        assertThat(net.getVertexCount(), equalTo(2));
+    }
+
+    @Test
+    public void youCanGetVertexDegreeOfInitialBayesNet() {
+        ConstantDoubleVertex probTrue = ConstantVertex.of(0.5);
+        BoolVertex a = new BernoulliVertex(probTrue);
+        BayesianNetwork net = new BayesianNetwork(a.getConnectedGraph());
+
+        BoolVertex b = new BernoulliVertex(0.5);
```
wesleyw72: This seems to be testing that adding vertices after a Bayes net has been created doesn't change the degree. I think this is worthy of it's own test, would be much clearer as to what's going on and the expected behaviour.

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -30,9 +31,26 @@
...
+    private final int vertexCount;
+    private final Map<VertexId, Integer> vertexDegrees;
+    private final double averageVertexDegree;
+
     public BayesianNetwork(Set<? extends Vertex> vertices) {
         this.vertices = ImmutableList.copyOf(vertices);
         this.vertexLabels = buildLabelMap(vertices);
+
+        this.vertexCount = vertices.size();
+        this.vertexDegrees = calculateVertexDegrees(vertices);
```
christophernorth: Is there any reason you're storing this for every Vertex?  It's not super difficult to calculate the degree for a particular Vertex (as in computationally expensive) so it seems a waste to store this even if people don't need it?

Even if we have something that could be expensive to calculate (like the average) - we could just cache that the first time someone uses it, instead of just calculating it in an unsolicited fashion?

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -63,6 +81,22 @@ public Vertex getVertexByLabel(VertexLabel label) {
         return vertices;
     }
 
+    public int getVertexCount() {
+        return vertexCount;
+    }
+
+    public Integer getVertexDegree(Vertex vertex) {
```
christophernorth: Should we have a vertex Label version of this as well?  I see that being the most useful in the "general case"?

christophernorth: In fact, if someone had the Vertex object itself, wouldn't they be best off just calling the Vertex method directly?

keanu-project/src/test/java/io/improbable/keanu/network/BayesianNetworkTest.java
```diff
@@ -102,4 +107,68 @@ public void networkWithNonSaveableVerticesThrowsExceptionOnSave() throws IOExcep
...
+
+        assertThat(net.getVertexDegree(a.getId()), equalTo(1));
+        assertThat(net.getVertexDegree(a), equalTo(1));
+    }
+
+    @Test
+    public void degreeOfNonExistentVertexIsNull() {
+        BoolVertex nonExistentVertex = new BernoulliVertex(0.5);
+        assertThat(network.getVertexDegree(nonExistentVertex), equalTo(null));
+        assertThat(network.getVertexDegree(nonExistentVertex.getId()), equalTo(null));
```
wesleyw72: These test two different methods, should they be two different tests?

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -63,6 +81,22 @@ public Vertex getVertexByLabel(VertexLabel label) {
...
+        return vertexDegrees.get(vertex.getId());
+    }
+
+    public Integer getVertexDegree(VertexId id) {
```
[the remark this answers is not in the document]
wesleyw72: Primitive `int` rather than the boxed `Integer`

keanu-project/src/test/java/io/improbable/keanu/network/BayesianNetworkTest.java
```diff
@@ -102,4 +107,68 @@ public void networkWithNonSaveableVerticesThrowsExceptionOnSave() throws IOExcep
...
+
+        BoolVertex b = new BernoulliVertex(0.5);
+        BoolVertex ored = a.or(b);
+
+        assertThat(net.getVertexDegree(a.getId()), equalTo(1));
+        assertThat(net.getVertexDegree(a), equalTo(1));
+    }
+
+    @Test
+    public void degreeOfNonExistentVertexIsNull() {
```
migwellian: wouldn't it make more sense to throw an exception?

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -63,6 +81,22 @@ public Vertex getVertexByLabel(VertexLabel label) {
         return vertices;
     }
 
+    public int getVertexCount() {
+        return vertexCount;
+    }
+
+    public Integer getVertexDegree(Vertex vertex) {
+        return vertexDegrees.get(vertex.getId());
```
migwellian: better to dedupe like this: `return getVertexDegree(vertex.getId());`

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -63,6 +81,22 @@ public Vertex getVertexByLabel(VertexLabel label) {
         return vertices;
     }
 
+    public int getVertexCount() {
+        return vertexCount;
```
migwellian: You can just `return this.vertices.size();` and there's no need to store `vertexCount` as a field

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -30,9 +31,26 @@
...
+    private final Map<VertexId, Integer> vertexDegrees;
+    private final double averageVertexDegree;
+
     public BayesianNetwork(Set<? extends Vertex> vertices) {
         this.vertices = ImmutableList.copyOf(vertices);
         this.vertexLabels = buildLabelMap(vertices);
+
+        this.vertexCount = vertices.size();
+        this.vertexDegrees = calculateVertexDegrees(vertices);
+        this.averageVertexDegree = calculateAverageVertexDegree(vertices);
```
migwellian: This class is already quite busy - I would prefer to have any metrics-related functionality in a separate class.

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -30,9 +31,26 @@
...
+    }
+
+    private Map<VertexId,Integer> calculateVertexDegrees(Set<? extends Vertex> vertices) {
+        return vertices.stream().collect(Collectors.toMap(Vertex::getId, Vertex::getDegree));
+    }
+
+    private double calculateAverageVertexDegree(Set<? extends Vertex> vertices) {
```
migwellian: the average can be inferred from `this.vertexDegrees` - no need to store or calculate it separately.

icantsayrural:
As discussed offline, the metrics API for a `BayesianNetwork` shouldn't be based on its state at construction. This means we no longer have to cache bunch of values in the constructor.

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -63,6 +65,14 @@ public Vertex getVertexByLabel(VertexLabel label) {
         return vertices;
     }
 
+    public int getNumVertices() {
```
gordoncaleb: `getNumberOfVertices` or `getVertexCount`

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -63,6 +65,14 @@ public Vertex getVertexByLabel(VertexLabel label) {
...
+        return getVertices().size();
+    }
+
+    public double getAverageVertexDegree() {
+        return getVertices().stream().mapToDouble(Vertex::getDegree).average().getAsDouble();
```
gordoncaleb: Nice stream!

PR #367, opened 2018-11-27
added a task to delete the python classpath before populating it

migwellian:
Just a little change because the python classpath was getting populated with multiple versions of the `keanu` jar which was a bad thing.

### DIFF
--- a/keanu-project/build.gradle
+++ b/keanu-project/build.gradle
@@ -149,8 +149,12 @@ task sourcesJar(type: Jar) {
     from sourceSets.main.allSource
 }
 
+task deletePythonClasspath(type: Delete) {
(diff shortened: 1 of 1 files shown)


keanu-project/build.gradle
```diff
@@ -149,8 +149,12 @@ task sourcesJar(type: Jar) {
     from sourceSets.main.allSource
 }
 
+task deletePythonClasspath(type: Delete) {
+    delete fileTree("../keanu-python/keanu/classpath/")
+}
```
wesleyw72: Nitpick: can we put a newline after the task?

migwellian: done

PR #369, opened 2018-11-28
fix return type of Tensor._to_ndarray

migwellian:
fixed bug: Tensor._to_ndarray was returning a primitive type if the java tensor was rank 0

### DIFF
--- a/keanu-python/tests/test_const.py
+++ b/keanu-python/tests/test_const.py
@@ -72,7 +72,7 @@ def test_const_does_not_take_generic_ndarray(generic):
     with pytest.raises(NotImplementedError) as excinfo:
         Const(ndarray)
 
-    assert str(excinfo.value) == "Generic types in an ndarray are not supported. Was given {}".format(type(generic))
+    assert str(excinfo.value) == "Generic types in an ndarray are not supported. Was given object"
 
 
 def test_const_does_not_take_generic(generic):
@@ -90,7 +90,7 @@ def test_const_does_not_take_empty_ndarray():
     with pytest.raises(ValueError) as excinfo:
         Const(ndarray)
 
-    assert str(excinfo.value) == "Cannot infer type because the ndarray is empty"
+    assert str(excinfo.value) == "Cannot infer type because array is empty"
 
 
 def test_const_takes_ndarray_of_rank_one():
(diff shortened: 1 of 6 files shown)


keanu-python/tests/test_const.py
```diff
@@ -90,7 +90,7 @@ def test_const_does_not_take_empty_ndarray():
     with pytest.raises(ValueError) as excinfo:
         Const(ndarray)
 
-    assert str(excinfo.value) == "Cannot infer type because the ndarray is empty"
+    assert str(excinfo.value) == "Cannot infer type because array is empty"
```
icantsayrural: It doesn't throw a `ValueError` anymore does it? And if the array is empty, I guess a Java exception would be thrown?

keanu-python/tests/test_tensor.py
```diff
@@ -81,11 +81,11 @@ def test_cannot_pass_generic_ndarray_to_Tensor(generic):
     assert str(excinfo.value) == "Generic types in an ndarray are not supported. Was given {}".format(type(generic))
 
 
-def test_cannot_pass_empty_ndarray_to_Tensor():
+def test_can_pass_empty_ndarray_to_Tensor():
     with pytest.raises(ValueError) as excinfo:
         Tensor(np.array([]))
 
-    assert str(excinfo.value) == "Cannot infer type because the ndarray is empty"
+    assert str(excinfo.value) == "Cannot infer type because array is empty"
```
icantsayrural: It doesn't throw a ValueError anymore does it? And if the array is empty, I guess a Java exception would be thrown?

keanu-python/tests/test_const.py
migwellian: it still throws a different `ValueError` - from `__infer_class_from_array`: https://github.com/improbable-research/keanu/blob/48762e3c27463b2b4b0759e2c1cd7ebcb192fc88/keanu-python/keanu/context.py#L86

keanu-python/tests/test_tensor.py
migwellian: it still throws a different `ValueError` - from `__infer_class_from_array`: https://github.com/improbable-research/keanu/blob/48762e3c27463b2b4b0759e2c1cd7ebcb192fc88/keanu-python/keanu/context.py#L86

keanu-python/tests/test_vertex.py
```diff
@@ -18,7 +19,9 @@ def jvm_view():
 def assert_vertex_value_equals_scalar(vertex, expected_type, scalar):
     vertex_value = vertex.get_value()
     assert vertex_value == scalar
-    assert type(vertex_value) == expected_type
+    assert type(vertex_value) == numpy_types
+    assert vertex_value.shape == ()
+    assert vertex_value.dtype == expected_type
```
gordoncaleb: Nice. That works.

PR #370, opened 2018-11-28
Feature/track acceptance rate

migwellian:
Putting this up early to get your feedback.

I've created a new interface `ProposalListener` and a class `AcceptanceRateTracker` that implements it.

Next step will be to write something in Python that plots this acceptance rate over time.


keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/proposal/AcceptanceRateTracker.java
```diff
@@ -0,0 +1,34 @@
...
+    @Override
+    public void onProposalApplied(Proposal proposal) {
+        Set<Vertex> key = proposal.getVerticesWithProposal();
+        log.trace(String.format(" Applied proposal: %.4f", getAcceptanceRate(key)));
+        Integer previousValue = numApplied.getOrDefault(key, 0);
+        numApplied.put(key, previousValue + 1);
+    }
+
+    public double getAcceptanceRate(Set<Vertex> verticesWithProposal) {
+        return 1. - (double) numRejected.getOrDefault(verticesWithProposal, 0) / numApplied.getOrDefault(verticesWithProposal, 0);
```
wesleyw72: Is the acceptance rate not the proportion of proposed samples that are accepted? So should be `numApplied / (numApplied + numRejected)`.

Also, if `numApplied` is zero, this will return `-Infinity`.

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/proposal/AcceptanceRateTrackerTest.java
```diff
@@ -0,0 +1,100 @@
...
+    Proposal proposal;
+    private AcceptanceRateTracker acceptanceRateTracker;
+
+    @Before
+    public void createTracker() throws Exception {
+        acceptanceRateTracker = new AcceptanceRateTracker();
+    }
+
+    @Test
+    public void theAcceptanceRateForAnUnrecognisedSetOfVerticesIsNaN() {
```
wesleyw72: Why not throw an exception if unrecognised vertices?

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/proposal/AcceptanceRateTrackerTest.java
```diff
@@ -0,0 +1,100 @@
...
+        proposal.setProposal(vertex1, 1.);
+        proposal.apply();
+        proposal.reject();
+        assertThat(acceptanceRateTracker.getAcceptanceRate(ImmutableSet.of(vertex1)), equalTo(0.));
+
+        proposal = new Proposal();
+        proposal.addListener(acceptanceRateTracker);
+        proposal.setProposal(vertex1, 2.);
+        proposal.apply();
+        assertThat(acceptanceRateTracker.getAcceptanceRate(ImmutableSet.of(vertex1)), equalTo(0.5));
```
wesleyw72: There has been three proposals, one has been accepted. Shouldn't this be equal to `1/3`?

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/proposal/AcceptanceRateTracker.java
migwellian: The meaning of "applied" is perhaps a bit unintuitive - every proposal gets applied, and then some of those might get rejected.

migwellian: So when `numApplied` is zero, `numRejected` will also be zero and the acceptance rate is `NaN`. I call this out in the test `theAcceptanceRateForAnUnrecognisedSetOfVerticesIsNaN`

wesleyw72: Ah that makes sense, thanks for clearing it up.

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/proposal/AcceptanceRateTrackerTest.java
migwellian: see my explanation above about "applied"

wesleyw72:
Thanks for clearing up my confusion on the maths. Might be worth getting another opinion on NaN vs Exception but the NaN reasoning makes sense to me.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/proposal/AcceptanceRateTracker.java
```diff
@@ -0,0 +1,34 @@
...
+@Slf4j
+public class AcceptanceRateTracker implements ProposalListener {
+
+    private Map<Set<Vertex>, Integer> numApplied = Maps.newHashMap();
+    private Map<Set<Vertex>, Integer> numRejected = Maps.newHashMap();
+
+    @Override
+    public void onProposalApplied(Proposal proposal) {
+        Set<Vertex> key = proposal.getVerticesWithProposal();
+        log.trace(String.format(" Applied proposal: %.4f", getAcceptanceRate(key)));
```
gordoncaleb: This getAcceptanceRate(key) will get called even if log.trace doesn't log anything, which means there's a performance cost to logging that may never be used and you pay the cost even when it's not used. This is one reason I don't like liberal logging all over the place. The other reason is YAGNI but that's more debatable. There's probably a clever way around this issue but we should be mindful of this sort of logging.

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/proposal/AcceptanceRateTrackerTest.java
gordoncaleb: You don't have to throw a checked exception. An IllegalArgument exception or something like that would make sense. It's very python to just return NaN and let the user figure it out.

migwellian: OK, I've implemented it as an `IllegalStateException`.

Whether or not it's a `RuntimeException` isn't the issue though - I still need to handle it in my `log.trace` logic so that it doesn't throw when logging the first call to `onProposalApplied`

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/proposal/AcceptanceRateTracker.java
migwellian: Ah, that's an interesting point. 

I would definitely debate "YAGNI" because I already did need it! However - it'll always be possible for the calling code to do the logging (especially with streaming samples). So I 
will remove it from this class.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/proposal/AcceptanceRateTracker.java
```diff
@@ -0,0 +1,34 @@
...
+
+public class AcceptanceRateTracker implements ProposalListener {
+
+    private Map<Set<Vertex>, Integer> numApplied = Maps.newHashMap();
+    private Map<Set<Vertex>, Integer> numRejected = Maps.newHashMap();
+
+    @Override
+    public void onProposalApplied(Proposal proposal) {
+        Set<Vertex> key = proposal.getVerticesWithProposal();
+        Integer previousValue = numApplied.getOrDefault(key, 0);
```
christophernorth: If I'm reading this correctly you'll end up doing two Hash code calculations for every Proposal because you're using an Immutable Integer Object to store the numApplied and numRejected.  Given how the Set.hashCode() is defined, this could be pretty expensive?

At the least could we have a little private class instead of Integer to keep the actual count so we can simply call an "increment" style method and avoid the second hash calculation?

Ideally though I'd prefer a different key for this Map - is there any way of refactoring the API here to simplify the key?

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/proposal/AcceptanceRateTracker.java
```diff
@@ -0,0 +1,34 @@
...
+        numApplied.put(key, previousValue + 1);
+    }
+
+    public double getAcceptanceRate(Set<Vertex> verticesWithProposal) {
```
christophernorth: As a user - how am I supposed to know the correct "Set" to pass here?  Might this not be dependent on the Proposal distribution in use?

migwellian: The user sets the type of `MHStepVariableSelector`. There are only two options currently:
1. select the variables one by one individually
2. select all the variables

So I _think_ this is fine, since the options are very simple and the user has control over the choice. It's possible that this may become more complicated in the future, although at present we always use option 1 so I don't think there's any demand for more choices.

I'll ask some of the data scientists for their opinion.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/proposal/AcceptanceRateTracker.java
migwellian: Thanks, that's a really interesting perspective. I hadn't given any thought to the performance of `Set.hashCode()`!

I checked the javadoc for `Set` and it says: 
> The hash code of a set is defined to be the sum of the hash codes of the elements in the set

That doesn't sound too bad?

Alternatively I could convert it to a `String` key like this, but would that be any better?

    private String generateKey(Set<Vertex> verticesWithProposal) {
        return verticesWithProposal.stream()
            .map(Vertex::getId)
            .sorted()
            .map(Object::toString)
            .collect(Collectors.joining(VERTEX_ID_DELIMITER));
    }

migwellian: To your other point - yes, I can use an `AtomicInteger` for the purpose:

        numApplied.computeIfAbsent(key, s -> new AtomicInteger(0)).incrementAndGet();

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/proposal/AcceptanceRateTracker.java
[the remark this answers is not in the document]
migwellian: yep - my bad

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/proposal/AcceptanceRateTrackerTest.java
```diff
@@ -0,0 +1,117 @@
...
+    private Vertex vertex2 = mock(Vertex.class);
+    Proposal proposal;
+    private AcceptanceRateTracker acceptanceRateTracker;
+
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
+
+    @Before
+    public void createTracker() throws Exception {
+        acceptanceRateTracker = new AcceptanceRateTracker();
```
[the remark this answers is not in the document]
migwellian: no particular reason, I could use a new `AcceptanceRateTracker` object each time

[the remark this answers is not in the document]
migwellian: This reflects how it works in practice. You pass your listener to a `ProposalDistribution`, which repeatedly creates a `Proposal` and passes the listener in to it. A `Proposal` object is short-lived and represents just one proposal, and so the listener will typically listen to, and receive an event from, many proposals.

keanu-python/keanu/algorithm/_proposal_distribution.py
```diff
@@ -0,0 +1,37 @@
...
+            raise TypeError("Unknown Proposal Distribution type %s" % type_)
+        ctor = proposal_distribution_types[type_]
+        args = []
+        if type_ == "gaussian":
+            if sigma is None:
+                raise TypeError("Gaussian Proposal Distribution requires a value for sigma")
+            args.append(Tensor(sigma).unwrap())
+        else:
+            if sigma is not None:
+                raise TypeError('Parameter sigma is not valid unless type is "gaussian"')
```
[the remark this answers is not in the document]
migwellian: It pains me to remove useful error handling, even if it's "unpythonic". Shall we put it to a team vote?

keanu-python/keanu/algorithm/_proposal_distribution.py
```diff
@@ -0,0 +1,37 @@
...
+
+proposal_distribution_types: Dict[str, Callable] = {
+    "gaussian": k.jvm_view().GaussianProposalDistribution,
+    "prior": k.jvm_view().PriorProposalDistribution,
+}
+
+
+class ProposalDistribution(JavaObjectWrapper):
+
+    def __init__(self, type_: str, sigma: numpy_types = None, listeners: List[Any] = []) -> None:
```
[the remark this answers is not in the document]
migwellian: The default is set on the Java side: it's "prior", although that is likely to change since it's not a very good proposal. So I don't want to implement a default on the python side as that could lead to them getting out of sync.

My intention was to allow you to specify no type on the python side, and it just works - but I haven't implemented that yet, so I will do that still.

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -43,17 +53,46 @@ def sample(net: BayesNet,
...
+def build_sampling_algorithm(algo, proposal_distribution: Optional[str],
+                             proposal_distribution_sigma: Optional[numpy_types],
+                             proposal_listeners: List[proposal_listener_types]):
+    if algo != "metropolis":
+        if proposal_distribution is not None:
+            raise TypeError("Only Metropolis Hastings supports the proposal_distribution parameter")
+        if len(proposal_listeners) > 0:
+            raise TypeError("Only Metropolis Hastings supports the proposal_listeners parameter")
+
+    if (proposal_distribution is None and len(proposal_listeners) > 0):
```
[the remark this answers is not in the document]
migwellian: I put it here because I need to trigger the `if` clause 5 lines later. But I can do this more nicely.

keanu-python/tests/test_sampling.py
```diff
@@ -93,3 +101,48 @@ def set_starting_state(model: Model) -> None:
     model.temperature.set_value(model.temperature.sample())
     model.thermometer_one.set_value(model.thermometer_one.sample())
     model.thermometer_two.set_value(model.thermometer_two.sample())
+
+
+def test_can_get_acceptance_rates(net: BayesNet) -> None:
+    acceptance_rate_tracker = AcceptanceRateTracker()
+    latents = list(net.get_latent_vertices())
+    print(latents)
```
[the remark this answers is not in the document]
migwellian: No!

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -43,17 +53,46 @@ def sample(net: BayesNet,
...
+    samples = samples.dropCount(drop).downSampleInterval(down_sample_interval)
+    sample_iterator: JavaObject = samples.stream().iterator()
 
     return _samples_generator(sample_iterator, vertices_unwrapped)
 
 
+def build_sampling_algorithm(algo, proposal_distribution: Optional[str],
+                             proposal_distribution_sigma: Optional[numpy_types],
+                             proposal_listeners: List[proposal_listener_types]):
+    if algo != "metropolis":
```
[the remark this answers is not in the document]
migwellian: Not really, because the mechanism on the Java side is quite different: you construct a `ProposalDistribution` object and pass it to the `MetropolisHastingsBuilder`. There'd be nowhere to put a `NotImplementedError`.

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/proposal/AcceptanceRateTrackerTest.java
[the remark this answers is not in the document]
icantsayrural: Ah, I see. It makes sense if `Proposal` is short-lived.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/proposal/AcceptanceRateTracker.java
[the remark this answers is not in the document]
migwellian: Re: your first question - the `Proposal` object is short-lived and created by the `ProposalDistribution`, so the user doesn't have access to it.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/proposal/AcceptanceRateTracker.java
migwellian: As discussed I have implemented a custom `Counter` class instead of using `AtomicInteger` - for better performance.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/proposal/AcceptanceRateTracker.java
migwellian: I asked Charlie, and he said that the other possible option would be for the user to define a partition of their latent vertices. The existing 2 options could be thought of as special cases of this partition.

So whichever of these 3 options  we choose, it's a partition, i.e. a given vertex will exist in strictly 1 set. Therefore I agree with you - we can use individual vertices for the key rather than sets.

I've implemented it to key on `VertexId`.

keanu-python/keanu/algorithm/_proposal_distribution.py
```diff
@@ -0,0 +1,37 @@
...
+        if type_ not in proposal_distribution_types.keys():
+            raise TypeError("Unknown Proposal Distribution type %s" % type_)
```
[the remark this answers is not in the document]
migwellian: done

keanu-python/keanu/algorithm/sampling.py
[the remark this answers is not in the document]
migwellian: Actually I can't check in Java, for the same reasons I describe below (the Java mechanism is different: it's impossible to pass in listeners without having actively chosen the proposal distribution type).

I've decided that it's better to raise an exception here than to assume the type is `prior`.

PR #371, opened 2018-11-28
make formatApply run more quickly

migwellian:
because it needn't depend on auto-generating the python code

### DIFF
--- a/keanu-python/build.gradle
+++ b/keanu-python/build.gradle
@@ -111,7 +111,6 @@ task populateExclusionPaths {
 }
 
 task formatApply {
-    dependsOn preparePythonEnvironment
     dependsOn populateExclusionPaths
     doLast {
         exec {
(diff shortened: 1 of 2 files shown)


keanu-python/build.gradle
```diff
@@ -111,7 +111,7 @@ task populateExclusionPaths {
 }
 
 task formatApply {
-    dependsOn preparePythonEnvironment
+    dependsOn installPipenv
```
wesleyw72: Does it not also depend on `installDependencies`? Otherwise `yapf` won't be available to the pipenv.

migwellian: Yes it does - that's frustrating because `installDependencies` is slow.

I've had a look at the various `pipenv` commands but it doesn't appear possible make it run really fast (e.g. tell it to only upgrade dependencies if their versions have changed)

wesleyw72:
Certainly runs faster, nice one! However, it requires that you have setup the `pipenv` at some point in the past. I think this should be documented to make it clear to people who onboard onto the repo.

PR #377, opened 2018-12-02
Python traceplot

icantsayrural:
## Labels
- automatically setting labels of vertices in `Model` on call to `__setattr__`
- change traces/samples to be a dictionary mapping label to ndarrays (controversial change)
- waiting on `VertexLabel` wrapper to be merged to properly use the python API instead of using the Java one

## traceplot
- a helper method for plotting traces in `plots/traceplot.py`
- argument in `sample` and `generate_samples` for plotting/live plotting samples
- new dev dependencies: `matplotlib` and `pytest-mpl` (for image testing). Note `matplotlib` is not a hard requirement for `keanu`. See `keanu/plots/traceplot.py` import statement.

## gif of live plotting with streams
traceplot


keanu-python/tests/test_sampling.py
```diff
@@ -7,31 +7,35 @@
 from examples import thermometers
 from keanu import BayesNet, KeanuRandom, Model
 from keanu.algorithm import sample, generate_samples
-from keanu.vertex import Gamma, Exponential, Cauchy
+from keanu.vertex import Gamma, Exponential, Cauchy, Bernoulli
+from typing import Any
+import matplotlib
+matplotlib.use('Cairo')
```
migwellian: can you add a comment to say what this means?

migwellian: does this line have to come before the `import` on the next line?

keanu-python/tests/test_model.py
```diff
@@ -20,6 +20,10 @@ def test_to_bayes_net() -> None:
     assert m.sigma.get_id() in net_vertex_ids
     assert m.gaussian.get_id() in net_vertex_ids
 
+    assert m.mu.get_label().getQualifiedName() == "mu"
```
migwellian: here and elsewhere in the tests - you should use snake case `get_qualified_name()`

keanu-python/tests/test_sampling.py
```diff
@@ -58,7 +62,20 @@ def test_down_sample_interval(net: BayesNet) -> None:
         net=net, sample_from=net.get_latent_vertices(), draws=draws, down_sample_interval=down_sample_interval)
 
     expected_num_samples = draws / down_sample_interval
-    assert all(len(vertex_samples) == expected_num_samples for vertex_id, vertex_samples in samples.items())
+    assert all(len(vertex_samples) == expected_num_samples for label, vertex_samples in samples.items())
+
+
+@pytest.mark.mpl_image_compare(filename='test_sample_with_plot.png')
```
migwellian: very cool!

keanu-python/tests/test_sampling.py
```diff
@@ -84,8 +101,25 @@ def test_iter_returns_same_result_as_sample(algo: str) -> None:
     samples_dataframe = pd.DataFrame()
     [samples_dataframe.append(pd.DataFrame(list(next_sample.items()))) for next_sample in islice(iter_samples, draws)]
 
-    for vertex_id in samples_dataframe:
-        np.testing.assert_almost_equal(samples_dataframe[vertex_id].mean(), np.average(samples[vertex_id]))
+    for vertex in samples_dataframe:
```
migwellian: `vertex` is actually a `vertex_label`, right? (A string).

keanu-python/tests/test_sampling.py
```diff
@@ -84,8 +101,25 @@ def test_iter_returns_same_result_as_sample(algo: str) -> None:
...
+    with Model() as m:
+        m.exp = Exponential(np.ones((2, 2)))
+        m.exp.observe(np.array([[3., 2.], [0., 1.]]))
+        m.bernoulli = Bernoulli(m.exp)
+    net = m.to_bayes_net()
+
+    fig, ax = plt.subplots(1, 1, squeeze=False)
+    samples = generate_samples(net=net, sample_from=net.get_observed_vertices(), live_plot=True, refresh_every=5, ax=ax)
+
+    for sample in islice(samples, 5):
```
migwellian: could you parametrize this test on the number of steps that you take in the slice?

keanu-python/tests/test_traceplot.py
```diff
@@ -0,0 +1,44 @@
+from keanu.plots import traceplot
+from keanu.vartypes import sample_types
+from numpy import array
+from numpy.testing import assert_array_equal
+import pytest
+from typing import Any
+import matplotlib
+matplotlib.use('Cairo')
```
migwellian: same comments as in previous file for this line

keanu-python/tests/test_traceplot.py
```diff
@@ -0,0 +1,44 @@
...
+
+@pytest.fixture
+def trace() -> sample_types:
+    return {
+        "gamma": [array([[1., 2.], [3., 4.]]), array([[2., 3.], [4., 5.]])],
+        "gaussian": [array([[0.1, 0.2], [0.3, 0.4]]), array([[0.2, 0.3], [0.4, 0.5]])]
+    }
+
+
+def test_traceplot_returns_axeplot_with_correct_data(trace: sample_types) -> None:
```
migwellian: typo axplot?

keanu-python/build.gradle
```diff
@@ -54,10 +57,7 @@ task pytest {
             commandLine 'pipenv', 'run', 'pytest', 'tests/test_lorenz.py'
         }
         exec {
-            commandLine 'pipenv', 'run', 'pytest', '--ignore=tests/test_gradient_optimization.py', '--ignore=tests/test_lorenz.py'
-        }
-        exec {
-            commandLine 'pipenv', 'run', 'pytest', '--doctest-modules', 'keanu'
+            commandLine 'pipenv', 'run', 'pytest', '--mpl', '--mpl-baseline-path=baseline', '-p', 'no:warnings', '--ignore=tests/test_gradient_optimization.py', '--ignore=tests/test_lorenz.py'
```
migwellian: I think `baseline/` ought to live underneath `tests/`

keanu-python/tests/test_sampling.py
migwellian: `test_iter_with_live_plot.png` is actually identical to `test_sample_with_live_plot.png`, so you don't need both

keanu-python/tests/test_sampling.py
```diff
@@ -58,7 +62,20 @@ def test_down_sample_interval(net: BayesNet) -> None:
...
+@pytest.mark.mpl_image_compare(filename='test_sample_with_plot.png')
+def test_sample_with_plot() -> Any:
+    with Model() as m:
+        m.exp = Exponential(np.ones((2, 2)))
+        m.exp.observe(np.array([[3., 2.], [0., 1.]]))
+        m.bernoulli = Bernoulli(m.exp)
+    net = m.to_bayes_net()
+
+    fig, ax = plt.subplots(1, 1, squeeze=False)
+    sample(net=net, sample_from=net.get_observed_vertices(), draws=5, plot=True, ax=ax)
```
migwellian: why is the graph constant? i.e. every sample is the same

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -25,40 +26,65 @@ def sample(net: BayesNet,
...
 
 def generate_samples(net: BayesNet,
                      sample_from: Iterable[Vertex],
                      algo: str = 'metropolis',
                      drop: int = 0,
-                     down_sample_interval: int = 1) -> sample_generator_types:
+                     down_sample_interval: int = 1,
+                     live_plot: bool = False,
+                     refresh_every: int = 100,
+                     ax: Any = None) -> sample_generator_types:
```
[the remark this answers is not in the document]
icantsayrural: I think the pythonic thing to do is silently do nothing

keanu-python/tests/test_sampling.py
icantsayrural: oops, forgot we could seed `KeanuRandom` in Python now.

keanu-python/tests/test_model.py
icantsayrural: can't do that yet, `get_label()` returns a raw Java Object still cause your `VertexLabel` stuff isn't merged yet

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -25,40 +26,65 @@ def sample(net: BayesNet,
...
         net.unwrap(), vertices_unwrapped, draws).drop(drop).downSample(down_sample_interval)
     vertex_samples = {
-        Vertex._get_python_id(vertex_unwrapped): list(
+        Vertex._get_python_label(vertex_unwrapped): list(
             map(Tensor._to_ndarray,
                 network_samples.get(vertex_unwrapped).asList())) for vertex_unwrapped in vertices_unwrapped
     }
 
+    if plot:
+        traceplot(vertex_samples, ax=ax)
```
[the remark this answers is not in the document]
icantsayrural: `traceplot` will just create a new one with all graphs stacked

keanu-python/tests/test_sampling.py
icantsayrural: not related to your comment, but this `samples_dataframe` is empty...

icantsayrural: fixed.

keanu-python/tests/test_sampling.py
icantsayrural: what do you want to test exactly?

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -25,40 +26,65 @@ def sample(net: BayesNet,
            algo: str = 'metropolis',
            draws: int = 500,
            drop: int = 0,
-           down_sample_interval: int = 1) -> sample_types:
+           down_sample_interval: int = 1,
+           plot: bool = False,
+           ax: Any = None) -> sample_types:
```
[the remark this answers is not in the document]
icantsayrural: I think the pythonic thing to do is silently do nothing

keanu-python/tests/test_sampling.py
migwellian: even just to test 2 different values (e.g. 1 and 5) would be nice. It's not a big deal though.

keanu-python/keanu/algorithm/sampling.py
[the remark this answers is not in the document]
migwellian: 😆

keanu-python/keanu/algorithm/sampling.py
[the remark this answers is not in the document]
migwellian: 😆

icantsayrural:
Windows has a different font so the diff I am getting is:

test_sampling-failed-diff

According to `pytest-mpl`:
> The default tolerance is 2, which is very strict. In some cases, you may want to relax this to account for differences in fonts across different systems.

So I'm setting the tolerance to match exactly what is reported on Appveyor (e.g. 11.328646396046912)

keanu-python/keanu/plots/traceplot.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def __integer_xaxis(ax: Any, x0: int, n: int) -> range:
+    x = range(x0, x0 + n)
+    ax.xaxis.set_major_locator(MaxNLocator(integer=True))
+    return x
+
+
+def __pause_for_crude_animation() -> None:
+    plt.pause(0.1)
```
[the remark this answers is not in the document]
icantsayrural: Not unless asked for in the future?

keanu-python/keanu/model.py
```diff
@@ -17,6 +17,8 @@ def __setattr__(self, k: str, v: Any) -> None:
             super(Model, self).__setattr__(k, v)
         else:
             self._vertices[k] = v
+            if isinstance(v, Vertex):
+                v.set_label(k)
```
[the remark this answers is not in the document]
icantsayrural: Hmmm .... good catch.

I think we should still change it? I can't imagine a scenario where it'd be unreasonably unfair for `Model` to change it?

keanu-python/tests/test_sampling.py
icantsayrural: So in my local dev Mac OS X environment, python is not installed as a framework .. and this threw a `RuntimeError` when I directly imported `matplotlib.pyplot` because matplotlib's default backend `MacOSX` for Mac users doesn't work with a flat Unix-y shared library.

Not a problem in our CI though, since Python is probably correctly installed.

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -25,40 +26,65 @@ def sample(net: BayesNet,
...
-            Vertex._get_python_id(vertex_unwrapped): Tensor._to_ndarray(network_sample.get(vertex_unwrapped))
+            Vertex._get_python_label(vertex_unwrapped): Tensor._to_ndarray(network_sample.get(vertex_unwrapped))
             for vertex_unwrapped in vertices_unwrapped
         }
+
+        if live_plot:
+            traces.append(sample)
+            if len(traces) % refresh_every == 0:
+                joined_trace = {k: [t[k] for t in traces] for k in sample.keys()}
+                if ax is None:
```
GeorgeNash: Why do we need the `if` here? We don't do it when we call `traceplot` from `sample`, we just pass the `None` `ax` straight to `traceplot`

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -25,40 +26,65 @@ def sample(net: BayesNet,
...
+                    ax = traceplot(joined_trace, x0=x0)
+                else:
+                    traceplot(joined_trace, ax=ax, x0=x0)
+                x0 += refresh_every
```
GeorgeNash: Can we give this a better name than `x0`, what is it? the number of x data points?

keanu-python/keanu/plots/traceplot.py
```diff
@@ -0,0 +1,63 @@
...
+    pass
+
+import numpy as np
+from keanu.vartypes import sample_types, numpy_types
+from typing import Any, List
+
+
+def traceplot(trace: sample_types, labels: List[str] = None, ax: Any = None, x0: int = 0) -> Any:
+    """
+    Plot samples values.
```
GeorgeNash: `Plot the samples from vertices`

icantsayrural:
`traceplot` does plot multiple lines. See `test_traceplot_generates_correct_image.png`!

keanu-python/keanu/algorithm/sampling.py
icantsayrural: It's so that between sample batches, we draw to the same window (by using the same `ax`)

keanu-python/keanu/algorithm/sampling.py
icantsayrural: `x0` is the index of the first data point in the batch. It's commonly used in other libs

keanu-python/tests/test_sampling.py
```diff
@@ -58,7 +62,17 @@ def test_down_sample_interval(net: BayesNet) -> None:
         net=net, sample_from=net.get_latent_vertices(), draws=draws, down_sample_interval=down_sample_interval)
 
     expected_num_samples = draws / down_sample_interval
-    assert all(len(vertex_samples) == expected_num_samples for vertex_id, vertex_samples in samples.items())
+    assert all(len(vertex_samples) == expected_num_samples for label, vertex_samples in samples.items())
+
+
+@pytest.mark.mpl_image_compare(filename='test_sampling_windows.png' if os.name == 'nt' else 'test_sampling_linux.png')
```
migwellian: This SO post suggests `platform.system() == 'Windows'` is a better catch-all check... not a big deal though, since all we really care about is that the test passes on CI.

https://stackoverflow.com/questions/1854/python-what-os-am-i-running-on

migwellian: Another optional suggestion: if we have several of these tests, it would be good to encapsulate the logic in a method `def construct_os_dependent_file_name(filename: str) -> str`

icantsayrural: Spoken to some people offline, we decided to not to image tests altogether for now. The reasoning was it's sufficient to test against axes. If we ever come to requiring e2e tests for this in the future, we can revisit this test method later.

PR #379, opened 2018-12-03
Ensure that scalar tensor shape isn't dropped during operations

ieva-improbable:
So far we have been treating scalar tensors (tensors with length one) as scalars and dropping their shape. For instance, when multiplying a double tensor `[[1, 2], [3, 4]]` (shape `[2, 2]`) with scalar tensor whose value is `1` and shape is `[1, 1, 1, 1]` we would drop the scalar tensor's shape and incorrectly return a tensor `[[1, 2], [3, 4]]` (shape `[2, 2]`). While the correct result is `[[[[1, 2], [3, 4]]]]` (shape `[1, 1, 2, 2]`).

This PR ensures that we take into account scalar tensor shapes when determining the shape of a result of tensor operations.

It also fixes small bugs with `TensorMultiply` and `MatrixMultiply` methods.

NOTE: due to the limitations of `ND4J` we can't do proper broadcasting, and operations on tensors with compatible shapes such as `[2, 1, 3]` and `[1, 2, 3]` are not supported. This was out of the scope of this task, but eventually we should make sure that broadcasting as described [here](https://docs.scipy.org/doc/numpy-1.15.0/user/basics.broadcasting.html) is fully supported.


keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -99,6 +175,250 @@ private static INDArray broadcastMinus(INDArray a, INDArray b) {
         }
     }
 
+    public static INDArray pow(INDArray left, INDArray right) {
```
GeorgeNash: Great work here but there's a lot of duplicated code! I think we can write some functions that generalise this rank 1 stuff for us. We're doing the same steps each time:

1) calculate shape
2) operation if left is length 1
3) operation if right is length 1
4) return it

I've had a quick go at one:

```
    private static INDArray lengthOne(INDArray left, INDArray right, BiFunction<INDArray, INDArray, INDArray> func, double leftScalar, double rightScalar) {
        long[] resultShape = Shape.broadcastOutputShape(left.shape(), right.shape());
        INDArray result;
        if (left.length() == 1) {
            result = func.apply(Nd4j.valueArrayOf(right.shape(), leftScalar), right);
        } else {
            result = func.apply(left, Nd4j.valueArrayOf(left.shape(), rightScalar));
        }
        return result.reshape(resultShape);
    }
```

That we can call with:

```
    public static INDArray pow(INDArray left, INDArray right) {
        if (left.length() == 1 || right.length() == 1) {
            return lengthOne(left, right, INDArrayShim::pow, left.getDouble(0), right.getDouble(0));
        }
        return Transforms.pow(left, right, false);
    }
```

Have a play around and see if we can extend it to the other ones!

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -99,6 +175,250 @@ private static INDArray broadcastMinus(INDArray a, INDArray b) {
...
+            new OldLessThanOrEqual(mask, right, mask, mask.length())
+        );
+        return mask;
+    }
+
+    private static boolean shapeAIsSmallerThanShapeB(long[] shapeA, long[] shapeB) {
+        if (shapeA.length != shapeB.length) {
+            return shapeA.length < shapeB.length;
+        }
+        for (int ind = 0; ind <shapeA.length; ind++) {
```
GeorgeNash: nit: `ind < shapeA.length`

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -27,17 +35,52 @@
...
 public class INDArrayShim {
 
+    private static final DataBuffer.Type BUFFER_TYPE = DataBuffer.Type.DOUBLE;
+
+    /*
+     * We need to load ND4J in a separate thread as on load it sets the FTZ and DAZ flags in the processor for the
+     * thread that does the load.  This causes issues with Apache Math that makes use of Sub-normal values (in
+     * particular to initialisation values for the BrentOptimizer).
+     *
+     * We have raised https://github.com/deeplearning4j/deeplearning4j/issues/6690 to address this
```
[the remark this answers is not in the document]
ieva-improbable: All credit goes to @christophernorth and @migwellian! I simply moved it here to be reusable for all classes where we use ND4J.

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -27,17 +35,52 @@
...
+     */
+    public static void startNewThreadForNd4j() {
+        Thread nd4jInitThread = new Thread(new Runnable() {
```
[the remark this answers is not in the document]
ieva-improbable: Nice! Looks much better.

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -27,17 +35,52 @@
  */
 public class INDArrayShim {
 
+    private static final DataBuffer.Type BUFFER_TYPE = DataBuffer.Type.DOUBLE;
```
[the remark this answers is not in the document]
ieva-improbable: Oh, I wanted to ask you guys about it. Would we always use `DataBuffer.Type.DOUBLE`? Or should I keep it more open and pass the buffer type in for methods where it is required?

keanu-project/src/test/java/io/improbable/keanu/tensor/ScalarTensorShapePreservationTest.java
```diff
@@ -0,0 +1,236 @@
...
+
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+import io.improbable.keanu.tensor.intgr.IntegerTensor;
+import org.junit.Before;
+import org.junit.Test;
+import org.nd4j.linalg.api.shape.Shape;
+
+import static org.junit.Assert.assertArrayEquals;
+
+public class ScalarTensorShapePreservationTest {
```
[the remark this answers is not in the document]
ieva-improbable: It's for both double and integer tensors, so I left it outside of any subfolders.

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -48,13 +91,24 @@ private static INDArray broadcastMultiply(INDArray a, INDArray b) {
...
+        }  else if (left.length() == 1 || right.length() == 1) {
+        return scalarDivideWithPreservedShape(left, right);
+    } else {
+        return broadcastDivide(left, right);
+    }
+}
+
+    private static INDArray scalarDivideWithPreservedShape(INDArray a, INDArray b) {
+        if (b.length() != 1) {
+            return scalarMultiplyWithPreservedShape(b.rdiv(1.0), a);
```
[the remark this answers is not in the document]
ieva-improbable: Optimising much :) Will do!

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -393,45 +397,45 @@ public DoubleTensor exp() {
...
     public DoubleTensor minus(DoubleTensor that) {
-        if (that.isLengthOne()) {
+        if (that.isScalar()) {
             return this.minus(that.scalar());
-        } else if (this.isLengthOne()) {
+        } else if (this.isScalar()) {
             return that.unaryMinus().plusInPlace(this);
         } else {
-            return this.duplicate().minusInPlace(that);
+            return this.duplicate().minusInPlace(that.duplicate());
```
GeorgeNash: Why do we need the `duplicate` on `that`?

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/ScalarDoubleTensor.java
```diff
@@ -154,20 +154,20 @@ public DoubleTensor times(double that) {
     }
 
     @Override
-    public DoubleTensor matrixMultiply(DoubleTensor value) {
-        if (value.isLengthOne()) {
-            return value.times(value);
+    public DoubleTensor matrixMultiply(DoubleTensor doubleTensor) {
```
GeorgeNash: Can we call it `that` rather than `doubleTensor`?

GeorgeNash: Nice work finding this bug btw!

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/ScalarDoubleTensor.java
```diff
@@ -154,20 +154,20 @@ public DoubleTensor times(double that) {
...
 
     @Override
-    public DoubleTensor tensorMultiply(DoubleTensor value, int[] dimsLeft, int[] dimsRight) {
-        if (value.isLengthOne()) {
+    public DoubleTensor tensorMultiply(DoubleTensor doubleTensor, int[] dimsLeft, int[] dimsRight) {
+        if (doubleTensor.isLengthOne()) {
             if (dimsLeft.length > 1 || dimsRight.length > 1 || dimsLeft[0] != 0 || dimsRight[0] != 0) {
                 throw new IllegalArgumentException("Tensor multiply sum dimensions out of bounds for scalar");
             }
-            return value.times(value);
```
GeorgeNash: another good find!

keanu-project/src/main/java/io/improbable/keanu/tensor/intgr/Nd4jIntegerTensor.java
```diff
@@ -151,22 +157,22 @@ public IntegerTensor pow(int exponent) {
 
     @Override
     public IntegerTensor minus(IntegerTensor that) {
-        return duplicate().minusInPlace(that);
+        return duplicate().minusInPlace(that.duplicate());
```
GeorgeNash: Why have we added `duplicate` on these?

ieva-improbable: Because we don't want the operation to be done in place, and we don't know which way around it will be applied at this point (as in, `this.minus(that)` or `that.minus(this)`).

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
ieva-improbable: Done it as far as I could.

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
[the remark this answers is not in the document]
GeorgeNash: I can't see any methods where you use it in `INDArrayShim`?

keanu-project/src/test/java/io/improbable/keanu/tensor/ScalarTensorShapePreservationTest.java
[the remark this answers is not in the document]
GeorgeNash: Oh I see - would you mind splitting it into a `double` and `integer` one? Easier to pin point any issues that way

keanu-project/src/test/java/io/improbable/keanu/tensor/ScalarTensorShapePreservationTest.java
```diff
@@ -0,0 +1,236 @@
...
+    public void setup() {
+        doubleTensor = DoubleTensor.create(new double[]{1, 2, 3, 4, 5, 6}, new long[]{2, 3});
+        scalarDoubleTensor = DoubleTensor.create(1.0, new long[]{1, 1, 1, 1});
+
+        intTensor = IntegerTensor.create(new int[]{5, 6, 7, 8, 9, 10}, new long[]{3, 2});
+        scalarIntTensor = IntegerTensor.create(1, new long[]{1, 1, 1});
+    }
+
+    @Test
+    public void doubleScalarMultiplicationPreservesShape() {
```
GeorgeNash: Can we also add in tests for `scalarDoubleTensor.times(doubleTensor)` etc? And `integerTensor.times(scalarIntTensor)`?

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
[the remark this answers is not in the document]
ieva-improbable: I'm currently passing it in as a parameter for several methods, and it's `DataBuffer.Type.DOUBLE` for all of them. So I was thinking I could just pull it out to a static variable, if we know that it will always be `DataBuffer.Type.DOUBLE`?

keanu-project/src/test/java/io/improbable/keanu/tensor/ScalarTensorShapePreservationTest.java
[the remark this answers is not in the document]
ieva-improbable: Done :)

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
```diff
@@ -96,7 +100,7 @@ public static Nd4jDoubleTensor arange(double start, double end, double stepSize)
     }
 
     static INDArray unsafeGetNd4J(DoubleTensor that) {
-        if (that.isLengthOne()) {
+        if (that.isScalar() || that instanceof ScalarDoubleTensor) {
```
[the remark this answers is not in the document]
ieva-improbable: I think there are some cases where a tensor is scalar but is not an instance of ScalarDoubleTensor or ScalarIntegerTensor. If that should never be the case, I can investigate where they come from and fix it (I think it can happen as a result of some operations).

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -27,17 +36,47 @@
...
+            nd4jInitThread.join();
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        }
+    }
+
     public static INDArray muli(INDArray left, INDArray right) {
         if (Arrays.equals(left.shape(), right.shape())) {
             return left.muli(right);
+        } else if (left.length() == 1 || right.length() == 1) {
```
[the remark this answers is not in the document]
ieva-improbable: Rewrote, does this make more sense now?

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -99,6 +171,233 @@ private static INDArray broadcastMinus(INDArray a, INDArray b) {
...
+            }
+            return mask.reshape(resultShape);
+        }
+        Nd4j.getExecutioner().exec(
+            new OldLessThanOrEqual(mask, right, mask, mask.length())
+        );
+        return mask;
+    }
+
+    private static INDArray performOperationWithScalarTensor(INDArray left, INDArray right, BiFunction<INDArray, INDArray, INDArray> operation) {
```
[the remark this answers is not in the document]
ieva-improbable: Done :)

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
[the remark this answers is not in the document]
ieva-improbable: Meh, decided to get rid of it.

keanu-project/src/main/java/io/improbable/keanu/tensor/intgr/Nd4jIntegerTensor.java
```diff
@@ -81,7 +87,7 @@ public static Nd4jIntegerTensor zeros(long[] shape) {
     }
 
     static INDArray unsafeGetNd4J(IntegerTensor that) {
-        if (that.isLengthOne()) {
+        if (that.isScalar() || that instanceof ScalarIntegerTensor) {
```
GeorgeNash: if `that` is an instance of `ScalarIntegerTensor` shouldn't it return `true` on `that.isScalar()`?

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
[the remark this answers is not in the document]
gordoncaleb: I think the check should still be for isLengthOne(). That's identifies the special case completely and doesn't couple the Nd4jDoubleTensor to the ScalarDoubleTensor.

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -27,16 +36,43 @@
...
+        } else if (left.length() == 1) {
+            return scalarMultiplyWithPreservedShape(left, right);
+        } else if (right.length() == 1) {
+            return scalarMultiplyWithPreservedShape(right, left);
         } else {
             return broadcastMultiply(left, right);
         }
     }
 
+    private static INDArray scalarMultiplyWithPreservedShape(INDArray a, INDArray b) {
```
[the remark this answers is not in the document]
ieva-improbable: But I don't feel super strongly about it, so could rename them to `left` and `right`.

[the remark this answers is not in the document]
ieva-improbable: Re method naming - that's super long, but if we're fine with long names I think it's a good idea. Makes it much clearer.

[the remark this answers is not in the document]
ieva-improbable: Hmm, I think I'll rename the variables to `scalar` and `tensor`. Would that be clearer?

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -99,6 +165,233 @@ private static INDArray broadcastMinus(INDArray a, INDArray b) {
...
+        }
+        return left.gt(right);
+    }
+
+    public static INDArray lessThanOrEqual(INDArray left, INDArray right) {
+        if (left.length() == 1 || right.length() == 1) {
+            long[] resultShape = Shape.broadcastOutputShape(left.shape(), right.shape());
+            INDArray result;
+            if (left.length() == 1) {
+                left = Nd4j.valueArrayOf(right.shape(), left.getDouble(0));
```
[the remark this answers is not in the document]
ieva-improbable: Changed it to assign straight to result and then do `Nd4j` executions on the result,

keanu-project/src/main/java/io/improbable/keanu/tensor/intgr/Nd4jIntegerTensor.java
ieva-improbable: Nope, scalars are rank 0, while `Scalar...Tensor`s can have any rank (a shape of `[1,1,...,1]`). It is confusing though, so maybe we should think about renaming `isScalar`?

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -99,6 +165,231 @@ private static INDArray broadcastMinus(INDArray a, INDArray b) {
...
+        INDArray result;
+        if (left.length() == 1) {
+            result = operation.apply(Nd4j.valueArrayOf(right.shape(), left.getDouble(0)), right);
+        } else {
+            result = operation.apply(left, Nd4j.valueArrayOf(left.shape(), right.getDouble(0)));
+        }
+        return result.reshape(resultShape);
+    }
+
+    private static boolean shapeAIsSmallerThanShapeB(long[] shapeA, long[] shapeB) {
```
migwellian: I would find this clearer:

    private static boolean shapeAIsSmallerThanShapeB(long[] shapeA, long[] shapeB) {
        if (shapeA.length == shapeB.length) {
            for (int ind = 0; ind < shapeA.length; ind++) {
                if (shapeA[ind] < shapeB[ind]) {
                    return true;
                }
            }
        }
        return shapeA.length < shapeB.length;
    }

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -99,6 +165,231 @@ private static INDArray broadcastMinus(INDArray a, INDArray b) {
...
+        if (shapeA.length != shapeB.length) {
+            return shapeA.length < shapeB.length;
+        }
+        for (int ind = 0; ind < shapeA.length; ind++) {
+            if (shapeA[ind] < shapeB[ind]) {
+                return true;
```
migwellian: so {1, 100} is smaller than {2, 1}, for example? Yet {1, 100} is larger than {1000}. It's a strange definition.

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -99,6 +165,231 @@ private static INDArray broadcastMinus(INDArray a, INDArray b) {
         }
     }
 
+    public static INDArray pow(INDArray left, INDArray right) {
```
migwellian: we can do more refactoring if we pass lamdba functions as arguments.

@gordoncaleb might not be keen on the idea though, for performance?

keanu-project/src/main/java/io/improbable/keanu/tensor/intgr/Nd4jIntegerTensor.java
migwellian: but then shouldn't the `duplicate()` happen inside `minusInPlace`?

keanu-project/src/test/java/io/improbable/keanu/tensor/dbl/DoubleScalarTensorShapePreservationTest.java
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    @Test
+    public void doubleScalarSubtractionPreservesShape() {
+        Tensor subtractionResult = doubleTensor.minus(scalarDoubleTensor);
+        resultShapeMatchesBroadcastShape(subtractionResult, doubleTensor, scalarDoubleTensor);
+    }
+
+    @Test
+    public void doubleScalarMultiplicationPreservesShapeReverse() {
```
migwellian: nit: put the two Multiplication tests next to each other, to match the pattern in the rest of the file

likewise for Addition, Division, Subtraction

keanu-project/src/test/java/io/improbable/keanu/tensor/intgr/IntegerScalarTensorShapePreservationTest.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    @Test
+    public void integerScalarSubtractionPreservesShape() {
+        Tensor subtractionResult = intTensor.minus(scalarIntTensor);
+        resultShapeMatchesBroadcastShape(subtractionResult, intTensor, scalarIntTensor);
+    }
+
+    @Test
+    public void integerScalarMultiplicationPreservesShapeReverse() {
```
migwellian: nit: put the two Multiplication tests next to each other, to match the pattern in the rest of the file

likewise for Addition, Division, Subtraction

keanu-project/src/main/java/io/improbable/keanu/tensor/intgr/Nd4jIntegerTensor.java
ieva-improbable: But we don't want the tensors duplicated if we just call `minusInPlace` directly, in that case we'd expect the operation to be done in place and the original arguments to be modified in the process. Correct me if I'm wrong though.

ieva-improbable: Oh, I just read up on the difference between in-place and not in-place operations, and I think I see what your saying. With ND4J's `x.subi(y)` you'd expect `x` to be modified but `y` to remain the same. My bad, I had misunderstood it previously. I'll change this around.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/Nd4jDoubleTensor.java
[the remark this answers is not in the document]
ieva-improbable: Changed it to `isLengthOne()`, good point.

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
ieva-improbable: Yup, I think that's exactly what I wanted to do and what Caleb talked me out of. We can have another chat about it though?

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
ieva-improbable: That's so much nicer, thanks!

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
ieva-improbable: Just shortened it a little bit.

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -99,6 +142,115 @@ private static INDArray broadcastMinus(INDArray a, INDArray b) {
...
+        return performOperationWithScalarTensorPreservingShape(left, right, INDArray::gt);
+    }
+
+    public static INDArray eq(INDArray left, INDArray right) {
+        return performOperationWithScalarTensorPreservingShape(left, right, INDArray::eq);
+    }
+
+    private static INDArray performOperationWithScalarTensorPreservingShape(INDArray left, INDArray right, BiFunction<INDArray, INDArray, INDArray> operation) {
+        if (left.length() == 1 || right.length() == 1) {
+            long[] resultShape = Shape.broadcastOutputShape(left.shape(), right.shape());
```
migwellian: this doesn't need to be declared until line 187

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -99,6 +142,115 @@ private static INDArray broadcastMinus(INDArray a, INDArray b) {
...
+            INDArray result;
+            if (left.length() == 1) {
+                result = operation.apply(Nd4j.valueArrayOf(right.shape(), left.getDouble(0)), right);
+            } else {
+                result = operation.apply(left, Nd4j.valueArrayOf(left.shape(), right.getDouble(0)));
+            }
+            return result.reshape(resultShape);
+        }
+        return operation.apply(left, right);
```
migwellian: the control flow is clearer if this is put inside an `else` clause

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -99,6 +142,115 @@ private static INDArray broadcastMinus(INDArray a, INDArray b) {
...
+                    baseTransformOpConstructor.apply(mask,
+                        valueArrayOf(mask.shape(), right.getDouble(0), bufferType),
+                        mask,
+                        mask.length()
+                    )
+                );
+            }
+            return mask.reshape(resultShape);
+        }
+        Nd4j.getExecutioner().exec(
```
migwellian: the control flow is clearer if this is put inside an `else` clause

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -99,6 +142,115 @@ private static INDArray broadcastMinus(INDArray a, INDArray b) {
...
+        }
+        return operation.apply(left, right);
+    }
+
+    @FunctionalInterface
+    interface QuadFunction<First, Second, Third, Fourth, Result> {
+        Result apply(First one, Second two, Third three, Fourth four);
+    }
+
+    public static INDArray executeNd4jTransformOpWithPreservedScalarTensorShape(INDArray mask, INDArray right, DataBuffer.Type bufferType, QuadFunction<INDArray, INDArray, INDArray, Long, BaseTransformOp> baseTransformOpConstructor) {
```
migwellian: can be private

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -99,6 +142,115 @@ private static INDArray broadcastMinus(INDArray a, INDArray b) {
...
+            return mask.reshape(resultShape);
+        }
+        Nd4j.getExecutioner().exec(
+            baseTransformOpConstructor.apply(mask, right, mask, mask.length())
+        );
+        return mask;
+    }
+
+    public static INDArray getGreaterThanMask(INDArray mask, INDArray right, DataBuffer.Type bufferType) {
+        QuadFunction<INDArray, INDArray, INDArray, Long, BaseTransformOp> oldGreaterThanConstructor = OldGreaterThan::new;
```
migwellian: this variable can be inlined. 

ditto for the other methods that call `executeNd4jTransformOpWithPreservedScalarTensorShape`

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -27,16 +37,37 @@
...
+        Thread nd4jInitThread = new Thread(() -> Nd4j.create(1));
+        nd4jInitThread.start();
+        try {
+            nd4jInitThread.join();
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        }
+    }
+
     public static INDArray muli(INDArray left, INDArray right) {
```
migwellian: `muli`, `divi`, `addi` and `subi` could also be deduped using lambdas (provided it doesn't impact perf)

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
```diff
@@ -27,16 +37,37 @@
...
+        } else if (left.length() == 1) {
+            return applyScalarTensorOperationWithPreservedShape(left, right, INDArray::muli);
+        } else if (right.length() == 1) {
+            return applyScalarTensorOperationWithPreservedShape(right, left, INDArray::muli);
         } else {
             return broadcastMultiply(left, right);
         }
     }
 
     private static INDArray broadcastMultiply(INDArray a, INDArray b) {
```
migwellian: The `broadcast*` methods could also be deduped using lambdas (provided it doesn't impact perf)

keanu-project/src/test/java/io/improbable/keanu/tensor/dbl/DoubleScalarTensorShapePreservationTest.java
migwellian: done

keanu-project/src/test/java/io/improbable/keanu/tensor/intgr/IntegerScalarTensorShapePreservationTest.java
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/tensor/INDArrayShim.java
migwellian: done

PR #385, opened 2018-12-04
Python autocorrelation

wesleyw72:
This accompanies the Java PR #378 and exposes autocorrelation in python.


keanu-python/tests/test_stats.py
```diff
@@ -0,0 +1,43 @@
+import numpy as np
+import pytest
+
+from keanu.vertex import Uniform
+from keanu.algorithm import sample
+from keanu import Model, stats
+
+
+def test_can_get_autocorrelation():
```
migwellian: please add a type annotation `-> None` to each test. This will cause MyPy to pick them up.

You will then have some new MyPy errors which Joanne or I can help you with - you'll need to add some more methods to the numpy stub.

keanu-python/keanu/stats.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def autocorrelation(arg: List[numpy_types]) -> ndarray:
+    if not issubdtype(arg[0].dtype, floating):
+        raise ValueError('Autocorrelation must be run on a list of numpy floating types.')
+    if not arg[0].shape == ():
+        raise ValueError("Autocorrelation must be run on a list of single element ndarrays.")
+    arg_array = hstack(arg)
+    autocorr = k.jvm_view().SampleStats.autocorrelation(k.to_java_array(arg_array))
+    return fromiter(autocorr, arg[0].dtype)
```
migwellian: nice neat implementation :-)

keanu-python/tests/test_stats.py
migwellian: Oh I see you've already added `hstack` to the stubs so I guess you know how to do it!

keanu-python/tests/test_stats.py
```diff
@@ -0,0 +1,43 @@
...
+    np.testing.assert_almost_equal(autocorr, expected)
+
+
+def test_can_get_autocorrelation_for_samples():
+    with Model() as m:
+        m.gaussian = Uniform(0, 1000)
+    net = m.to_bayes_net()
+    samples = sample(net=net, sample_from=list(net.get_latent_vertices()), algo="metropolis", draws=1000)
+    valid_key = list(samples.keys())[0]
+    autocorr = stats.autocorrelation(samples.get(valid_key))
```
migwellian: can you assert something about `autocorr`?

keanu-python/tests/test_stats.py
wesleyw72: Sorry, I had added them and forgot to push. Should be there now.

keanu-python/tests/test_stats.py
wesleyw72: I could do a similar assertion that I do in the Java side. I assert that for a randomly generated sequence, the autocorrelation is close to 0. I am wondering it's worth repeating that test here though?

keanu-python/tests/test_stats.py
```diff
@@ -0,0 +1,70 @@
...
+    np.testing.assert_almost_equal(autocorr, expected)
+
+
+def test_can_get_autocorrelation_for_samples() -> None:
+    with Model() as m:
+        m.uniform = Uniform(0, 1000)
+    net = m.to_bayes_net()
+    samples = sample(net=net, sample_from=net.get_latent_vertices(), algo="metropolis", draws=1000)
+    valid_key = list(samples.keys())[0]
+    autocorr = stats.autocorrelation(samples.get(valid_key))  # type: ignore
```
icantsayrural: nit: might be a bit silly, it'd be nice to add some assertions on the return type and value (e.g. autocorrelation of lag 0 is greater than or equal to the rest).

keanu-python/keanu/stats.py
```diff
@@ -0,0 +1,21 @@
...
+
+k = KeanuContext()
+
+java_import(k.jvm_view(), "io.improbable.keanu.algorithms.SampleStats")
+
+
+def autocorrelation(arg: List[numpy_types]) -> ndarray:
+    if not issubdtype(arg[0].dtype, floating):
+        raise ValueError('Autocorrelation must be run on a list of numpy floating types.')
+    if not arg[0].shape == ():
```
icantsayrural: This seems too restrictive. Maybe we should do something like allowing the user to specify the index instead (with a sensible default)?

keanu-python/keanu/stats.py
```diff
@@ -0,0 +1,21 @@
...
+from .context import KeanuContext
+
+k = KeanuContext()
+
+java_import(k.jvm_view(), "io.improbable.keanu.algorithms.SampleStats")
+
+
+def autocorrelation(arg: List[numpy_types]) -> ndarray:
+    if not issubdtype(arg[0].dtype, floating):
+        raise ValueError('Autocorrelation must be run on a list of numpy floating types.')
```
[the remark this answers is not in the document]
wesleyw72: I think there's cases where you can sample from discrete and boolean distributions though. I don't think casting these to floats would be right?

keanu-python/keanu/stats.py
```diff
@@ -0,0 +1,21 @@
...
+from keanu.vartypes import numpy_types
+from .context import KeanuContext
+
+k = KeanuContext()
+
+java_import(k.jvm_view(), "io.improbable.keanu.algorithms.SampleStats")
+
+
+def autocorrelation(arg: List[numpy_types]) -> ndarray:
+    if not issubdtype(arg[0].dtype, floating):
```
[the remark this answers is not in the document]
wesleyw72: Good point, we should check all of them in case users use this in a way that doesn't get the samples from our sampling. I will change this and the shape check.

keanu-python/tests/test_stats.py
```diff
@@ -0,0 +1,70 @@
...
+    expected = [1., 0.625, 0.27380952, -0.0297619, -0.26190476, -0.39880952, -0.41666667, -0.29166667]
+    np.testing.assert_almost_equal(autocorr, expected)
+
+
+def test_can_get_autocorrelation_for_samples() -> None:
+    with Model() as m:
+        m.uniform = Uniform(0, 1000)
+    net = m.to_bayes_net()
+    samples = sample(net=net, sample_from=net.get_latent_vertices(), algo="metropolis", draws=1000)
+    valid_key = list(samples.keys())[0]
```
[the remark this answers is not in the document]
wesleyw72: I added it to ensure that no exception was thrown when using autocorrelation on samples.

keanu-python/tests/test_stats.py
wesleyw72: I agree with the return type. I will assert that it is an `ndarray`. I am hesitant to add anymore value checks on the python side as we are already doing so on the Java side. It could unnecessarily make tests longer to run. Happy to be convinced otherwise though.

keanu-python/keanu/stats.py
wesleyw72: This is a good point, something I completely missed. I am working on adding an argument for the index, which will make it very similar to how it's done in Java.

keanu-python/tests/test_stats.py
wesleyw72: Added a check on the datatype. Check that it is an ndarray and that it is the same type as the data it is given.

keanu-python/keanu/shape_validation.py
```diff
@@ -0,0 +1,14 @@
+from typing import Tuple, List, Union
+
+
+def check_index_is_valid(shape: Tuple[int, ...], index: Union[Tuple[int], Tuple]) -> None:
```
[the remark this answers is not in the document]
wesleyw72: Not quite. The empty tuple wouldn't be satisfied by Tuple[int] but would be by Tuple. I needed this for dealing with scalars. However, `Tuple[int, ...]` has the functionality I needed so I have replaced with that.

keanu-python/tests/test_stats.py
```diff
@@ -0,0 +1,82 @@
...
+    np.testing.assert_almost_equal(actual, expected)
+
+
+def test_autocorr_returns_ndarray_of_correct_dtype() -> None:
+    with Model() as m:
+        m.uniform = Uniform(0, 1000)
+    net = m.to_bayes_net()
+    samples = sample(net=net, sample_from=net.get_latent_vertices(), algo="metropolis", draws=1000)
+    valid_key = list(samples.keys())[0]
+    autocorr = stats.autocorrelation(samples.get(valid_key))  # type: ignore
```
migwellian: Please add a comment in the code that says why you need to `type: ignore` here

keanu-python/tests/test_stats.py
```diff
@@ -0,0 +1,82 @@
...
+    assert type(autocorr) == np.ndarray
+    assert autocorr.dtype == samples.get(valid_key)[0].dtype  # type: ignore
```
migwellian: Please add a comment in the code that says why you need to `type: ignore` here

keanu-python/tests/test_stats.py
migwellian: It might be better to do this, so it's more explicit:

    sample_: List[np.ndarray[Any]] = samples.get(  # type: ignore # dropping Optional as we know this won't be None
        valid_key)
    autocorr = stats.autocorrelation(sample_)
    assert type(autocorr) == np.ndarray
    assert autocorr.dtype == sample_[0].dtype

migwellian: Actually there's a MyPy config setting `--no-strict-optional` we can use so you don't need a `type: ignore` at all.

That would imply we're being less strict in our type checking, but maybe we need to make this compromise so that it doesn't slow down our python development. 

Deets at https://mypy.readthedocs.io/en/latest/kinds_of_types.html#optional-types-and-the-none-type

migwellian: OK, this is better - you can simply check for `None` and you don't need any `type: ignore`:

    sample_ = samples.get(valid_key)
    if sample_ is not None:
        autocorr = stats.autocorrelation(sample_)
        assert type(autocorr) == np.ndarray
        assert autocorr.dtype == sample_[0].dtype

migwellian: Neat - even this works:

    sample_ = samples.get(valid_key)
    assert sample_ is not None

    autocorr = stats.autocorrelation(sample_)
    assert type(autocorr) == np.ndarray
    assert autocorr.dtype == sample_[0].dtype

wesleyw72: Nice, thanks for finding that. Have used your last solution.

keanu-python/tests/test_stats.py
wesleyw72: Fixed by your solution on other thread.

Issue #342, opened 2018-11-14
Converting VertexSamples to DoubleTensor


johannespetrat:
At the moment it is quite hard to work with samples coming from MH and NUTS because they are in the `NetworkSamples` (or `VertexSamples`) format. `VertexSamples` has an `asList()` method. Can we please also have a `asTensor` method?

GeorgeNash:
You can get a `List<DoubleTensor>` of the samples by doing: `List<DoubleTensor> samples = networkSamples.getDoubleTensorSamples(probabilityA).asList();`

What behaviour would you want from an `asTensor` method? Let's say you have a vertex of shape `2x2` and you've taken `n` samples. Would `asTensor` return a tensor of shape `2x2xn`?

GeorgeNash:
Merged and closing

PR #386, opened 2018-12-06
Add asTensor method to IntegerVertexSamples and DoubleVertexSamples.

icantsayrural:
For request #342.

Also:
- Rename IntegerTensorVertexSamples to IntegerVertexSamples
- Make sliceAlongDimension a default method for all NumberTensors
- Add some tests for IntegerVertexSamples (most of which are copied from DoubleVertexSamples)
- Add stack


keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/DoubleVertexSamplesTest.java
```diff
@@ -58,4 +56,20 @@ public void doesCalculateVariance() {
         );
         assertThat(variances, allCloseTo(1e-8, expectedValues));
     }
+
+    @Test
+    public void canGetSamplesAsTensor() {
+        List<DoubleTensor> samplesAsList = samples.asList();
+
+        DoubleTensor samplesAsTensor = samples.asTensor();
+        List<DoubleTensor> samplesAsTensorSliced = samplesAsTensor.sliceAlongDimension(0, 0, samplesAsTensor.getShape()[0]);
```
wesleyw72: You can get this list by doing         
`List<DoubleTensor> samplesAsTensorSliced = samplesAsTensor.split(0,1,2,3,4);`

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/IntegerVertexSamplesTest.java
```diff
@@ -0,0 +1,81 @@
...
+        int mode = samples.getModeAtIndex(0, 1);
+        assertThat(mode, equalTo(16));
+    }
+
+    @Test
+    public void canGetSamplesAsTensor() {
+        List<IntegerTensor> samplesAsList = samples.asList();
+
+        IntegerTensor samplesAsTensor = samples.asTensor();
+        List<IntegerTensor> samplesAsTensorSliced = samplesAsTensor.sliceAlongDimension(0, 0, samplesAsTensor.getShape()[0]);
```
wesleyw72: Similar to my point above, you can use split to do this.

keanu-project/src/main/java/io/improbable/keanu/tensor/NumberTensor.java
```diff
@@ -77,6 +79,16 @@
     T setWithMask(T mask, N value);
 
     T apply(Function<N, N> function);
+
+    default List<T> sliceAlongDimension(int dimension, long indexStart, long indexEnd) {
+        List<T> slicedTensors = new ArrayList<>();
```
wesleyw72: I don't think this is necessary. See below, you can achieve what you're doing here by using `split`.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertexSamples.java
```diff
@@ -39,4 +39,7 @@ public DoubleTensor getVariances() {
             .divInPlace(samples.size()-1);
     }
 
+    public DoubleTensor asTensor() {
+        return DoubleTensor.concat(samples.toArray(new DoubleTensor[0]));
```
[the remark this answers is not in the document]
icantsayrural: Actually, Intellij complained if I used a presized array 😆 

> In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow. However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version. Also passing pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray call which may result in extra nulls at the end of the array, if the collection was concurrently shrunk during the operation.

>

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/IntegerVertexSamplesTest.java
icantsayrural: Not implemented for Integer tensors either, so I don't think we gain much? And I think slicing the entire first dimension makes more sense?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertexSamples.java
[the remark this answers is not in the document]
wesleyw72: Oh nice, I did not know this!

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/IntegerVertexSamplesTest.java
wesleyw72: Good point. I am not keen on adding `sliceAlongDimension` to `NumberTensor` at the moment if it only is being used in this test. See what everyone else thinks but I think it would be better for the `sliceAlongDimension` to sit with the test.

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/DoubleVertexSamplesTest.java
```diff
@@ -14,10 +14,10 @@
...
 public class DoubleVertexSamplesTest {
     List<DoubleTensor> values = ImmutableList.of(
-        DoubleTensor.create(new double[]{0, 16, 4}, new long[]{1, 3}),
-        DoubleTensor.create(new double[]{-4, -8, 4}, new long[]{1, 3}),
-        DoubleTensor.create(new double[]{8, -4, 12}, new long[]{1, 3}),
-        DoubleTensor.create(new double[]{4, 4, 8}, new long[]{1, 3})
+        DoubleTensor.create(new double[]{0, 16, 4}, 1, 3),
+        DoubleTensor.create(new double[]{-4, -8, 4}, 1, 3),
+        DoubleTensor.create(new double[]{8, -4, 12}, 1, 3),
+        DoubleTensor.create(new double[]{4, 4, 8}, 1, 3)
```
ieva-improbable: Just curious, why have you changed this? I think I find explicitly creating an array for shape more readable (but it's super subjective, of course).

icantsayrural: Yeah, I don't have a strong opinion on it - I kind of did it because intellij flagged is redundant

ieva-improbable: Oh, ok, I thought that there might be some more IntelliJ wisdom going on :)

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/IntegerVertexSamplesTest.java
migwellian: That makes sense @wesleyw72 .

`sliceAlongDimension` is currently only used in one place (`Multinomial`). Can we replace that with a `split` too?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertexSamples.java
GeorgeNash: Very close but I don't think this behaviour is quite right. In your test we are doing an `asTensor` on 4 samples of shape `1x3`. This should return a `4x1x3` but instead we're getting a `4x3`. Sounds like concat is not adding on an extra dimension? You may have to do a reshape after the concat.

icantsayrural: Done.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
```diff
@@ -92,6 +93,21 @@ static DoubleTensor scalar(double scalarValue) {
         return new ScalarDoubleTensor(scalarValue);
     }
 
+    /**
+     * @param toPile    an array of DoubleTensor
+     * @return  a DoubleTensor with toPile joined along a new dimension 0
```
migwellian: An example here would be helpful

keanu-project/src/main/java/io/improbable/keanu/tensor/intgr/IntegerTensor.java
```diff
@@ -66,6 +67,22 @@ static IntegerTensor scalar(int scalarValue) {
         return new ScalarIntegerTensor(scalarValue);
     }
 
+    /**
+     * @param toPile    an array of IntegerTensor
+     * @return  an IntegerTensor with toPile joined along a new dimension 0
```
migwellian: An example here would be helpful

migwellian: and/or an explanation of how this differs from `concat` (one creates an extra dimension, the other doesn't, right?)

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
migwellian: and/or an explanation of how this differs from `concat` (one creates an extra dimension, the other doesn't, right?)

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
```diff
@@ -92,6 +93,21 @@ static DoubleTensor scalar(double scalarValue) {
...
+     */
+    static DoubleTensor pile(DoubleTensor... toPile) {
```
migwellian: is `pile` standard terminology? Is this what's called `stack` in TensorFlow?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertexSamples.java
```diff
@@ -39,4 +39,7 @@ public DoubleTensor getVariances() {
             .divInPlace(samples.size()-1);
     }
 
+    public DoubleTensor asTensor() {
+        return DoubleTensor.pile(samples.toArray(new DoubleTensor[0]));
```
migwellian: I think we agreed previously (I think at your instigation??) that `new DoubleTensor[0]` is esoteric and it's clearer to do `samples.stream().toArray(DoubleTensor[]::new)`

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/IntegerVertexSamplesTest.java
```diff
@@ -0,0 +1,88 @@
...
+        for (long i = 0; i < samplesAsTensor.getShape()[0]; i++) {
+            samplesAsTensorSliced.add(samplesAsTensor.slice(0, i));
+        }
+
+        for (int i = 0; i < samplesAsList.size(); i++) {
+            IntegerTensor sampleFromList = samplesAsList.get(i);
+            IntegerTensor sampleFromTensor = samplesAsTensorSliced.get(i);
+
+            assertThat(sampleFromList.getShape(), equalTo(sampleFromTensor.getShape()));
+            assertThat(sampleFromList.asFlatDoubleArray(), equalTo(sampleFromTensor.asFlatDoubleArray()));
```
migwellian: FYI there are `TensorMatchers` to help with these.

However in this case you can equivalently do `assertThat(sampleFromList, equalTo(sampleFromTensor));`

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
icantsayrural: I am calling it `pile` because it's nd4j calls it that for when you are stacking everything at dimension 0

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
icantsayrural: Yup. Sounds good.

keanu-project/src/main/java/io/improbable/keanu/tensor/intgr/IntegerTensor.java
icantsayrural: Yup. Sounds good.

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/IntegerVertexSamplesTest.java
icantsayrural: Good call.

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
icantsayrural: renamed it to `stack` and to take a dimension parameter so it behaves like `numpy.stack`.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertexSamples.java
```diff
@@ -39,4 +39,7 @@ public DoubleTensor getVariances() {
             .divInPlace(samples.size()-1);
     }
 
+    public DoubleTensor asTensor() {
+        return DoubleTensor.stack(0, samples.toArray(new DoubleTensor[0]));
```
migwellian: `samples.stream().toArray(DoubleTensor[]::new)` is preferable since `[0]` is misleading

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
```diff
@@ -235,4 +235,19 @@ public static boolean isLengthOne(long[] shape) {
...
 
+    /**
+     * Finds the absolute dimension for a shape
+     *
+     * @param dimension the negative or positive dimension to find the absolute of
+     * @param rank      the rank
+     *
+     * @return an absolute dimension from a shape
+     */
+    public static int getAbsoluteDimension(int dimension, int rank) {
```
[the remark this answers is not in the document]
GeorgeNash: yep

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
```diff
@@ -92,11 +99,47 @@ static DoubleTensor scalar(double scalarValue) {
...
+     * DoubleTensor.stack(0, A, B, C) gives DoubleTensor.ones(3, 4, 2)
+     * <p>
+     * DoubleTensor.stack(1, A, B, C) gives DoubleTensor.ones(4, 3, 2)
+     * <p>
+     * DoubleTensor.stack(2, A, B, C) gives DoubleTensor.ones(4, 2, 3)
+     * <p>
+     * DoubleTensor.stack(-1, A, B, C) gives DoubleTensor.ones(4, 2, 3)
+     */
+    static DoubleTensor stack(int dimension, DoubleTensor... toStack) {
+        long[] shape = toStack[0].getShape();
```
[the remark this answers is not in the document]
GeorgeNash: I think it's overkill to check the length of `toStack`, that's holding everyones hand a bit to much. Good point about the same shape, I've added that to the java doc of `toStack`

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
```diff
@@ -92,11 +99,47 @@ static DoubleTensor scalar(double scalarValue) {
...
+     * <p>
+     * DoubleTensor.stack(0, A, B, C) gives DoubleTensor.ones(3, 4, 2)
+     * <p>
+     * DoubleTensor.stack(1, A, B, C) gives DoubleTensor.ones(4, 3, 2)
+     * <p>
+     * DoubleTensor.stack(2, A, B, C) gives DoubleTensor.ones(4, 2, 3)
+     * <p>
+     * DoubleTensor.stack(-1, A, B, C) gives DoubleTensor.ones(4, 2, 3)
+     */
+    static DoubleTensor stack(int dimension, DoubleTensor... toStack) {
```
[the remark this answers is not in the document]
GeorgeNash: It now checks as part of the `getAbsoluteDimension`

Issue #388, opened 2018-12-10
Implement logging with a specific name rather than using the root logger


robertperrotta:
Logging calls in `base.py` call the module functions directly, which log messages under the name "root." It would be better if `base.py` and `snipper_writer.py` used a named logger like in `context.py`.

https://github.com/improbable-research/keanu/blob/d2ca8302ba6db6e439f74b72c1d861f35eaa3f73/keanu-python/keanu/base.py#L26

https://github.com/improbable-research/keanu/blob/11d95ee051a3a80b62c9e1e3226195a5997f97a8/keanu-python/keanu/context.py#L30

migwellian:
Thanks for flagging this, Rob. I've implemented a fix so it will be shipped in the next release.

migwellian:
Fixed in Python release `0.0.16.dev2`

PR #389, opened 2018-12-11
Feature/spotless python rules

migwellian:
I've created a new Spotless configuration that checks our python source code for important problems:

1. using `logging` with the root logger
2. making a test without type annotations

Not intended for subjective code styling!


icantsayrural:
https://github.com/improbable-research/keanu/issues/388

PR #390, opened 2018-12-11
Bug/save vertex shape

### LINKED ISSUE #383: Include Shape in SaveLoad interface
**Describe the bug**
I have a vertex whose shape I define manually. This allows me to do a correct matrix multiplication when I define my Bayesian Network. Unfortunately, shape information is not stored in the protobuf saved version so when I save and load my graph, I get an error because the matrix multiplication doesn't like the shape of my vertex.

**Expected behavior**
I would like to be able to save and load Bayesian Networks which include vertices that are reliant on shape information without getting errors.

christophernorth:
The shapes for all vertices are now saved to the protobuf
Updated constructor annotations for all vertices that take an explicit shape so they can be constructed with the shape from the protobuf.
Added test case that covers the initial problem.
Resolves #383

### DIFF
--- a/keanu-project/src/test/java/io/improbable/keanu/util/io/JsonTest.java
+++ b/keanu-project/src/test/java/io/improbable/keanu/util/io/JsonTest.java
@@ -6,6 +6,7 @@ import com.google.common.io.Resources;
 import com.google.protobuf.util.JsonFormat;
 import io.improbable.keanu.KeanuSavedBayesNet;
 import io.improbable.keanu.network.BayesianNetwork;
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
 import io.improbable.keanu.vertices.VertexId;
 import io.improbable.keanu.vertices.dbl.DoubleVertex;
 import io.improbable.keanu.vertices.dbl.nonprobabilistic.ConstantDoubleVertex;
@@ -55,7 +56,7 @@ public class JsonTest {
         DoubleVertex mu = new ConstantDoubleVertex(0);
         DoubleVertex sigma = new ConstantDoubleVertex(new double[]{3.0, 4.0});
         DoubleVertex gaussianVertex = new GaussianVertex(mu, sigma);
-        gaussianVertex.observe(0.5);
+        gaussianVertex.observe(DoubleTensor.ones(2));
         gaussianVertex.setLabel("GaussianVertex");
         net = new BayesianNetwork(gaussianVertex.getConnectedGraph());
 
@@ -103,7 +104,7 @@ public class JsonTest {
         assertThat(gaussianVertex.getMu().getValue().scalar(), closeTo(0.0, 1e-10));
         assertThat(gaussianVertex.getSigma().getValue(0), closeTo(3.0, 1e-10));
         assertThat(gaussianVertex.getSigma().getValue(1), closeTo(4.0, 1e-10));
-        assertThat(gaussianVertex.getValue().scalar(), closeTo(0.5, 1e-10));
+        assertThat(gaussianVertex.getValue().scalar(), closeTo(1.0, 1e-10));
     }
 
     @Test
--- a/keanu-project/src/test/resources/jsonFiles/IncompatibleLabels.json
+++ b/keanu-project/src/test/resources/jsonFiles/IncompatibleLabels.json
@@ -39,7 +39,8 @@
                 "parentVertex": {
                     "id": "[1]"
                 }
-            }]
+            }],
+            "shape": ["2"]
         }]
     },
     "networkState": {
--- a/keanu-project/src/test/resources/jsonFiles/IncompatibleValueSpecified.json
+++ b/keanu-project/src/test/resources/jsonFiles/IncompatibleValueSpecified.json
@@ -39,7 +39,8 @@
                 "parentVertex": {
                     "id": "[1]"
                 }
-            }]
+            }],
+            "shape": ["2"]
         }]
     },
     "networkState": {
--- a/keanu-project/src/test/resources/jsonFiles/JsonModelOutput.json
+++ b/keanu-project/src/test/resources/jsonFiles/JsonModelOutput.json
@@ -40,7 +40,8 @@
                 "parentVertex": {
                     "id": "[1]"
                 }
-            }]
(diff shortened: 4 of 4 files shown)


keanu-project/src/test/java/io/improbable/keanu/network/ProtobufTest.java
```diff
@@ -458,12 +475,17 @@ private void checkSaveableVertex(Class<? extends Vertex> vertexClass) {
...
+            LoadVertexParam parameterAnnotation = parameter.getAnnotation(LoadVertexParam.class);
+            LoadShape shapeAnnotation = parameter.getAnnotation(LoadShape.class);
+            assertThat("Annotation has to be present on all Constructor params for class: " + vertexClass,
+                parameterAnnotation != null || shapeAnnotation != null);
+            if (parameterAnnotation != null) {
+                assertThat("Annotation can only be used once for class: " + vertexClass, requiredParameters,
+                    not(hasKey(parameterAnnotation.value())));
+                requiredParameters.put(parameterAnnotation.value(), parameter.getType());
+            } else {
+                assertThat("Shape Arguments must be long[]", parameter.getType().isAssignableFrom(long[].class));
```
wesleyw72: The name of this method (`getRequiredParamsAndCheckOnlyUsedOnce`) doesn't make it that clear to me that anything is being done with the shape annotation. Is there a way we could either make that more clear or split it out from here?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/SmoothUniformVertex.java
```diff
@@ -107,7 +106,9 @@ public SmoothUniformVertex(long[] tensorShape, double xMin, double xMax, double
         this(tensorShape, new ConstantDoubleVertex(xMin), new ConstantDoubleVertex(xMax), edgeSharpness);
     }
 
-    public SmoothUniformVertex(long[] tensorShape, DoubleVertex xMin, DoubleVertex xMax) {
+    public SmoothUniformVertex(@LoadShape long[] tensorShape,
```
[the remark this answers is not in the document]
christophernorth: 👍

keanu-project/src/test/java/io/improbable/keanu/network/ProtobufTest.java
```diff
@@ -94,6 +94,23 @@ public void youCanSaveAndLoadANetworkWithValues() throws IOException {
...
+        BayesianNetwork bayesNet = new BayesianNetwork(output.getConnectedGraph());
+
+        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+        ProtobufSaver saver = new ProtobufSaver(bayesNet);
+        saver.save(outputStream, false);
+
+        ByteArrayInputStream inputStream = new ByteArrayInputStream(outputStream.toByteArray());
+        ProtobufLoader loader = new ProtobufLoader();
+
+        BayesianNetwork readNet = loader.loadNetwork(inputStream);
```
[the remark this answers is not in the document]
christophernorth: I left that implicit as the shapes have to be "correct" due to the matrix multiply (as in that will assert if they're wrong).  I can add in an explicit check though as well.

keanu-project/src/test/java/io/improbable/keanu/network/ProtobufTest.java
christophernorth: Renamed the function

keanu-project/src/main/java/io/improbable/keanu/network/ProtobufLoader.java
```diff
@@ -213,13 +205,39 @@ private Vertex instantiateVertex(Class vertexClass,
...
+                                KeanuSavedBayesNet.Vertex vertex) {
+        LoadVertexParam paramAnnotation;
+        LoadShape shapeAnnotation;
+
+        if ((paramAnnotation = methodParameter.getAnnotation(LoadVertexParam.class)) != null) {
+            Object parameter = paramMap.get(paramAnnotation.value());
+            if (parameter == null) {
+                throw new IllegalArgumentException("Failed to create vertex due to missing parent: "
+                    + paramAnnotation.value());
+            }
```
migwellian: for consistency (with line 224) I think you should put an `else` here

keanu-project/src/main/java/io/improbable/keanu/network/ProtobufLoader.java
```diff
@@ -213,13 +205,39 @@ private Vertex instantiateVertex(Class vertexClass,
...
+            return parameter;
+        } else if ((shapeAnnotation = methodParameter.getAnnotation(LoadShape.class)) != null) {
```
migwellian: `shapeAnnotation` is never used

keanu-project/src/test/java/io/improbable/keanu/network/ProtobufTest.java
```diff
@@ -94,6 +94,23 @@ public void youCanSaveAndLoadANetworkWithValues() throws IOException {
         latentGaussianVertex.sample();
     }
 
+    @Test
```
migwellian: Do you need to add some tests for failure cases relating to shape? Similar to your other `loadFailsIf...` tests.

keanu-project/src/main/java/io/improbable/keanu/network/ProtobufLoader.java
christophernorth: Fixed

keanu-project/src/main/java/io/improbable/keanu/network/ProtobufLoader.java
christophernorth: Yeah - I just added it for consistency with the paramAnnotation case.  I guess it's not required - I'll remove.

keanu-project/src/test/java/io/improbable/keanu/network/ProtobufTest.java
christophernorth: There isn't really a failure case for shape - it can happily be left out entirely from the protobuf ( in fact that's what happens for scalars anyway).

PR #391, opened 2018-12-11
Feature/nuts

GeorgeNash:
Breaks NUTS out from the original mega-file

- Pulled out the `TreeBuilder`, `Leapfrog` and `Stepsize` classes
- Tests for `Leapfrog`
- Tests for `Stepsize`
- You can now record sampling statistics for all the samples in your NUTS run. Step size, log prob, mean tree acceptance prob and tree size. These are just stored as a simple dictionary with a list of doubles.

All issues I found while testing could be resolved through careful tweaking of the parameters (especially the relationship between `Target_acceptance_prob` and the tuning of the step size. This is especially apparent when starting in low likelihood areas). Hopefully the sampling statistics will help people see what's going wrong with their run and can tune their parameters accordingly.

Next...

- Tests for `TreeBuilder` (oh lord)
- Go through the applied dev diary for use cases that NUTS is currently failing on and add these to `NUTSTest`


keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Stepsize.java
```diff
@@ -0,0 +1,165 @@
...
+    private double adaptCount;
+    private double shrinkageTarget;
+
+    /**
+     * @param stepsize                  the step size
+     * @param targetAcceptanceProb      the target acceptance probability (lower target equates to a higher step size when tuning)
+     * @param adaptCount                the number of samples to adapt for
+     */
+    Stepsize(double stepsize, double targetAcceptanceProb, int adaptCount) {
+        this.averageAcceptanceProb = 0.;
```
migwellian: Java initialises `double` fields to `0.` implicitly

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Stepsize.java
```diff
@@ -0,0 +1,165 @@
...
+        this.targetAcceptanceProb = targetAcceptanceProb;
+        this.stepsize = stepsize;
+        this.logStepSizeFrozen = Math.log(stepsize);
+        this.adaptCount = adaptCount;
+        this.shrinkageTarget = Math.log(10 * stepsize);
```
migwellian: `shrinkageTarget` never changes - you could implement it as a private method instead of a field

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Stepsize.java
```diff
@@ -0,0 +1,165 @@
...
+ */
+public class Stepsize {
+
+    private static final double STABILISER = 10;
+    private static final double SHRINKAGE_FACTOR = 0.05;
+    private static final double TEND_TO_ZERO_EXPONENT = 0.75;
+    private static final double STARTING_STEPSIZE = 1;
+
+    private double stepsize;
+    private double averageTreeAcceptanceProb;
```
migwellian: can be a local variable instead of a field

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Stepsize.java
```diff
@@ -0,0 +1,165 @@
...
+    private double averageAcceptanceProb;
+    private double targetAcceptanceProb;
+    private double logStepSizeFrozen;
+    private double adaptCount;
+    private double shrinkageTarget;
```
migwellian: make `final`

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Stepsize.java
```diff
@@ -0,0 +1,165 @@
...
+    private static final double SHRINKAGE_FACTOR = 0.05;
+    private static final double TEND_TO_ZERO_EXPONENT = 0.75;
+    private static final double STARTING_STEPSIZE = 1;
+
+    private double stepsize;
+    private double averageTreeAcceptanceProb;
+    private double averageAcceptanceProb;
+    private double targetAcceptanceProb;
+    private double logStepSizeFrozen;
+    private double adaptCount;
```
migwellian: make `final`

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Stepsize.java
```diff
@@ -0,0 +1,165 @@
...
+
+    private static final double STABILISER = 10;
+    private static final double SHRINKAGE_FACTOR = 0.05;
+    private static final double TEND_TO_ZERO_EXPONENT = 0.75;
+    private static final double STARTING_STEPSIZE = 1;
+
+    private double stepsize;
+    private double averageTreeAcceptanceProb;
+    private double averageAcceptanceProb;
+    private double targetAcceptanceProb;
```
migwellian: make `final`

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Stepsize.java
```diff
@@ -0,0 +1,165 @@
...
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.diff.LogProbGradientCalculator;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Used by NUTS as an epsilon for the leap frog
+ */
+public class Stepsize {
```
migwellian: make this class package private?

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Stepsize.java
```diff
@@ -0,0 +1,165 @@
...
+    public static double findStartingStepSize(Map<VertexId, DoubleTensor> position,
+                                       Map<VertexId, DoubleTensor> gradient,
+                                       List<Vertex<DoubleTensor>> vertices,
+                                       List<Vertex> probabilisticVertices,
+                                       LogProbGradientCalculator logProbGradientCalculator,
+                                       double initialLogOfMasterP,
+                                       KeanuRandom random) {
+        double stepsize = STARTING_STEPSIZE;
+        Map<VertexId, DoubleTensor> momentums = new HashMap<>();
+        initializeMomentumForEachVertex(vertices, momentums, random);
```
migwellian: you can nuke this method and do

        Map<VertexId, DoubleTensor> momentums = vertices.stream()
            .collect(Collectors.toMap(
                Vertex::getId,
                v -> random.nextGaussian(v.getShape())
            ));

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Stepsize.java
```diff
@@ -0,0 +1,165 @@
...
+        Leapfrog leapfrog = new Leapfrog(position, momentums, gradient);
+        double pThetaR = initialLogOfMasterP - leapfrog.halfDotProductMomentum();
+
+        Leapfrog delta = leapfrog.step(vertices, logProbGradientCalculator, STARTING_STEPSIZE);
+
+        double probAfterLeapfrog = ProbabilityCalculator.calculateLogProbFor(probabilisticVertices);
+        double pThetaRAfterLeapFrog = probAfterLeapfrog - delta.halfDotProductMomentum();
+
+        double logLikelihoodRatio = pThetaRAfterLeapFrog - pThetaR;
+        double scalingFactor = logLikelihoodRatio > Math.log(0.5) ? 1 : -1;
```
migwellian: `signum` would be a better name, since it's always +1 or -1

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Stepsize.java
```diff
@@ -0,0 +1,165 @@
...
+
+    /**
+     * Adapts the step size based on the state of the tree and network after computing a sample
+     *
+     * @param tree          the balanced binary tree
+     * @param sampleNum     the number of samples that have been taken
+     *
+     * @return a new step size
+     */
+    public double adaptStepSize(TreeBuilder tree, int sampleNum) {
```
migwellian: I would find it clearer if you pull out another method so that this becomes:

    public double adaptStepSize(TreeBuilder tree, int sampleNum) {

        double logStepSize = logStepSizeFrozen;

        if (sampleNum < adaptCount) {
            logStepSize = updateLogStepSize(tree, sampleNum);
        }

        stepsize = Math.exp(logStepSize);
        return stepsize;
    }

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/NUTSTest.java
```diff
@@ -53,18 +97,19 @@ public void samplesContinuousPrior() {
 
         BayesianNetwork bayesNet = MCMCTestDistributions.createSumOfGaussianDistribution(20.0, 1.0, 46., 15.0);
 
-        int sampleCount = 6000;
+        int sampleCount = 5000;
         NUTS nuts = NUTS.builder()
             .adaptCount(sampleCount)
-            .maxTreeHeight(4)
+            .maxTreeHeight(10)
```
migwellian: This is now very slow... but we didn't want to mark this test as `@Slow` because it's the only non-slow one we have, that we run every time.

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/NUTSTest.java
```diff
@@ -25,6 +28,46 @@ public void setup() {
...
+
+        Statistics statistics = nuts.getStatistics();
+
+        List<Double> stepSize = statistics.get("stepSize");
+        List<Double> logProb = statistics.get("logProb");
+        List<Double> meanTreeAccept = statistics.get("meanTreeAccept");
+        List<Double> treeSize = statistics.get("treeSize");
+
+        Assert.assertTrue(stepSize.get(0) == initStepSize);
+        Assert.assertTrue(stepSize.get(1) == initStepSize);
```
migwellian: better to use Hamcrest, e.g.

        assertThat(stepSize, contains(initStepSize, initStepSize));

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/NUTSTest.java
```diff
@@ -25,6 +28,46 @@ public void setup() {
...
+        Assert.assertTrue(logProb.get(0) < 0);
+        Assert.assertTrue(logProb.get(1) < 0);
```
migwellian: assertThat(logProb, everyItem(lessThan(0.)));

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/NUTSTest.java
```diff
@@ -25,6 +28,46 @@ public void setup() {
...
+        Assert.assertTrue(meanTreeAccept.get(0) >= 0 && meanTreeAccept.get(0) <= 1);
+        Assert.assertTrue(meanTreeAccept.get(1) >= 0 && meanTreeAccept.get(1) <= 1);
```
migwellian: assertThat(meanTreeAccept, everyItem(both(greaterThanOrEqualTo(0.)).and(lessThanOrEqualTo(1.))));

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/NUTSTest.java
```diff
@@ -25,6 +28,46 @@ public void setup() {
...
+        Assert.assertTrue(treeSize.get(0) < Math.pow(2, maxTreeHeight));
+        Assert.assertTrue(treeSize.get(1) < Math.pow(2, maxTreeHeight));
```
migwellian: assertThat(treeSize, everyItem(lessThan(Math.pow(2, maxTreeHeight))));

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/StepsizeTest.java
```diff
@@ -0,0 +1,103 @@
...
+            gradient,
+            Arrays.asList(vertex),
+            bayesianNetwork.getLatentVertices(),
+            logProbGradientCalculator,
+            ProbabilityCalculator.calculateLogProbFor(vertices),
+            random
+        );
+
+        double startingEpsilon = 1.0;
+        Assert.assertTrue(startingStepsize < startingEpsilon);
```
migwellian: replace all `assertTrue` with hamcrest matchers for better error reporting

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/StepsizeTest.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Test
+    public void canFindLargeStartingStepsizeForLargeSpace() {
```
migwellian: lots of code duplication that could be refactored between `canFindSmallStartingStepsizeForSmallSpace` and `canFindLargeStartingStepsizeForLargeSpace`

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/StepsizeTest.java
```diff
@@ -0,0 +1,103 @@
...
+            0.65,
+            50
+        );
+
+        TreeBuilder mockedLessLikelyTree = mock(TreeBuilder.class);
+        when(mockedLessLikelyTree.getDeltaLikelihoodOfLeapfrog()).thenAnswer(i -> -50.);
+        when(mockedLessLikelyTree.getTreeSize()).thenAnswer(i -> 8.);
+        double adaptedStepSizeLessLikely = tune.adaptStepSize(mockedLessLikelyTree, 1);
+
+        Assert.assertTrue(adaptedStepSizeLessLikely < startingStepsize);
```
migwellian: I'm not sure how valuable this is, but you can verify the methods that were called on your mocks:

        verify(mockedLessLikelyTree).getDeltaLikelihoodOfLeapfrog();
        verify(mockedLessLikelyTree).getTreeSize();
        verifyNoMoreInteractions(mockedLessLikelyTree);

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/NUTSSampler.java
```diff
@@ -120,288 +114,47 @@ public void step() {
...
-            forward += forwardMinusBackward.times(momentumForward.get(latentId)).sum();
-            backward += forwardMinusBackward.timesInPlace(momentumBackward.get(latentId)).sum();
-        }
-
-        return (forward >= 0.0) && (backward >= 0.0);
+    private void recordSamplerStatistics() {
+        statistics.store("stepSize", stepsize.getStepsize());
+        statistics.store("logProb", tree.getLogOfMasterPAtAcceptedPosition());
+        statistics.store("treeSize", tree.getTreeSize());
+        statistics.store("meanTreeAccept", stepsize.getAverageAcceptanceProb());
```
christophernorth: I'm not a fan of using Strings as a key for these stats, it doesn't seem to buy much in terms of clarity and it's going to be quite costly to have to hash that every time you want to increment stats (yeah, it is optional for us to turn these on, but given they're useful it would be good to have them on by default IMO).

Perhaps instead each sampler should implement their own stats class that just has hard coded methods for each stat and then we have a statistics interface that converts that in to a format suitable for consumption (for instance in to this style of map) once when stats collection is complete?

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Stepsize.java
GeorgeNash: Agreed but we've kept the terminology the same as from the paper to make it easier to compare

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/NUTS.java
```diff
@@ -79,6 +87,12 @@ public static NUTS withDefaultConfig(KeanuRandom random) {
     @Builder.Default
     private int maxTreeHeight = 10;
 
+    //Sets whether or not to save debug STATISTICS. The STATISTICS available are: Step size, Log Prob, Mean Tree Acceptance Prob, Tree Size.
+    @Getter
+    @Setter
+    @Builder.Default
```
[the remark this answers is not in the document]
GeorgeNash: Which other fields? I think we need Getter and Setter on them all

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Statistics.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public List<Double> get(String key) {
+        return statistics.get(key);
+    }
+
+    public Set<String> keys() {
+        return statistics.keySet();
+    }
+
+    public double average(String key) {
```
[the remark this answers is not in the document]
GeorgeNash: I use it a lot while debugging in the `script executor` thing - same for `keys`

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/TreeBuilder.java
```diff
@@ -0,0 +1,433 @@
...
+                                        List<Vertex> probabilisticVertices,
+                                        LogProbGradientCalculator logProbGradientCalculator,
+                                        final List<? extends Vertex> sampleFromVertices,
+                                        Leapfrog leapfrog,
+                                        double logU,
+                                        int buildDirection,
+                                        int treeHeight,
+                                        double epsilon,
+                                        double logOfMasterPMinusMomentumBeforeLeapfrog,
+                                        KeanuRandom random) {
```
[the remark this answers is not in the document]
GeorgeNash: you're right - but I'd rather keep these as parameters so that the function matches the pseudocode of the paper, or else it becomes very difficult (especially with the recursive function in there) to piece together what's going on

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/LeapfrogTest.java
```diff
@@ -0,0 +1,132 @@
...
+        Leapfrog leap = start.step(vertices, mockedGradientCalculator, epsilon);
+
+        Assert.assertEquals(1.0, leap.getPosition().get(aID).scalar(), 1e-6);
+        Assert.assertEquals(1.0, leap.getPosition().get(bID).scalar(), 1e-6);
+
+        Assert.assertEquals(2.5, leap.getMomentum().get(aID).scalar(), 1e-6);
+        Assert.assertEquals(1.5, leap.getMomentum().get(bID).scalar(), 1e-6);
+
+        Assert.assertEquals(1.0, leap.getGradient().get(aID).scalar(), 1e-6);
+        Assert.assertEquals(-1.0, leap.getGradient().get(bID).scalar(), 1e-6);
```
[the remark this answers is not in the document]
GeorgeNash: why? Doesn't that check that all the values of a tensor are close to each other and not the actual value?

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/NUTSTest.java
GeorgeNash: Reduced the number of samples & started it a little closer to the right answer

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/nuts/NUTSSampler.java
```diff
@@ -0,0 +1,198 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Algorithm 6: "No-U-Turn Sampler with Dual Averaging".
+ * The No-U-Turn Sampler: Adaptively Setting Path Lengths in Hamiltonian Monte Carlo
+ * https://arxiv.org/pdf/1111.4246.pdf
+ */
+public class NUTSSampler implements SamplingAlgorithm {
```
migwellian: make package private

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/nuts/Tree.java
```diff
@@ -0,0 +1,359 @@
...
+import static io.improbable.keanu.algorithms.mcmc.SamplingAlgorithm.takeSample;
+
+
+/**
+ * Used by NUTS as a balanced binary tree to compute and store information
+ * about leapfrogs that are taken forwards and backwards in space.
+ * <p>
+ * The tree is reset for each sample.
+ */
+public class Tree implements SaveStatistics {
```
migwellian: make package private

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/nuts/Tree.java
```diff
@@ -0,0 +1,359 @@
...
+        if (buildDirection == -1) {
+            currentTree.leapfrogBackward = otherHalfTree.leapfrogBackward;
+        } else {
+            currentTree.leapfrogForward = otherHalfTree.leapfrogForward;
+        }
+
+        return otherHalfTree;
+    }
+
+    private static Tree buildTree(List<Vertex<DoubleTensor>> latentVertices,
```
migwellian: Earlier I suggested pulling the builder out into a separate class, and you can put most of these arguments into its constructor to make it clearer what's invariant and what's not.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/NUTSSampler.java
GeorgeNash: Agreed - changed it to be an ENUM. The sampler defines the set of ENUM's that are passed to the stats class

keanu-project/src/main/java/io/improbable/keanu/algorithms/Statistics.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class Statistics {
+
+    private Map<Enum, List<Double>> statistics;
```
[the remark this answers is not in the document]
GeorgeNash: I think we should leave it as `Enum` in case we add metrics for `MetropolisHastings` etc

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/nuts/Leapfrog.java
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * Leapfrog performs a movement through physical space with the introduction of a momentum variable.
+ * This is required for sampling in Hamiltonian and NUTS.
+ */
+public class Leapfrog {
+
+    private final Map<VertexId, DoubleTensor> position;
+    private final Map<VertexId, DoubleTensor> momentum;
+    private final Map<VertexId, DoubleTensor> gradient;
```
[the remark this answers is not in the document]
GeorgeNash: I'd rather keep it as a `HashMap` for performance reasons

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/NUTS.java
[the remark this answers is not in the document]
GeorgeNash: My mistake, I thought you needed `Setter` for it to appear in the builder

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/nuts/Leapfrog.java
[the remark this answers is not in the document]
gordoncaleb: 👍

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/nuts/Tree.java
```diff
@@ -0,0 +1,359 @@
...
+     * A helper method to create a basic tree
+     *
+     * @param position                 the starting position
+     * @param momentum                 the starting momentum
+     * @param gradient                 the starting gradient
+     * @param initialLogOfMasterP      the initial log prob
+     * @param sampleAtAcceptedPosition the initial sample
+     * @return a basic tree
+     */
+    public static Tree createBasicTree(Map<VertexId, DoubleTensor> position,
```
gordoncaleb: Nit: createBaseTree since it's a recursive algo and this is the base case.

gordoncaleb: Oh wait. Maybe initialTree?

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Stepsize.java
migwellian: fair enough

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/TreeBuilder.java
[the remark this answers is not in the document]
migwellian: ok

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/Stepsize.java
migwellian: Oh OK. Carry on.

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/LeapfrogTest.java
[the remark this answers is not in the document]
migwellian: No... but since these are scalars anyway, this is not very important, so leave it.

keanu-project/src/main/java/io/improbable/keanu/algorithms/Statistics.java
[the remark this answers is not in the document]
migwellian: OK

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/nuts/Leapfrog.java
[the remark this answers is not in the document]
migwellian: OK

keanu-project/src/main/java/io/improbable/keanu/algorithms/Statistics.java
```diff
@@ -0,0 +1,39 @@
+package io.improbable.keanu.algorithms;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class Statistics {
```
migwellian: Sorry this comes late but I just had a thought.

Do we always expect that the lists should all be the same length? (because at each sample step we record every statistic).

If yes:

We could enforce that, but representing Statistics as a `List` of `Map`s instead of a `Map` of `List`s. But I expect you'll tell me that you did it this way round for better performance of `get` and `average`? 

Would it be useful to have a method that returns a boolean to tell you if all the lists are the same length? Or even a `getLength` method that throws if they're not the same length?

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/nuts/NUTS.java
```diff
@@ -31,6 +36,12 @@
     private static final int DEFAULT_ADAPT_COUNT = 1000;
     private static final double DEFAULT_TARGET_ACCEPTANCE_PROB = 0.65;
 
+    enum Metrics {
```
migwellian: This needs to be public, since in real world usage it will be referenced by calling code outside of this package.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/NUTS.java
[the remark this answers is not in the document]
migwellian: Can we still get rid of the last remaining `@Setter` (on the `random` field)? I don't understand why it's useful for this to be mutable.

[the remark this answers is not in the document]
migwellian: That question is probably for @gordoncaleb since he wrote a test for it, `canDefaultToSettingsInBuilderAndIsConfigurableAfterBuilding`

keanu-project/src/main/java/io/improbable/keanu/algorithms/Statistics.java
```diff
@@ -0,0 +1,39 @@
...
+
+    private Map<Enum, List<Double>> statistics = new HashMap<>();
+
+    public Statistics(Enum[] keys) {
```
christophernorth: Should we perhaps take a "length hint" in here to allow us to pre-allocate a sensible sized set of ArrayLists?  Obviously this won't work for streaming samples, but where we know up front how many samples we're generating etc - this might help us make a first guess at how many stats are coming in and stop us having to reallocate and copy stuff around a bit?

keanu-project/src/main/java/io/improbable/keanu/algorithms/Statistics.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class Statistics {
+
+    private Map<Enum, List<Double>> statistics = new HashMap<>();
```
christophernorth: Do we need this to be a Map now that we just have an enum provided at init time?  Presumably we could just make this an array and just mandate they use a "sensible" numbering scheme for their enums?

GeorgeNash: Agreed - that would make it faster. I just think the clarity of having these stored in a map is so much clearer for users who are debugging. It enables nice things like letting them see all the available stats really easily with `statistics.keySet()` and I think this is worth the cost. Remember this is only meant for `debugging` purposes

keanu-project/src/main/java/io/improbable/keanu/algorithms/Statistics.java
GeorgeNash: Honestly I think that for this debugging tool it's better to keep it simple. If the debug statistics becomes something that people begin to use on every single run then yes, I think we should begin doing stuff like this

PR #393, opened 2018-12-12
Out of bounds sampling fix

GeorgeNash:
NUTS has an algorithm to find a good value of the `stepsize` before it begins sampling. The way this algorithm works is by continually increasing it, taking a leapfrog and measuring the delta in log likelihood until it goes under some threshold. We were not resetting the value of the vertices to their initial value after running this algorithm to find a good starting size, meaning it was left in the state that produced the low likelihood. In the case of vertices with bounds (`half_gaussian`, `half_cauchy`) this low likelihood area happened to be a negative value.

This PR resets the value of the vertices to their starting value after finding a good value for the stepsize.

This PR also adds boundary checks for `dLogProb` for `halfGaussian` and `halfCauchy` and returns zero if it's invalid, we were previously just calling to the super no matter what the value of `x` was.

This can be merged in separately to the NUTS refactor.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/HalfCauchyVertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/HalfCauchyVertex.java
@@ -1,14 +1,14 @@
 package io.improbable.keanu.vertices.dbl.probabilistic;
 
-import java.util.Map;
-import java.util.Set;
-
 import io.improbable.keanu.annotation.ExportVertexToPythonBindings;
 import io.improbable.keanu.tensor.dbl.DoubleTensor;
 import io.improbable.keanu.vertices.LoadVertexParam;
 import io.improbable.keanu.vertices.Vertex;
 import io.improbable.keanu.vertices.dbl.DoubleVertex;
 import io.improbable.keanu.vertices.dbl.KeanuRandom;
+
+import java.util.Map;
+import java.util.Set;
 
 public class HalfCauchyVertex extends CauchyVertex {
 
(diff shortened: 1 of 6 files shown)


keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/HalfCauchyVertex.java
```diff
@@ -44,6 +48,19 @@ public double logProb(DoubleTensor value) {
         return Double.NEGATIVE_INFINITY;
     }
 
+    @Override
+    public Map<Vertex, DoubleTensor> dLogProb(DoubleTensor value, Set<? extends Vertex> withRespectTo) {
```
gordoncaleb: Was this just wrong before?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/HalfGaussianVertex.java
```diff
@@ -45,6 +49,19 @@ public double logProb(DoubleTensor value) {
         return Double.NEGATIVE_INFINITY;
     }
 
+    @Override
+    public Map<Vertex, DoubleTensor> dLogProb(DoubleTensor value, Set<? extends Vertex> withRespectTo) {
```
gordoncaleb: How did we miss this from before?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/HalfCauchyVertex.java
GeorgeNash: Yeah - we have masks in place for the `logProb` but not `dLogProb`, we just missed it

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/probabilistic/HalfGaussianVertex.java
GeorgeNash: No idea!

PR #395, opened 2018-12-14
made keanu-examples tests shorter

migwellian:
Reduced the number of samples we take in the `keanu-examples` tests

### DIFF
--- a/keanu-examples/bayesianMethodsForHackers/src/main/java/com/examples/TextMessaging.java
+++ b/keanu-examples/bayesianMethodsForHackers/src/main/java/com/examples/TextMessaging.java
@@ -50,7 +50,7 @@ public class TextMessaging {
         BayesianNetwork net = new BayesianNetwork(textsForDay.getConnectedGraph());
         net.probeForNonZeroProbability(1000);
 
-        final int numSamples = 800;
+        final int numSamples = 1000;
         NetworkSamples posteriorSamples = MetropolisHastings.withDefaultConfig()
             .generatePosteriorSamples(net, net.getLatentVertices())
             .dropCount(numSamples / 2)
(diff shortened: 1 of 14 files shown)


keanu-examples/bayesianMethodsForHackers/src/main/java/com/examples/TextMessaging.java
```diff
@@ -50,10 +50,10 @@ public static TextMessagingResults run() {
         BayesianNetwork net = new BayesianNetwork(textsForDay.getConnectedGraph());
         net.probeForNonZeroProbability(1000);
 
-        final int numSamples = 50000;
+        final int numSamples = 800;
```
christophernorth: By the looks of things - this is failing on Travis: Perhaps we need to make sure this test is deterministic? Or increase the number of samples slightly?

wesleyw72:
Interesting that Travis is failing here, I have rerun on my machine ~10 times with no problem

keanu-examples/bayesianMethodsForHackers/src/main/java/com/examples/TextMessaging.java
migwellian: It already has a `KeanuRandom.setDefaultRandomSeed(1)`

migwellian: `KeanuRandom` uses `org.apache.commons.math3.random.MersenneTwister` under the hood. Would you expect these to generate different numbers on different platforms (given the same seed)?

migwellian: I confirmed that our random seeds work the same on all platforms with a new test `KeanuRandomTest#randomNumberGenerationIsPlatformIndependent`.

I added `VertexId.ID_GENERATOR.set(1);` to every test and now it passes.

PR #400, opened 2018-12-17
Fix bug that causes Lambda Section discovery to be slow

gordoncaleb:
The LambdaSection depth first search was revisiting nodes when it shouldn't have. This was found on a very large graph where it was especially bad. I've added some LambdaSection testing that was missing from before and a new test to ensure that a given node is only visited once when traversing.


keanu-project/src/main/java/io/improbable/keanu/network/LambdaSection.java
```diff
@@ -90,18 +90,18 @@ public static LambdaSection getDownstreamLambdaSection(Vertex<?> aVertex, boolea
...
-        Set<Vertex> visited = new HashSet<>();
-        Deque<Vertex> stack = new ArrayDeque<>(nextVertices.apply(vertex));
+        Collection<Vertex> initialNext = nextVertices.apply(vertex);
+        Set<Vertex> queued = new HashSet<>(initialNext);
+        Deque<Vertex> stack = new ArrayDeque<>(initialNext);
         Set<Vertex> result = new HashSet<>();
         result.add(vertex);
 
         while (!stack.isEmpty()) {
             Vertex<?> visiting = stack.pop();
```
christophernorth: Given we're using the Raw Type everywhere else here should we drop the type specifier here?

keanu-project/src/test/java/io/improbable/keanu/network/LambdaSectionTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @Test
+    public void doesNotVisitVerticesMoreThanOnce() {
+        GaussianVertex A = new GaussianVertex(0, 1);
+        SinVertex sinA = A.sin();
+        AdditionVertex APlusSinA = A.plus(sinA);
+        GaussianVertex C = new GaussianVertex(APlusSinA, 1);
+
+        AtomicInteger callsToNext = new AtomicInteger(0);
+        AtomicInteger callsToPredicate = new AtomicInteger(0);
```
christophernorth: Not a huge fan of using AtomicIntegers for this sort of thing (given we don't need the Atomic behaviour.  Perhaps something like MutableInt (https://commons.apache.org/proper/commons-lang/javadocs/api-2.4/org/apache/commons/lang/mutable/MutableInt.html) unless it's a pain?

keanu-project/src/main/java/io/improbable/keanu/network/LambdaSection.java
```diff
@@ -90,18 +90,18 @@ public static LambdaSection getDownstreamLambdaSection(Vertex<?> aVertex, boolea
      * @param shouldAdd    true when a give vertex should be included in the result false otherwise
      * @return A Set of vertices that are in the direction implied by nextVertices and filtered by shouldAdd
      */
-    private static Set<Vertex> getVerticesDepthFirst(Vertex vertex,
-                                                     Function<Vertex, Collection<Vertex>> nextVertices,
-                                                     Predicate<Vertex> shouldAdd) {
+    public static Set<Vertex> getVerticesDepthFirst(Vertex vertex,
```
GeorgeNash: why the change to `public`? Can be package private if we're just making it public to access it in tests

keanu-project/src/test/java/io/improbable/keanu/network/LambdaSectionTest.java
```diff
@@ -0,0 +1,103 @@
...
+import static org.hamcrest.Matchers.containsInAnyOrder;
+
+public class LambdaSectionTest {
+
+    @Test
+    public void doesGetDownstreamProbabilisticVertices() {
+
+        GaussianVertex A = new GaussianVertex(0, 1);
+        GaussianVertex B = new GaussianVertex(0, 1);
+        GaussianVertex C = new GaussianVertex(A.plus(B), 1);
```
GeorgeNash: Make these member variables as we use them in all the tests

keanu-project/src/test/java/io/improbable/keanu/network/LambdaSectionTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+        LambdaSection lambdaSection = LambdaSection.getDownstreamLambdaSection(A, false);
+        Set<Vertex> verticesDepthFirst = lambdaSection.getLatentAndObservedVertices();
+
+        assertEquals(2, verticesDepthFirst.size());
+        assertThat(verticesDepthFirst, containsInAnyOrder(A, C));
```
GeorgeNash: If you can be bothered it may be worth checking the `order` actually matches depth first

GeorgeNash: It also feels a bit odd that the `getDownstream` includes the vertex we called it on? Why is `A` in this?

keanu-project/src/main/java/io/improbable/keanu/network/LambdaSection.java
gordoncaleb: 👍

keanu-project/src/test/java/io/improbable/keanu/network/LambdaSectionTest.java
gordoncaleb: 👍

keanu-project/src/main/java/io/improbable/keanu/network/LambdaSection.java
gordoncaleb: I'll need it to be public in a auto diff refactor I'm working on. Plus it's a pure static function so there's no harm in exposing it. Plus the package private for testing is a bit cheeky.

keanu-project/src/test/java/io/improbable/keanu/network/LambdaSectionTest.java
christophernorth: Given this is a hash set - I'm not sure we want to be assuming a particular order?

gordoncaleb: Good point about it being named getVerticesDepthFirst but there's no test to enforce this. I think in general whether it's DFS or BFS the user of this doesn't care. I've changed the name to just getVertices in order to not imply a specific traversal implementation.

keanu-project/src/test/java/io/improbable/keanu/network/LambdaSectionTest.java
```diff
@@ -0,0 +1,103 @@
...
+        GaussianVertex A = new GaussianVertex(0, 1);
+        SinVertex sinA = A.sin();
+        AdditionVertex APlusSinA = A.plus(sinA);
+        GaussianVertex C = new GaussianVertex(APlusSinA, 1);
+
+        AtomicInteger callsToNext = new AtomicInteger(0);
+        AtomicInteger callsToPredicate = new AtomicInteger(0);
+
+        Set<Vertex> verticesDepthFirst = LambdaSection.getVerticesDepthFirst(A, (v) -> {
+            callsToNext.getAndIncrement();
```
christophernorth: do you need the getAndIncrement version here?  Perhaps a plain increment() is better?

keanu-project/src/main/java/io/improbable/keanu/network/LambdaSection.java
GeorgeNash: ok

PR #401, opened 2018-12-17
Feature/jmh benchmarking

migwellian:
Our first JMH benchmarks. These ones are for comparing different ways to do "scalar.op(tensor)" for the basic operations `+ - * /`.


benchmarks/src/main/java/io/improbable/keanu/benchmarks/Nd4jDoubleTensorScalarOperations.java
```diff
@@ -0,0 +1,83 @@
...
+        DIVIDE {
+            public DoubleTensor apply(DoubleTensor lhs, DoubleTensor rhs) {
+                return lhs.div(rhs);
+            }
+        };
+
+        public abstract DoubleTensor apply(DoubleTensor lhs, DoubleTensor rhs);
+    }
+
+    @Param({"PLUS", "MINUS", "TIMES", "DIVIDE"})
```
gordoncaleb: Nice JMH skills.

benchmarks/src/test/java/io/improbable/keanu/benchmarks/Nd4jDoubleTensorScalarOperationsTest.java
```diff
@@ -0,0 +1,18 @@
...
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+
+public class Nd4jDoubleTensorScalarOperationsTest {
+
+    @Test
+    public void theScalarsHaveTheExpectedRank() {
```
gordoncaleb: Unit tests for the performance tests?

benchmarks/src/main/java/io/improbable/keanu/benchmarks/Nd4jDoubleTensorScalarOperations.java
```diff
@@ -0,0 +1,83 @@
...
+    @Benchmark
+    public long baseline() {
+        DoubleTensor product = operation.apply(scalars[0], tensor);
+        return product.getLength();
+    }
+
+    @Benchmark
+    public long rank1() {
+        DoubleTensor product = operation.apply(scalars[1], tensor);
+        return product.getLength();
```
gordoncaleb: Why return the length? I don't think getLength does much but if it did could that bias the benchmark?

benchmarks/src/main/java/io/improbable/keanu/benchmarks/Nd4jDoubleTensorScalarOperations.java
```diff
@@ -0,0 +1,83 @@
...
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+
+@State(Scope.Benchmark)
+public class Nd4jDoubleTensorScalarOperations {
+
+    public DoubleTensor tensor;
+
+    public static DoubleTensor[] scalars = new DoubleTensor[] {
+        Nd4jDoubleTensor.scalar(1.),
```
gordoncaleb: Maybe use 2 just to make sure nothing magical is happening with * 1 and /1?

benchmarks/src/main/java/io/improbable/keanu/benchmarks/Nd4jDoubleTensorScalarOperations.java
```diff
@@ -0,0 +1,83 @@
...
+    public double tensorLength;
+
+    @Setup
+    public void createTensor() {
+        tensor = DoubleTensor.arange(0., tensorLength);
+    }
+
+    @Benchmark
+    public long baseline() {
+        DoubleTensor product = operation.apply(scalars[0], tensor);
```
gordoncaleb: These tests will always have one of the operands be an Nd4j tensor, which I think means ALL of the ops will actually be done in Nd4j. I think we want to run the ops on two ScalarDoubleTensor(s) and two scalar Nd4jDoubleTensor(s) and then compare their performance. Credit: @christophernorth

benchmarks/src/test/java/io/improbable/keanu/benchmarks/Nd4jDoubleTensorScalarOperationsTest.java
migwellian: The purpose is to make sure that the baseline does what we expect (use a rank 0). If this test fails (which could happen with an ND4J upgrade), it's an indication that the benchmark needs to be rewritten.

benchmarks/src/main/java/io/improbable/keanu/benchmarks/Nd4jDoubleTensorScalarOperations.java
migwellian: Returning a value is a good way to make sure the compiler doesn't optimize away your code: it might otherwise detect that `product` does nothing and can be removed.

benchmarks/src/main/java/io/improbable/keanu/benchmarks/Nd4jDoubleTensorScalarOperations.java
migwellian: done

benchmarks/build.gradle
```diff
@@ -0,0 +1,28 @@
...
+    archiveName = 'benchmarks.jar'
+    classifier = 'benchmarks'
+    manifest.attributes('Main-Class': 'org.openjdk.jmh.Main')
+}
+
+task runAllBenchmarks(type: Exec) {
+    dependsOn(shadowJar)
+    commandLine 'java',
+        '-jar', 'build/libs/benchmarks.jar',
+        '-w', '1s', '-r', '1s', '-wi', '3', '-i', '5', '-f', '3'
```
gordoncaleb: What do these options do?

gordoncaleb: It would be nice to save the results as json or csv for easy diffing with other benchmark results. We can decide that when we come up with a nightly testing plan.

PR #402, opened 2018-12-17
List of distributions in Sphinx

wesleyw72:
Added a list to Sphinx for the available distributions available in Python. Ideally this will be automatically generated in the future. Have added a ticket to backlog for this.

Main files to be reviewed `keanu-python/docs/`. The rest are generated.


keanu-python/docs/distributions.rst
```diff
@@ -0,0 +1,28 @@
...
+* :meth:`keanu.vertex.MultivariateGaussian`
+* :meth:`keanu.vertex.Pareto`
+* :meth:`keanu.vertex.SmoothUniform`
+* :meth:`keanu.vertex.StudentT`
+* :meth:`keanu.vertex.Triangular`
+* :meth:`keanu.vertex.Uniform`
+* :meth:`keanu.vertex.Binomial`
+* :meth:`keanu.vertex.Multinomial`
+* :meth:`keanu.vertex.Poisson`
+* :meth:`keanu.vertex.UniformInt`
```
christophernorth: Perhaps we need to split this out in to continuous (Double) vs discrete (Integer)?

wesleyw72: Will leave it as it is for this release. Will add a note for when we automatically generate to split them up.

PR #403, opened 2018-12-17
Feature/python save load

christophernorth:
Simple wrapper around Save Load code (for Protobuf/Json + Dot) in the Python API.

Includes a very simple test that just dumps the output and loads it back again and ensures we don't get exceptions (it relies on the Java tests to actually test functionality though).


keanu-python/keanu/context.py
```diff
@@ -6,6 +6,7 @@
 from py4j.java_collections import JavaList, JavaArray, JavaSet, JavaMap
 from typing import Dict, Any, Iterable, List, Collection, Set
 from _io import TextIOWrapper
+from .base import JavaObjectWrapper
```
migwellian: I'm wary of adding this as it's likely to cause us grief with circular dependencies

keanu-python/keanu/context.py
```diff
@@ -66,15 +67,19 @@ def to_java_map(self, python_map: Dict[Any, Any]) -> JavaMap:
...
+            m.put(new_k, new_v)
 
         return m
 
     def to_java_object_list(self, l: Iterable[Any]) -> JavaList:
         lst = self._gateway.jvm.java.util.ArrayList()
 
         for o in l:
-            lst.add(o.unwrap())
+            o = o.unwrap() if isinstance(o, JavaObjectWrapper) else o
```
migwellian: I don't think you need this change - it's not used in your PR

keanu-python/keanu/context.py
```diff
@@ -66,15 +67,19 @@ def to_java_map(self, python_map: Dict[Any, Any]) -> JavaMap:
         m = self._gateway.jvm.java.util.HashMap()
 
         for (k, v) in python_map.items():
-            m.put(k.unwrap(), v.unwrap())
+            new_k = k.unwrap() if isinstance(k, JavaObjectWrapper) else k
+            new_v = v.unwrap() if isinstance(v, JavaObjectWrapper) else v
```
migwellian: Other ways we could do this - not as neat but avoids the dependency on `base`:
* before you call this method, wrap your `str` objects in a class that has an `unwrap` method
* check for the opposite case e.g. `isinstance(v, runtime_primitive_types) or isinstance(v, str)`
* try/except

keanu-python/keanu/network_io.py
```diff
@@ -0,0 +1,74 @@
...
+        saver.unwrap().save(output_stream, save_values, k.to_java_map(metadata))
+
+
+class ProtobufSaver(JavaObjectWrapper):
+
+    def __init__(self, net: BayesNet):
+        super(ProtobufSaver, self).__init__(k.jvm_view().ProtobufSaver(net.unwrap()))
+
+    def save(self, filename: str, save_values: bool = False, metadata: Optional[Dict[str, str]] = None):
+        _save_network(self, filename, save_values, metadata)
```
migwellian: can you do this with an abstract base class instead of the static `_save_network` method? It would avoid some code duplication.

migwellian: Ditto for `_load_network`

keanu-python/tests/test_net.py
```diff
@@ -62,3 +63,32 @@ def test_probe_for_non_zero_probability_from_bayes_net() -> None:
...
+    gamma = Gamma(1.0, 1.0)
+    gamma.set_value(2.5)
+    net = BayesNet(gamma.get_connected_graph())
+    metadata = {"Team": "GraphOS"}
+    protobuf_saver = ProtobufSaver(net)
+    protobuf_saver.save(PROTO_FILE, True, metadata)
+    json_saver = JsonSaver(net)
+    json_saver.save(JSON_FILE, True, metadata)
+    dot_saver = DotSaver(net)
+    dot_saver.save(DOT_FILE, True, metadata)
```
migwellian: Can you assert something about the `DOT_FILE`? Ideally, diff it against another file that you expect to be identical.

keanu-python/tests/test_net.py
```diff
@@ -62,3 +63,32 @@ def test_probe_for_non_zero_probability_from_bayes_net() -> None:
 
     assert gamma.has_value()
     assert poisson.has_value()
+
+
+def test_can_save_and_load() -> None:
+    PROTO_FILE = "temp.proto"
```
migwellian: PyTest has a cheeky feature called `tmpdir` that you can use out the box. See `test_lambdas.py` for an example.

keanu-python/keanu/network_io.py
GeorgeNash: +1

keanu-python/tests/test_net.py
```diff
@@ -62,3 +63,32 @@ def test_probe_for_non_zero_probability_from_bayes_net() -> None:
...
+    json_saver = JsonSaver(net)
+    json_saver.save(JSON_FILE, True, metadata)
+    dot_saver = DotSaver(net)
+    dot_saver.save(DOT_FILE, True, metadata)
+
+    protobuf_loader = ProtobufLoader()
+    json_loader = JsonLoader()
+    new_net_from_proto = protobuf_loader.load(PROTO_FILE)
+    new_net_from_json = json_loader.load(JSON_FILE)
+    latents = list(new_net_from_proto.get_latent_vertices())
```
GeorgeNash: can you call these `proto_latents` and `json_latents`? It looks like you're asserting the same thing twice otherwise

keanu-python/tests/test_net.py
```diff
@@ -62,3 +63,32 @@ def test_probe_for_non_zero_probability_from_bayes_net() -> None:
...
+    assert len(latents) == 1
+    new_gamma = latents[0]
+    assert new_gamma.get_value() == 2.5
+    latents = list(new_net_from_json.get_latent_vertices())
+    assert len(latents) == 1
+    new_gamma = latents[0]
+    assert new_gamma.get_value() == 2.5
```
GeorgeNash: do we assert that the meta data is saved correctly anywhere?

keanu-python/keanu/context.py
christophernorth: I actually do - the Bayesian network construction fails otherwise since the vertices returned by the getAllVertices() function are JavaObject and not JavaObjectWrapper

keanu-python/keanu/context.py
christophernorth: Perhaps the best way to solve this is to just collapse these two files down to like a core.py or something to stop pretending they're not interrelated

keanu-python/keanu/context.py
christophernorth: I'll have a think, but as you say - definitely not as clean and/or a lot more brittle.

keanu-python/keanu/network_io.py
christophernorth: Cool - will resolve.

keanu-python/tests/test_net.py
christophernorth: I've avoided doing that given there's already substantial testing on the Java side and don't want to add unnecessary testing.  Would checking that there are the correct number of lines in the file be sufficient?

keanu-python/tests/test_net.py
christophernorth: Cool - will do

keanu-python/tests/test_net.py
christophernorth: It's tested pretty extensively in Java but you're right, we don't check that it gets to Java properly in this case.  Would checking the metadata for the dot case be sufficient (given that all of this takes the same path, that seems enough to me)

keanu-python/tests/test_net.py
migwellian: That's a fair point.

You could do a bit more than counting lines, e.g. they always start with "`digraph BayesianNetwork {`"

keanu-python/tests/test_net.py
GeorgeNash: sounds good man

PR #409, opened 2018-12-18
nexus publish plugin

GeorgeNash:
- Upgrade to gradle `4.10.2`
- Upgrade to more recent `maven-publish` plugin
- Use `de.marcphilipp.nexus-publish` to avoid the issue with Travis changing IP during a nexus publish
- Use `nexusPublish` to publish to OSS when a master commit is triggered


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-bin.zip
```
gordoncaleb: Nice!

PR #411, opened 2018-12-18
Snippet writer removes leading whitespace

wesleyw72:
`snippet_writer.py` is a script we use for extracting snippets out of Java examples for the docs. Previously it did a straight copy from the source file and therefore included any leading whitespace the block of code had in the snippet. This required either poorly formatting the source files or modifying the generated snippet so that it formatted correctly on the shiny docs.

# In this PR
 - `snippet_writer.py` modified to remove leading whitespace from the block
 - Existing Java source code reformatted using IntelliJ formatter
 - Generated snippets updated with this formatting.

Note: commit 46550fc is where the generated snippets are updated with IntelliJ formatting so it may be easier to initially review without the diff of this commit.

### DIFF
--- a/docs/bin/snippet_writer.py
+++ b/docs/bin/snippet_writer.py
@@ -115,6 +115,12 @@ def check_output_dir_exists(output_dir):
     if (not os.path.exists(output_dir)):
         os.mkdir(output_dir)
 
+def strip_block_whitespace(string_list):
+    """Treats a list of strings as a code block and strips
+        whitespace so that the min whitespace line sits at char 0 of line."""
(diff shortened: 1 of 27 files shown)


docs/bin/snippet_writer.py
```diff
@@ -115,6 +115,12 @@ def check_output_dir_exists(output_dir):
     if (not os.path.exists(output_dir)):
         os.mkdir(output_dir)
 
+def strip_block_whitespace(string_list):
+    """Treats a list of strings as a code block and strips
+        whitespace so that the min whitespace line sits at char 0 of line."""
+    min_ws = min([(len(x) - len(x.lstrip())) for x in string_list if x != '\n'])
```
GeorgeNash: nice

PR #413, opened 2018-12-19
Refactor PartialDerivatives to contain single partial

gordoncaleb:
This PR contains quite a big refactor that does not contain many functional changes but significantly decreases the complexity of the forward and reverse mode auto diff logic. These refactors include:

- PartialDerivatives has been renamed PartialDerivative and no longer contains multiple partial derivatives.

- Changed the way forward mode auto diff is calculated in order to avoid needing to store more than a single partial derivative with respect to anything at a time.

- Operations that support operands that broadcast (e.g. +-*/) are now responsible for correcting shape changes due to that broadcast. This removes the need for lots of complex code to handle many edge cases in the partial multiply and in the reverse mode autodiff algorithm itself.

- The Differentiator reverse mode method now returns a PartialsOf object that contains a collection of partials with respect to many inputs but always of a single output. For example, `Differentiator.reverseModeAutoDiff(A, B, C)` would get the derivative of A with respect to B AND C. The object returned from this contains a `withRespectTo(...)` that would return the derivative of A with respect to B OR C.

- The Differentiator forward mode method now returns a PartialsWithRespectTo object that like the PartialsOf class contains a collection of partials but are all with respect to the same input but of many outputs. For example, `Differentiator.forwardModeAutoDiff(A, B, C)` would get the derivative of B and C with respect to A. The object returned from this contains a `of(...)` that would return the derivative of a single B OR C with respect to A.

- `Differentiable::getDerivativeWrtLatents` used to be backed by forward mode auto diff and used quite heavily in our tests. It has been removed and replaced with a direct call to the forward mode algorithm (`Differentiator::forwardModeAutoDiff`).

- Some operations have had their forward/reverse mode AD code refactored to reflect new guarantees provided by the fact that there is only a single partial derivative to deal with at a time.


keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/Differentiator.java
```diff
@@ -44,46 +43,45 @@ public static PartialDerivatives reverseModeAutoDiff(Vertex<?> ofVertex, Partial
...
-                collectPartials(partialDerivatives, dwrtOf, visiting.getShape());
-            }
-
             if (!visiting.isProbabilistic()) {
-                for (Vertex parent : visiting.getParents()) {
-                    if (!alreadyQueued.contains(parent) && parent instanceof Differentiable) {
-                        priorityQueue.offer(parent);
-                        alreadyQueued.add(parent);
+
+                if (visiting instanceof Differentiable) {
```
GeorgeNash: Would be worth adding a helper `.isDifferentiable` on to Vertex just like we have for `.isProbabilistic()`

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/AutoDiffBroadcast.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * This class is meant to help with auto diff in operations that support implicit broadcasting. E.g. In
+ * addition/subtraction/multiplication/division scalar operands can be operated with non-scalar operands.
+ */
+public class AutoDiffBroadcast {
+
+    public static PartialDerivative correctForScalarPartialForward(PartialDerivative partial, long[] partialOfShape, long[] targetOfShape) {
+
+        if (shouldCorrectPartialForScalarForward(partial, partialOfShape, targetOfShape)) {
```
GeorgeNash: This feels a bit backwards. We should really be checking `shouldCorrectPartialForScalarForward` before calling `correctForScalarPartialForward`

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/AutoDiffBroadcast.java
```diff
@@ -0,0 +1,54 @@
...
+
+            long[] wrtShape = partial.getWrtShape(partialOfShape);
+            DoubleTensor correctedPartial = DoubleTensor
+                .zeros(TensorShape.concat(targetOfShape, wrtShape))
+                .plus(partial.getPartial());
```
GeorgeNash: not a big fan of `getPartial()` - `getDerivative()`?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/AutoDiffBroadcast.java
```diff
@@ -0,0 +1,54 @@
...
+                targetWrtShape
+            );
+
+            return new PartialDerivative(partialSummed.reshape(resultShape));
+        } else {
+            return partial;
+        }
+    }
+
+    public static boolean shouldCorrectForPartialScalarReverse(PartialDerivative partial, long[] partialWrtShape, long[] targetWrtShape) {
```
GeorgeNash: isn't this the same as `shouldCorrectPartialForScalarForward`? Can we just call it `shouldCorrectPartialForScalar`

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/LogProbGradients.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    public LogProbGradients add(Map<VertexId, DoubleTensor> addition) {
+
+        Map<VertexId, DoubleTensor> initial = partials;
+
+        for (Map.Entry<VertexId, DoubleTensor> entry : addition.entrySet()) {
+            VertexId id = entry.getKey();
+            if (initial.containsKey(id)) {
+
```
GeorgeNash: nit: this looks a bit funky just delete the white space

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/LogProbGradients.java
```diff
@@ -0,0 +1,51 @@
+package io.improbable.keanu.vertices.dbl.nonprobabilistic.diff;
+
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+import io.improbable.keanu.vertices.VertexId;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class LogProbGradients {
```
GeorgeNash: Can we add javadoc to this? Is `LogProbGradients` now managing the summed partials which was previously being done by each `PartialDerivatives`?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/PartialDerivative.java
```diff
@@ -0,0 +1,177 @@
...
+        DoubleTensor divisorFromLeft = increaseRankByAppendingOnesToShape(divisor, partial.getRank());
+        DoubleTensor result = partial.div(divisorFromLeft);
+
+        return new PartialDerivative(result);
+    }
+
+    public static PartialDerivative matrixMultiplyAlongOfDimensions(PartialDerivative partial, DoubleTensor multiplier, boolean partialIsLeft) {
+
+        if (partial.isEmpty()) {
+            return partial;
```
GeorgeNash: why does this return `partial` and not `this`?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/PartialsOf.java
```diff
@@ -0,0 +1,33 @@
...
+    public PartialDerivative withRespectTo(Vertex vertex) {
+        return withRespectTo(vertex.getId());
+    }
+
+    public PartialDerivative withRespectTo(VertexId id) {
+        return partials.get(id);
+    }
+
+    public Map<VertexId, DoubleTensor> asMap() {
+        return partials.entrySet().stream()
```
GeorgeNash: will this be on the hot path when computing gradients in the optimizer / samplers?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/AutoDiffBroadcast.java
gordoncaleb: How about just get() since it's already a PartialDerivative?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/PartialDerivative.java
gordoncaleb: Because it's static and there is no `this`. It's the same concept though.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/AutoDiffBroadcast.java
GeorgeNash: aye

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/AutoDiffBroadcast.java
gordoncaleb: Good eye. It is the same. The different names clarify what's going on but I've changed it to be one and added some javadoc to explain.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/LogProbGradients.java
gordoncaleb: I've refactored this a bit. Kept some whitespace but also improved it a bit.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/LogProbGradients.java
gordoncaleb: 👍

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/AutoDiffBroadcast.java
gordoncaleb: I did that originally but since the check was done in so many places I just made it a one liner.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/Differentiator.java
gordoncaleb: Once upon a time instanceof was considered slow. These days it's not a problem and I don't think worth refactoring due to it not being as simple as Probabilistic/NonProbabilistic.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/PartialsOf.java
gordoncaleb: I made a mental note to change this but totally forgot. 👍

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/Differentiator.java
```diff
@@ -21,7 +20,7 @@
 
 public class Differentiator {
```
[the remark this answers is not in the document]
gordoncaleb: I've always been on the fence about @UtilityClass. It seems like the downside of the magic it introduces comes very close to outweighing the boilerplate code it removes. We use it in a few other places to I'll continue it here.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/Differentiator.java
```diff
@@ -93,73 +91,52 @@ private static void collectPartials(Map<Vertex, PartialDerivatives> partialDeriv
         }
     }
 
-    public static PartialDerivatives reverseModeAutoDiff(Vertex ofVertex, Set<DoubleVertex> wrt) {
-        return reverseModeAutoDiff(ofVertex, PartialDerivatives.withRespectToSelf(ofVertex.getId(), ofVertex.getShape()), ofVertex.getShape(), wrt);
+    public static PartialsOf reverseModeAutoDiff(Vertex ofVertex, Set<DoubleVertex> wrt) {
```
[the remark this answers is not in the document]
gordoncaleb: 👍

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/Differentiator.java
```diff
@@ -93,73 +91,52 @@ private static void collectPartials(Map<Vertex, PartialDerivatives> partialDeriv
...
+        if (ofVertex.isObserved()) {
+            return new PartialsOf(ofVertex, Collections.emptyMap());
+        } else {
+            return reverseModeAutoDiff(ofVertex, Differentiable.withRespectToSelf(ofVertex.getShape()), wrt);
+        }
     }
 
-    public static PartialDerivatives reverseModeAutoDiff(Vertex ofVertex, DoubleVertex... wrt) {
+    public static PartialsOf reverseModeAutoDiff(Vertex ofVertex, DoubleVertex... wrt) {
```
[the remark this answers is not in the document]
gordoncaleb: 👍

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/PartialsOf.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public PartialDerivative withRespectTo(VertexId id) {
+        return partials.get(id);
+    }
+
+    public Map<VertexId, DoubleTensor> asMap() {
+        final Map<VertexId, DoubleTensor> tensorMap = new HashMap<>();
+
+        for (Map.Entry<VertexId, PartialDerivative> entry : partials.entrySet()) {
```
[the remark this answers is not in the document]
gordoncaleb: This is very much in the hot path so I went for a for loop.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/Differentiator.java
```diff
@@ -93,73 +91,52 @@ private static void collectPartials(Map<Vertex, PartialDerivatives> partialDeriv
...
-                if (ofWrt.containsKey(ofsEntry.getKey())) {
-                    ofWrt.get(ofsEntry.getKey()).putWithRespectTo(wrtOfEntry.getKey(), ofsEntry.getValue());
-                } else {
-                    ofWrt.put(ofsEntry.getKey(), new PartialDerivatives(wrtOfEntry.getKey(), ofsEntry.getValue()));
-                }
-            }
-        }
-
-        return ofWrt;
+    public static <V extends Vertex & Differentiable> PartialsWithRespectTo forwardModeAutoDiff(V wrt, V... of) {
```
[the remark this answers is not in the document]
gordoncaleb: 👍

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/AutoDiffBroadcast.java
```diff
@@ -0,0 +1,58 @@
...
+import io.improbable.keanu.tensor.TensorShape;
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+
+import java.util.Arrays;
+
+/**
+ * This class is meant to help with auto diff in operations that support implicit broadcasting. E.g. In
+ * addition/subtraction/multiplication/division scalar operands can be operated with non-scalar operands.
+ */
+public class AutoDiffBroadcast {
```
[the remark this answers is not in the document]
gordoncaleb: 👍

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/DifferentiatorTest.java
```diff
@@ -156,4 +157,34 @@ public void canReverseAutoDiffOfMultiplicationLogSinAndSumWithSingleConditionalO
...
+    @Test
+    public void doesNotPerformUnnecessaryForwardModeAutoDiffCalculations() {
+        AtomicInteger n = new AtomicInteger(0);
+        AtomicInteger m = new AtomicInteger(0);
+        UniformVertex start = new UniformVertex(0, 1);
+
+        int links = 20;
+        TestGraphGenerator.SumVertex end = TestGraphGenerator.addLinks(start, n, m, links);
+
+        Differentiator.forwardModeAutoDiff(start, end);
```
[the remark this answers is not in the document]
gordoncaleb: Factored these tests out to their own class with their own before step.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/PartialsWithRespectTo.java
```diff
@@ -0,0 +1,20 @@
...
+
+import java.util.Map;
+
+@AllArgsConstructor
+public class PartialsWithRespectTo {
+
+    @Getter
+    private final Vertex wrt;
+
+    private final Map<Vertex, PartialDerivative> partials;
```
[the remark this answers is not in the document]
gordoncaleb: No reason. I've changed them to both key on VertexId.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/PartialDerivative.java
```diff
@@ -0,0 +1,177 @@
...
+    public static final PartialDerivative EMPTY = new PartialDerivative();
+
+    private final DoubleTensor partial;
+
+    public PartialDerivative(DoubleTensor partial) {
+        this.partial = partial;
+    }
+
+    private PartialDerivative() {
+        this.partial = null;
```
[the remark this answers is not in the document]
gordoncaleb: Removed the need for it altogether.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/PartialDerivative.java
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    public boolean isPresent() {
+        return partial != null;
+    }
+
+    public boolean isEmpty() {
+        return !isPresent();
+    }
```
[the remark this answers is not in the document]
gordoncaleb: I was going for an Optional like api, which only has isPresent so I've removed isEmpty.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/LogProbGradients.java
```diff
@@ -0,0 +1,53 @@
...
+
+            VertexId id = entry.getKey();
+            DoubleTensor existingPartialDerivative = partials.get(id);
+
+            if (existingPartialDerivative != null) {
+                partials.put(id, existingPartialDerivative.plusInPlace(entry.getValue()));
+            } else {
+                partials.put(id, entry.getValue().duplicate());
+            }
+        }
```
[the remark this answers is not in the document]
gordoncaleb: I've avoided using computeIfAbsent due to the lambda being a closure and therefore creates a new object every iteration. I like your point about only needing to do a single put. I've changed this to not re-put the double tensor object.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/Differentiator.java
```diff
@@ -1,165 +1,143 @@
...
+    private static void ensureGraphValuesAndShapesAreSet(Vertex<?> vertex) {
+        vertex.getValue();
     }
 
-    private static Set<Vertex> parentsThatPartialIsNotCalculated(Map<Vertex, PartialDerivatives> partialsOf, Collection<? extends Vertex> parents) {
-        Set<Vertex> notCalculatedParents = new HashSet<>();
-        for (Vertex next : parents) {
-            if (!partialsOf.containsKey(next) && next instanceof Differentiable) {
-                notCalculatedParents.add(next);
+    private static void collectPartials(Map<Vertex, PartialDerivative> partialDerivatives,
```
christophernorth: So the broadcast/reduce to scalar stuff is now handled in the vertices themselves (with help from the AutoDiffBroadcast class) now?  If so, much cleaner than handling that here IMO - good change.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/AutoDiffBroadcast.java
```diff
@@ -0,0 +1,60 @@
...
+
+import io.improbable.keanu.tensor.TensorShape;
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+import lombok.experimental.UtilityClass;
+
+import java.util.Arrays;
+
+/**
+ * This class is meant to help with auto diff in operations that support implicit broadcasting. E.g. In
+ * addition/subtraction/multiplication/division scalar operands can be operated with non-scalar operands.
```
christophernorth: Obviously it's not just scalars that can be broadcast and lead to implicit partial changes - it's also compatible Tensors - presumably given the naming that's still a bug we have to fix at some point?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/LogProbGradients.java
```diff
@@ -0,0 +1,53 @@
...
+
+        for (Map.Entry<VertexId, DoubleTensor> entry : addition.entrySet()) {
+
+            VertexId id = entry.getKey();
+            DoubleTensor existingPartialDerivative = partials.get(id);
+
+            if (existingPartialDerivative == null) {
+                partials.put(id, entry.getValue().duplicate());
+            } else {
+                existingPartialDerivative.plusInPlace(entry.getValue());
```
christophernorth: Don't we have a potential bug hiding here now?  plusInPlace sometimes doesn't actually do things in place...

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/PartialDerivative.java
```diff
@@ -0,0 +1,170 @@
+package io.improbable.keanu.vertices.dbl.nonprobabilistic.diff;
+
+import io.improbable.keanu.tensor.TensorShape;
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+
+import java.util.Arrays;
+
+public class PartialDerivative {
```
christophernorth: I know we might add a bit more state to this in future to help with tracking dimensions etc - but for now it could be argued that having this as an explicit object is just costing us in terms of performance (due to the extra level of indirection around the Tensor object).  I think it's fine for now, but something to consider in future that given every method call has to go via this object and then to the tensor object we're hopping around memory quite a bit.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/PartialDerivative.java
christophernorth: FWIW - I like this explanation.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/PartialsOf.java
```diff
@@ -0,0 +1,38 @@
...
+            tensorMap.put(entry.getKey(), entry.getValue().get());
```
christophernorth: Given this is in the hot path of reverse mode auto-diff - I'm not sure I love having to do this translation from one map type to another - just seems to be adding busy work to avoid other people seeing PartialDerivatives rather than DoubleTensors ?

keanu-project/src/test/java/io/improbable/keanu/util/io/ProtobufTest.java
```diff
@@ -131,22 +132,29 @@ public void saveLoadGradientTest() throws IOException {
         ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
         ProtobufSaver saver = new ProtobufSaver(complexNet);
         saver.save(outputStream, true);
-        DoubleIfVertex outputVertex = (DoubleIfVertex)complexNet.getVertexByLabel(new VertexLabel(OUTPUT_NAME));
-        DoubleVertex inputVertex = (DoubleVertex)complexNet.getVertexByLabel(new VertexLabel(INPUT_NAME));
+        DoubleIfVertex outputVertex = (DoubleIfVertex) complexNet.getVertexByLabel(new VertexLabel(OUTPUT_NAME));
+        DoubleVertex inputVertex = (DoubleVertex) complexNet.getVertexByLabel(new VertexLabel(INPUT_NAME));
```
christophernorth: Should we just cast this straight to the proper vertex type to avoid having to do the cast in the Differentiator below?

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/DifferentiatorEfficiencyTest.java
```diff
@@ -0,0 +1,40 @@
+package io.improbable.keanu.vertices.dbl;
```
christophernorth: Do you think it would be useful to wrap this as a Benchmark so we can monitor changes to this perf?

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/DifferentiatorTest.java
```diff
@@ -42,10 +40,10 @@ public void canReverseAutoDiffOfMultiplicationWithSingleOutputWithRespectToMany(
         DoubleVertex B = new GaussianVertex(0, 1);
         DoubleVertex C = A.times(B);
 
-        PartialDerivatives dC = Differentiator.reverseModeAutoDiff(C, ImmutableSet.of(A, B));
+        PartialsOf dC = Differentiator.reverseModeAutoDiff(C, ImmutableSet.of(A, B));
 
-        DoubleTensor dCdA = dC.withRespectTo(A);
-        DoubleTensor dCdB = dC.withRespectTo(B);
+        DoubleTensor dCdA = dC.withRespectTo(A).get();
+        DoubleTensor dCdB = dC.withRespectTo(B).get();
```
christophernorth: Bit ugly we have to call this get() method all over the place?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/Differentiable.java
```diff
@@ -1,29 +1,30 @@
...
         }
     }
 
-    default Map<Vertex, PartialDerivatives> reverseModeAutoDifferentiation(PartialDerivatives derivativeOfOutputsWithRespectToSelf) {
-        return Collections.singletonMap(
-            ((Vertex)this),
-            PartialDerivatives.withRespectToSelf(((Vertex)this).getId(), ((Vertex)this).getShape())
+    static PartialDerivative withRespectToSelf(long[] shape) {
+        return new PartialDerivative(
+            DoubleTensor.eye((int) TensorShape.getLength(shape)).reshape(TensorShape.concat(shape, shape))
```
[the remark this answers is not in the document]
gordoncaleb: Very good eye. It's from before the days where length was a long.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/Differentiator.java
gordoncaleb: Yup. It's no longer the responsibility of the autodiff algo to check and handle these shape change issues. The operation logic is in a much better place to handle any shape change.

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/DifferentiatorTest.java
gordoncaleb: If you want it as a DoubleTensor then yes. We could call it `toDoubleTensor()` or we could just return the DoubleTensor from the `withRespectTo(...)` method. It doesn't look like the `withRespectTo(...)` method is ever used to get the PartialDerivative.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/PartialDerivative.java
gordoncaleb: It's not ideal but it provides some autodiff context that DoubleTensor wouldn't give us. If we want to track the shape wrt/of then we'll need a class to hold it and it's a nice way to represent a null partial derivative. The null partial could probably be replaced with some clever refactoring but the wrt/of shape is a autodiff domain concept that doesn't map to DoubleTensor. If we move towards graphifying the autodiff then this indirection will be compiled out anyways.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/AutoDiffBroadcast.java
gordoncaleb: Those non-scalar broadcast aren't supported yet and strictly prohibited through the shape check in the operations that could broadcast. When we add that support then this code will need to be tweaked to support the new cases.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/LogProbGradients.java
gordoncaleb: At the moment it's impossible for the existing partial to have a different shape to the next partial. All partials at this point are the correct deterministic shape [of, wrt]. That being said, the inPlace contract doesn't guarantee that the object will always be the same. I've gone with the double put to make sure this is never an issue.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/PartialsOf.java
gordoncaleb: I tried avoiding this earlier but gave up. I've refactored it to to the conversion at the point the double tensor is used. This avoids the double iteration and the new object creation.

keanu-project/src/test/java/io/improbable/keanu/util/io/ProtobufTest.java
gordoncaleb: Nice clean up.

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/DifferentiatorEfficiencyTest.java
gordoncaleb: This tests "efficiency", which just means it's not doing any more calls to autodiff than needed. I think the benchmarks should try quite a few more complex graphs that vary in length, depth, width, etc. The benchmarks *could* use this for inspiration but I don't think there's much benefit to coupling the benchmarks and this unit test.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/Differentiator.java
christophernorth: 👍

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/AutoDiffBroadcast.java
christophernorth: Cool stuff - If you see that as a bigger refactor, no bother.  I wonder if there's anything we can do to mark this as something that will need fixing if we add broadcasting properly so this isn't missed?

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/LogProbGradients.java
christophernorth: I think better to be safe than sorry definitely!

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/diff/PartialDerivative.java
christophernorth: indeed - as I say, not a prob for now.

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/DifferentiatorTest.java
gordoncaleb: Having `withRespectTo(...)` return a DoubleTensor removes all of these `get()` calls. Nice clean up.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/Differentiable.java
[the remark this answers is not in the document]
gordoncaleb: Agreed offline that the use cases where this would be a problem aren't enough to cause concern.

PR #416, opened 2018-12-20
Assert vertex

wesleyw72:
Adds an assert vertex that can be used to ensure a `BoolVertex` predicate is true.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/AssertVertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/AssertVertex.java
@@ -11,6 +11,22 @@ public class AssertVertex extends BoolVertex implements NonProbabilistic<Boolean
     private final BooleanTensor expected;
     private final String errorMessage;
 
+    public static AssertVertex allTrue(Vertex<? extends BooleanTensor> predicate) {
+        return new AssertVertex(predicate, BooleanTensor.trues(predicate.getShape()));
+    }
+
+    public static AssertVertex allTrue(Vertex<? extends BooleanTensor> predicate, String errorMessage) {
+        return new AssertVertex(predicate, BooleanTensor.trues(predicate.getShape()), errorMessage);
+    }
+
+    public static AssertVertex allFalse(Vertex<? extends BooleanTensor> predicate) {
+        return new AssertVertex(predicate, BooleanTensor.falses(predicate.getShape()));
+    }
+
+    public static AssertVertex allFalse(Vertex<? extends BooleanTensor> predicate, String errorMessage) {
+        return new AssertVertex(predicate, BooleanTensor.falses(predicate.getShape()), errorMessage);
+    }
+
     public AssertVertex(Vertex<? extends BooleanTensor> predicate, BooleanTensor expected,
                         String errorMessage) {
         super(TensorShapeValidation.checkAllShapesMatch(predicate.getShape(), expected.getShape()));
@@ -21,18 +37,9 @@ public class AssertVertex extends BoolVertex implements NonProbabilistic<Boolean
     }
 
     public AssertVertex(Vertex<? extends BooleanTensor> predicate, BooleanTensor expected) {
-        this(predicate,expected,"Failed assertion");
+        this(predicate, expected, "Failed assertion");
     }
 
-    private boolean predicateMatchesExpected() {
-        return predicate.getValue().xor(expected).allFalse();
-    }
-
-    private void assertion() {
-        if(!predicateMatchesExpected()) {
-            throw new AssertionError(errorMessage);
-        }
-    }
 
     @Override
     public BooleanTensor calculate() {
@@ -40,8 +47,20 @@ public class AssertVertex extends BoolVertex implements NonProbabilistic<Boolean
         return predicate.getValue();
     }
 
+    private void assertion() {
+        if (!predicateMatchesExpected()) {
+            throw new AssertionError(errorMessage);
+        }
+    }
+
+    private boolean predicateMatchesExpected() {
+        return predicate.getValue().xor(expected).allFalse();
+    }
+
+
     @Override
     public BooleanTensor sample(KeanuRandom random) {
         return predicate.sample();
     }
+
 }
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/AssertVertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/AssertVertex.java
@@ -7,34 +7,41 @@ import io.improbable.keanu.vertices.dbl.KeanuRandom;
 
 public class AssertVertex extends BoolVertex implements NonProbabilistic<BooleanTensor>, NonSaveableVertex {
 
-    private final Vertex<? extends BooleanTensor> actual;
+    private final Vertex<? extends BooleanTensor> predicate;
     private final BooleanTensor expected;
+    private final String errorMessage;
 
-    public AssertVertex(Vertex<? extends BooleanTensor> predicate, BooleanTensor expected) {
+    public AssertVertex(Vertex<? extends BooleanTensor> predicate, BooleanTensor expected,
+                        String errorMessage) {
         super(TensorShapeValidation.checkAllShapesMatch(predicate.getShape(), expected.getShape()));
-        this.actual = predicate;
+        this.predicate = predicate;
         this.expected = expected;
+        this.errorMessage = errorMessage;
         setParents(predicate);
     }
 
-    private boolean actualMatchesExpected() {
-        return actual.getValue().xor(expected).allFalse();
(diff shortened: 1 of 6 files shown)


keanu-project/src/test/java/io/improbable/keanu/vertices/AssertVertexTest.java
```diff
@@ -0,0 +1,88 @@
...
+    @Test
+    public void lazyEvalThrowsOnFalseConstBool() {
+        thrown.expect(AssertionError.class);
+        ConstantBoolVertex constBool = new ConstantBoolVertex(BooleanTensor.create(false));
+        constBool.assertTrue().lazyEval();
+    }
+
+    @Test
+    public void assertPassesOnRVWithTruePredicate() {
+        UniformVertex uniform = new UniformVertex(0,5);
```
GeorgeNash: nit: run a code formatter on this to get your spacing correct between commas `(0, 5)`

keanu-project/src/test/java/io/improbable/keanu/vertices/AssertVertexTest.java
```diff
@@ -0,0 +1,88 @@
...
+        thrown.expect(AssertionError.class);
+        UniformVertex uniform = new UniformVertex(0,5);
+        BoolVertex predicate = uniform.greaterThan(new ConstantDoubleVertex(new double[]{10}));
+        predicate.assertTrue().eval();
+    }
+
+    @Test
+    public void samplingWithAssertionWorks() {
+        thrown.expect(AssertionError.class);
+        GaussianVertex uniform = new GaussianVertex(5,1);
```
GeorgeNash: rename this to `gaussian` not `uniform`. Why do we need the `observingVertex`? I don't really understand what it adds to this test

keanu-project/src/test/java/io/improbable/keanu/vertices/AssertVertexTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Test
+    public void optimizerWithAssertionWorks() {
+        thrown.expect(AssertionError.class);
+        UniformVertex temperature = new UniformVertex(20,30);
+        GaussianVertex observedTemp = new GaussianVertex(temperature,1);
+        observedTemp.observe(29);
+        temperature.greaterThan(new ConstantDoubleVertex(34)).assertTrue();
+        KeanuOptimizer.of(temperature.getConnectedGraph()).maxAPosteriori();
+        System.out.println(temperature.getValue().scalar());
```
GeorgeNash: Delete this print

keanu-project/src/test/java/io/improbable/keanu/vertices/AssertVertexTest.java
```diff
@@ -0,0 +1,88 @@
...
+import io.improbable.keanu.vertices.bool.BoolVertex;
+import io.improbable.keanu.vertices.bool.nonprobabilistic.ConstantBoolVertex;
+import io.improbable.keanu.vertices.dbl.nonprobabilistic.ConstantDoubleVertex;
+import io.improbable.keanu.vertices.dbl.probabilistic.GaussianVertex;
+import io.improbable.keanu.vertices.dbl.probabilistic.UniformVertex;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+public class AssertVertexTest {
```
GeorgeNash: Nice tests! Would be good to see one where we check it handles value propagation correctly. Assert that a vertex is less than a value, expect it to pass, change the vertex's value to now be much larger, assert that it fails.

Also - we use the optimiser / sampler and RV's quite a bit in this test class, can you set the `DeterministicRule`?

wesleyw72: Good idea. Have added two which I think should cover this now. They check whether both a valid and invalid `setAndCascade` has the correct behaviour. Also added `DeterministicRule`.

keanu-project/src/test/java/io/improbable/keanu/vertices/AssertVertexTest.java
```diff
@@ -0,0 +1,104 @@
...
+        thrown.expect(AssertionError.class);
+        UniformVertex temperature = new UniformVertex(20, 30);
+        GaussianVertex observedTemp = new GaussianVertex(temperature, 1);
+        observedTemp.observe(29);
+        temperature.greaterThan(new ConstantDoubleVertex(34)).assertTrue();
+        KeanuOptimizer.of(temperature.getConnectedGraph()).maxAPosteriori();
+    }
+
+    @Test
+    public void assertGivesCorrectErrorMessage() {
```
migwellian: could you please add tests for the `expectMessage` when:
* there is no message in the `AssertVertex`
* the `AssertVertex` has no label

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/BoolVertex.java
```diff
@@ -85,6 +86,10 @@ public static final BoolVertex not(Vertex<BooleanTensor> vertex) {
         return new NotVertex(vertex);
     }
 
+    public AssertVertex assertTrue() { return new AssertVertex(this); }
```
gordoncaleb: Nice touch.

keanu-project/src/test/java/io/improbable/keanu/vertices/AssertVertexTest.java
```diff
@@ -0,0 +1,132 @@
...
+    public void assertPassesOnRVWithFalsePredicate() {
+        thrown.expect(AssertionError.class);
+        UniformVertex uniform = new UniformVertex(0, 5);
+        BoolVertex predicate = uniform.greaterThan(new ConstantDoubleVertex(new double[]{10}));
+        predicate.assertTrue().eval();
+    }
+
+    @Test
+    public void samplingWithAssertionWorks() {
+        thrown.expect(AssertionError.class);
```
christophernorth: I think it would be good to add some tests that ensure inference still works correctly even when we have a graph that contains asserts (ie most of these tests check that the assertion correctly fires - it would be good to have ones where the assertion *isn't* supposed to fire).

Even better if we have a test that checks we can still use a gradient optimiser even when we have asserts?

keanu-project/src/main/java/io/improbable/keanu/vertices/AssertVertex.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public BooleanTensor calculate() {
+        assertion();
+        return predicate.getValue();
+    }
+
+    private void assertion() {
+        if (!predicate.getValue().allTrue()) {
+            throw new AssertionError(buildAssertMessage());
```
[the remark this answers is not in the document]
wesleyw72: Good point. Replaced it with a `GraphAssertionException`.

keanu-project/src/test/java/io/improbable/keanu/vertices/AssertVertexTest.java
wesleyw72: Good points. Added some for ensuring an assertion that shouldn't fire doesn't fire for optimising/sampling. Also added the gradient optimiser check you suggested.

keanu-project/src/test/java/io/improbable/keanu/vertices/AssertVertexTest.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    @Test
+    public void assertGivesCorrectErrorWhenPlain() {
+        thrown.expect(AssertionError.class);
+        thrown.expectMessage("AssertVertex");
+        ConstantBoolVertex constBool = new ConstantBoolVertex(BooleanTensor.create(false));
+        AssertVertex assertVertex = constBool.assertTrue();
+        assertVertex.eval();
+    }
```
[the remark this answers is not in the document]
wesleyw72: Added assertions at multiple stages of a more complex graph in the `optimizerWithAssertThatShouldntFire` test. Also added a double assertion test.

keanu-project/src/test/java/io/improbable/keanu/vertices/utility/AssertVertexTest.java
```diff
@@ -0,0 +1,211 @@
...
+        double maxB = B.getValue().scalar();
+
+        assertEquals(22, maxA, 0.1);
+        assertEquals(22, maxB, 0.1);
+    }
+
+    @Test
+    public void doubleAssertion() {
+        DoubleVertex A = new ConstantDoubleVertex(20);
+        A.lessThan(new ConstantDoubleVertex(30)).assertTrue().and(ConstantBoolVertex.TRUE).assertTrue();
```
gordoncaleb: .eval()?

PR #422, opened 2019-01-04
Tweaks to Python build

alexsparrow:
A few tweaks to the Python build that I found helpful:

- Bump `pipenv` version. This seemed to help with some flakiness for me.
- Remove `pythonVersionInfo` task from `build`/`test`. This was adding noise to the build so propose we run this as needed for debugging.
- Change the way exclusion paths are setup. Haven't tested this on Windows yet but can't see why the existing special casing would be needed.
- Add `buildBinaryDistribution` task to package the library
- Switch default `matplotlib` backend to `Agg` to workaround issues with non-framework Python on OSX.


keanu-python/build.gradle
```diff
@@ -6,7 +6,7 @@ import static org.apache.tools.ant.taskdefs.condition.Os.isFamily
 task installPipenv {
     doLast {
         exec {
-            commandLine 'pip', 'install', 'pipenv==2018.10.13'
+            commandLine 'pip', 'install', 'pipenv==2018.11.26'
```
migwellian: We could drop the version number (i.e. always install latest). 

The version number was needed temporarily when we first set up our python build, because coincidentally at that time a new release of pip created a bug that caused pipenv to fail.

keanu-python/build.gradle
```diff
@@ -143,6 +143,6 @@ pythonVersionInfo.mustRunAfter(installPipenv)
 installDependencies.dependsOn(installPipenv)
 generateDocumentation.dependsOn(preparePythonEnvironment)
 test.dependsOn(preparePythonEnvironment)
-test.dependsOn(formatCheck)
 test.dependsOn(pytest)
+check.dependsOn(formatCheck)
```
migwellian: actually we want `formatCheck` to happen earlier, before `test`. When it fails we want to know early.

keanu-python/matplotlibrc
```diff
@@ -0,0 +1 @@
+backend: Agg
```
migwellian: This is a good idea. I had previously worked around the problem locally by putting this in `~/.matplotlib/matplotlibrc`

migwellian: I believe it still yields warnings:

> UserWarning: Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.

I don't know if it's possible to do better, but the warnings seem innocuous.

keanu-python/build.gradle
```diff
@@ -43,7 +43,6 @@ task installDependencies {
 
 task preparePythonEnvironment {
     dependsOn(installPipenv)
-    dependsOn(pythonVersionInfo)
```
migwellian: fine by me. We only really needed this diagnostic information when we were first setting the python build up to run on the CI servers.

keanu-python/build.gradle
```diff
@@ -101,40 +100,41 @@ task generateDocumentation {
...
 task formatApply {
     mustRunAfter preparePythonEnvironment // this is a dependency but only needs to be run once
     mustRunAfter ':spotlessPythonApply'   // spotless does some checks of the python code too
-    dependsOn populateExclusionPaths
 
     doLast {
         exec {
-            commandLine 'pipenv', 'run', 'yapf', '--in-place', '--recursive', '--parallel',
-            '--exclude', keanuVertexGeneratedPath, '--exclude', keanuVertexInitPath, '.'
+            commandLine(['pipenv', 'run', 'yapf', '--in-place', '--recursive', '--parallel', project.rootDir].addAll(formatExcludes))
```
migwellian: the `addAll` is neat

keanu-python/build.gradle
```diff
@@ -101,40 +100,41 @@ task generateDocumentation {
...
-    def keanuVertexPath = Paths.get("keanu", "vertex")
-    if(isFamily(FAMILY_WINDOWS)) {
-        keanuVertexPath = Paths.get(".").resolve(keanuVertexPath)
-    }
-    keanuVertexGeneratedPath = keanuVertexPath.resolve("generated.py")
-    keanuVertexInitPath = keanuVertexPath.resolve("__init__.py")
-}
+List<String> formatExcludes = [
+        '--exclude', project.file('keanu/vertex/generated.py'),
+        '--exclude', project.file('keanu/vertex/__init__.py'),
```
migwellian: I would expect this to fail on Windows because (1) the path separator needs to be `\` (2) For some reason it only works on Windows when the path is prefixed by `.\` prefix

keanu-python/matplotlibrc
alexsparrow: Yep I looked briefly at getting rid of the warnings but couldn't seem to. Will look again.

keanu-python/build.gradle
alexsparrow: Yeah I'd hoped `.file` would somehow magically deal with this for us but I'm guessing it's an issue with `yapf`? Should it break the CI builds?

keanu-python/build.gradle
alexsparrow: Yeah fair enough will revert.

keanu-python/build.gradle
alexsparrow: Agree that would probably be fine. Given the flakiness that I've seen between `pipenv` version, I think I'd be in favour of keeping the explicit version check for now. Maybe things have got better since but see e.g. https://chriswarrick.com/blog/2018/07/17/pipenv-promises-a-lot-delivers-very-little/#the-break-neck-pace-of-pipenv

keanu-python/build.gradle
alexsparrow: Seems like  the `appveyor` build should trigger issues? If not, perhaps we should ensure it does?

keanu-python/matplotlibrc
alexsparrow: Have added a filter for the warnings now.

keanu-python/build.gradle
alexsparrow: Done.

keanu-python/build.gradle
migwellian: fine by me

keanu-python/build.gradle
migwellian: I'm surprised it's passing on AppVeyor. @wesleyw72 had a lot of grief with this when he did the original implementation.

migwellian: Or maybe it doesn't fail, it just ignores the `--exclude`?

keanu-python/matplotlibrc
migwellian: nice

keanu-python/build.gradle
migwellian: Oh, I see what's happening now - you're actually not executing any command in `formatCheck` and `formatApply`!

The `addAll` returns `true`, so all you are actually just executing `commandLine(true)` which trivially succeeds.

So instead you want something like:

        def command = ['pipenv', 'run', 'yapf', '--diff', '--recursive', '--parallel', project.rootDir]
        command.addAll(formatExcludes)
        exec {
            commandLine(command)
        }

alexsparrow:
Haha @migwellian good spot. That is pretty bad behaviour :-) Will fix.

keanu-python/build.gradle
migwellian: On the plus side - `project.file` does format it correctly for Windows as you predicted (I checked on a laptop). But it actually fails for a different reason, which is that it is running (and failing) against `keanu\docs`. This is because you are pointing yapf at `project.rootDir` (i.e. `keanu`) when it should be `.` (i.e. `keanu\keanu-python`)

keanu-python/tests/test_sampling.py
```diff
@@ -64,6 +64,7 @@ def test_down_sample_interval(net: BayesNet) -> None:
     assert all(len(vertex_samples) == expected_num_samples for label, vertex_samples in samples.items())
 
 
+@pytest.mark.filterwarnings('ignore:Matplotlib is currently using agg')
```
GeorgeNash: nit: pull the 'ignore:Matplotlib is currently using agg' out into a global if you'd like

alexsparrow: Ah yeah good call.

keanu-python/build.gradle
alexsparrow: Yep just realised that. I've made a change now to use `.yapfignore` instead. We'll see how that goes.

migwellian: 👍

keanu-python/tests/test_sampling.py
alexsparrow: I've added this to `pytest.ini` now. Take a look when you get a chance @GeorgeNash .

alexsparrow:
@migwellian This should now have been fixed with your help on the Windows stuff.

keanu-python/build.gradle
wesleyw72: Nice! I wanted to do it with an ignore file when I first did it but they didn't have the functionality then. Good spot that they've added this!

keanu-python/.yapfignore
```diff
@@ -1,3 +1,8 @@
 keanu/vertex/generated.py
 keanu/vertex/__init__.py
 build/**
+
+# Need to duplicate patterns for Windows as yapf doesn't seem to do anything special
```
wesleyw72: It's annoying that we have to do this. I have raised an [issue](https://github.com/google/yapf/issues/661) on the `yapf` repo.

alexsparrow: Nice one totally agree it should normalise the paths or something.

keanu-python/build.gradle
```diff
@@ -101,40 +100,34 @@ task generateDocumentation {
...
+        }
+    }
+}
+
+task buildSourceDistribution {
+    mustRunAfter preparePythonEnvironment
+
+    doLast {
+        exec {
+            commandLine 'pipenv', 'run', 'python', project.file("setup.py"), 'sdist'
```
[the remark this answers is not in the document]
alexsparrow: Will do. I think my reasoning was to fix the `python` version against the one in the `Pipfile`. Not super compelling though so can take it out.

Will switch to `bdist_wheel` 👍

keanu-python/build.gradle
```diff
@@ -101,40 +100,35 @@ task generateDocumentation {
...
+    }
+}
+
+task buildWheelDistribution {
+    mustRunAfter preparePythonEnvironment
+    mustRunAfter preparePythonEnvironment
+
+    doLast {
+        exec {
+            commandLine 'pipenv', 'run', 'python', project.file("setup.py"), 'bdist_wheel'
```
migwellian:
```suggestion
            commandLine 'pipenv', 'run', 'python', project.file("setup.py"), 'sdist', 'bdist_wheel'
```

keanu-python/build.gradle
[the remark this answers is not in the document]
migwellian: Sorry, I meant that you should do both (`sdist` and `bdist_wheel`).

I have doneit as an "insert suggestion" in the github UI so you can just press the "accept" button to apply it.

PR #424, opened 2019-01-08
Improved differentiable check

wesleyw72:
An improved check for whether a graph is differentiable w.r.t latents. This is necessary to enable graphs with non differentiable vertices (e.g assert) that don't sit on a differentiable path to be used with gradient based sampling/optimisation.


keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableChecker.java
```diff
@@ -0,0 +1,90 @@
...
+        while (!queue.isEmpty()) {
+            Vertex visiting = queue.poll();
+
+            if (constantValueVerticesCache.contains(visiting)) {
+                continue;
+            }
+
+            if (visiting.isProbabilistic()) {
+                if (visiting.isObserved()) {
+                    continue;
```
icantsayrural: Same here, continue in the middle of a while block is confusing to read.

keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableChecker.java
```diff
@@ -0,0 +1,90 @@
...
+            }
+        }
+        return true;
+    }
+
+    private boolean isVertexConstant(Vertex vertex, Set<Vertex> constantValueVerticesCache) {
+        if(vertex.isProbabilistic() && !vertex.isObserved()) {
+            return false;
+        }
+        if(!isVertexParentsValueConstant(vertex, constantValueVerticesCache)) {
```
icantsayrural: If it's probabilistic and observed, do we need to check its parents?

keanu-project/src/test/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableCheckerTest.java
```diff
@@ -0,0 +1,166 @@
...
+        observedLatent.observe(4);
+        FloorVertex nonDiffableVertex = new FloorVertex(observedLatent);
+        GaussianVertex gaussian = new GaussianVertex(nonDiffableVertex, 1.);
+        BayesianNetwork bayesianNetwork = new BayesianNetwork(gaussian.getConnectedGraph());
+        boolean differentiable = DifferentiableChecker.isDifferentiable(bayesianNetwork.getLatentOrObservedVertices());
+        assertEquals(true, differentiable);
+    }
+
+    @Test
+    public void constantAsParentOfNonDiff() {
```
migwellian:
```suggestion
    public void constantAsParentOfNonDiffIsDiffable() {
```

keanu-project/src/test/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableCheckerTest.java
```diff
@@ -0,0 +1,166 @@
...
+        GaussianVertex gaussianA = new GaussianVertex(5., 1.);
+        gaussianA.lessThan(new ConstantDoubleVertex(90)).assertTrue();
+        GaussianVertex gaussianB = new GaussianVertex(gaussianA, 1.);
+        BayesianNetwork bayesianNetwork = new BayesianNetwork(gaussianB.getConnectedGraph());
+        boolean differentiable = DifferentiableChecker.isDifferentiable(bayesianNetwork.getLatentOrObservedVertices());
+        assertEquals(true, differentiable);
+    }
+
+    @Test
+    public void multipleNonDiffableWithConstantParents() {
```
migwellian:
```suggestion
    public void multipleNonDiffableWithConstantParentsIsDiffable() {
```

keanu-project/src/test/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableCheckerTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+    @Test
+    public void observedVertexWithNonDiffableConstantParentIsDiffable() {
+        DoubleVertex constDouble = new ConstantDoubleVertex(4);
+        FloorVertex nonDiffable = new FloorVertex(constDouble);
+        GaussianVertex observed = new GaussianVertex(nonDiffable, 1.);
+        observed.observe(4.);
+        BayesianNetwork bayesianNetwork = new BayesianNetwork(observed.getConnectedGraph());
+        boolean differentiable = DifferentiableChecker.isDifferentiable(bayesianNetwork.getLatentOrObservedVertices());
+        assertEquals(true, differentiable);
```
migwellian: it's probably worth factoring out these 3 lines into `assertIsDiffable(observed.getConnectedGraph())`

keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableChecker.java
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * Checks whether a set of vertices are differentiable
+     * w.r.t the latents in a graph.
+     *
+     * @param vertices the collection of vertices to check
+     * @return true if the vertices are differentiable w.r.t latents
+     */
+    public boolean isDifferentiable(Collection<Vertex> vertices) {
+        Queue<Vertex> queue = new LinkedList<>(vertices);
+        Set<Vertex> queued = new HashSet<>(vertices);
```
migwellian: can you give this a better name?

keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableChecker.java
```diff
@@ -0,0 +1,90 @@
...
+        Set<Vertex> constantValueVerticesCache = new HashSet<>();
```
migwellian: I think "cache" is a misleading name here as it suggests it's just an optimisation and doesn't affect the functionality.

keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableChecker.java
```diff
@@ -0,0 +1,90 @@
...
+                if (!queued.contains(next) && !constantValueVerticesCache.contains(next)) {
+                    queue.offer(next);
+                    queued.add(next);
+                }
+            }
+        }
+        return true;
+    }
+
+    private boolean isVertexConstant(Vertex vertex, Set<Vertex> constantValueVerticesCache) {
```
migwellian: there's an inconsistency to your method names: isVertexConstant vs. isVertexParents*Value*Constant. I'd suggest dropping the word "value" as it doesn't really aid the understandability.

keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableChecker.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.Queue;
+import java.util.Set;
+
+@UtilityClass
+public class DifferentiableChecker {
+
+    /**
```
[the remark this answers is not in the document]
migwellian: +1

keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableChecker.java
```diff
@@ -0,0 +1,90 @@
...
+            if (!visiting.isDifferentiable()) {
+                if (isVertexConstant(visiting, constantValueVerticesCache)) {
+                    continue;
+                } else {
+                    return false;
+                }
+            }
+
+            Collection<Vertex> nextVertices = visiting.getParents();
+            for (Vertex next : nextVertices) {
```
gordoncaleb: If you're always passing in all latent and observed and trying to determine if the total logProb is differentiable then this will always work. If I wanted to know if the LogLikelihood function was differentiable then you'd have to not traverse past variables in this network and only consider the observed vertices. We should chat about this offline.

keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableChecker.java
wesleyw72: Discussed offline. Will include some reasoning in comments as to why this check is necessary.

keanu-project/src/test/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableCheckerTest.java
wesleyw72: Good idea!

keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableChecker.java
wesleyw72: Yes. I have gone `isDifferentiableWrtLatents` hopefully this is a bit better.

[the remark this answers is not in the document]
wesleyw72: Good point, I agree. Have called it `visited`.

keanu-project/src/test/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableCheckerTest.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
```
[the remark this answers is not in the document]
wesleyw72: I have added a mock to ensure that the cache is working by mocking the `isObserved` on a vertex. It would be good to add the check you mention, however as `isDifferentiable()` is final I can't mock it or spy on it.

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/UnifiedMCMCTest.java
```diff
@@ -3,30 +3,30 @@
...
 import static org.junit.Assert.assertTrue;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
 
 public class UnifiedMCMCTest {
 
     @Test
     public void checkMHIsRunForNonDifferentiableNetwork() {
-        BayesianNetwork network = mock(BayesianNetwork.class);
-        when(network.getNonDifferentiableVertices()).thenReturn(singletonList(mock(Vertex.class)));
```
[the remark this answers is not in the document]
wesleyw72: This test mocks the bayes net in order to see whether `getNonDifferentiableVertices` is called which is what used to define a non differentiable net. The new check that this PR implements is more complex, involving exploring the vertices. I think mocking would now make this test more complicated.

keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableChecker.java
wesleyw72: Have made some changes to the algorithm. It now is just an optimisation.

keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableChecker.java
```diff
@@ -0,0 +1,90 @@
...
+                if (!queued.contains(next) && !constantValueVerticesCache.contains(next)) {
+                    queue.offer(next);
+                    queued.add(next);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
wesleyw72: Refactored the parent adding bit into its own method so should make it easier to read.

keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableChecker.java
wesleyw72: Changed the algorithm based on our chat. Hopefully it should cover this case now.

keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableChecker.java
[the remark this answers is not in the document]
wesleyw72: Algorithm has changed a bit, so this method doesn't exist anymore.

keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableChecker.java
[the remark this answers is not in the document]
wesleyw72: Have improved the Javadoc for the class. Hopefully it's a bit clearer how it works now. Could you let me know what you think? I think it makes the test cases clearer but happy to add diagrams if you think that would help still.

docs/current_docs/autocorrelation.md
```diff
@@ -73,4 +73,4 @@ posterior_samples = sample(net=bayes_net, sample_from=bayes_net.get_latent_verti
                            algo="metropolis", draws=100)
 vertex_samples = posterior_samples.get('a')
 ac = stats.autocorrelation(vertex_samples, (0,1))
-```
\ No newline at end of file
+```
```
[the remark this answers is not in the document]
wesleyw72: Yeah this is generated ind the build step. Can't really see the difference, maybe an encoding thing? Anyway, when I merged in develop it disappeared from the diff so the change must have been committed there recently.

keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableChecker.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    private static boolean isUnobservedProbabilistic(Vertex vertex) {
+        return vertex.isProbabilistic() && !vertex.isObserved();
+    }
+
+    // We know whether these are constant. For cases such as a MultiplicationVertex we would need to
+    // explore its parents to ensure its constant.
+    private static boolean isValueKnownToBeConstant(Vertex vertex, Set<Vertex> constantValueVerticesCache) {
+        return vertex instanceof ConstantVertex || vertex.isObserved() || constantValueVerticesCache.contains(vertex);
```
migwellian: the double meaning of "constant" here is yucky, but we have a task on the backlog to rename `ConstantVertex` which will make this better.

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -156,7 +155,7 @@ public void setState(NetworkState state) {
      * @return a list of all vertices that are not differentiable (i.e., there are points at which they do not have a derivative).
      */
     public List<Vertex> getNonDifferentiableVertices() {
-        return vertices.stream().filter(vertex -> !(vertex instanceof Differentiable))
+        return vertices.stream().filter(vertex -> !(((Vertex) vertex).isDifferentiable()))
```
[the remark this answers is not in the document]
wesleyw72: You're right, didn't see that it added that in. However, this PR means this code is not used any more so I have removed the method completely.

keanu-project/src/test/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableCheckerTest.java
```diff
@@ -0,0 +1,198 @@
...
+import java.util.Collection;
+
+import static junit.framework.TestCase.assertTrue;
+import static org.junit.Assert.assertFalse;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+
+public class DifferentiableCheckerTest {
```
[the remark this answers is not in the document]
wesleyw72: Yeah that makes sense. I have added some ascii art to the top of `DifferentiableChecker`.

keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableChecker.java
```diff
@@ -0,0 +1,128 @@
...
+
+    private static boolean diffableOrConstantUptoNextRV(Collection<Vertex> vertices, Set<Vertex> constantValueVerticesCache) {
+        return bfs(vertices,
+            vertex -> isNonDiffableAndNotConstant(vertex, constantValueVerticesCache),
+            vertex -> !vertex.isProbabilistic(),
+            doNothing -> {
+            });
+    }
+
+    private static boolean bfs(Collection<Vertex> vertices,
```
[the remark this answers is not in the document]
wesleyw72: Discussed offline. Have made a BreadthFirstSearch class that separates the BFS done here, which hopefully can be used in future cases. Have added a task in the backlog in order to see if we can reduce the number of (D/B)FS implementations we have into a common implementation.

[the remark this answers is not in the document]
wesleyw72: Put this into a BreadthFirstSearch class.

keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/DifferentiableChecker.java
```diff
@@ -0,0 +1,126 @@
...
+import io.improbable.keanu.vertices.Vertex;
+import io.improbable.keanu.vertices.dbl.DoubleVertex;
+import lombok.experimental.UtilityClass;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
```
gordoncaleb: Looks like the autoformatting has blasted your nice ascii formatting...

keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/BreadthFirstSearch.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Queue;
+import java.util.Set;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+
+public class BreadthFirstSearch {
+
+    public static boolean bfs(Collection<Vertex> vertices,
+                              Predicate<Vertex> failureCondition,
+                              Predicate<Vertex> shouldAddParents,
```
gordoncaleb: You can traverse in either direction. It might not always be right to call it parents, it could be children.

keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/BreadthFirstSearch.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Queue;
+import java.util.Set;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+
+public class BreadthFirstSearch {
+
+    public static boolean bfs(Collection<Vertex> vertices,
```
gordoncaleb: Looks like bfs does more than just traverse and visit. It seems to allow some true/false check. This is fine but we need to do better to describe this in the name of the class and method.

keanu-project/src/main/java/io/improbable/keanu/algorithms/graphtraversal/BreadthFirstSearch.java
wesleyw72: Good point. I will replace this with a `Function<Vertex, Collection<Vertex>>` that will calculate the next vertices to visit. That way the predicate `shouldAddParents` can disappear and instead be represented in the function by returning an empty collection when parents don't need to be added.

PR #425, opened 2019-01-09
Fix progress bar dual percentage displaying

wesleyw72:
In the case that samples are dropped, the progress bar showed two percentages after the dropping had occurred. This is due to the difference in the string length between `Dropping samples...` and `Sampling...` which meant the dropping samples percentage wasn't being cleared. I have fixed this by appending spaces to clear previous output when the previous output was longer. 

This is a bit hacky so open to other ideas. Tried using `\b` as a backspace but that didn't seem to work.
screenshot 2019-01-09 at 11 15 14


alexsparrow:
Looks reasonable to me. Only alternative I could think of would be to make the progress bar fixed width.

PR #427, opened 2019-01-09
Bump gradle to 5.1

alexsparrow:
Just for fun. ~~Not yet tested that the publishing flow still works.~~

The `task`s had to move due to `gradle` no longer deferring execution of the `publishing` block (see https://docs.gradle.org/5.0/userguide/upgrading_version_4.html#potential_breaking_changes).


alexsparrow:
Thanks to @GeorgeNash for testing this out. Added a fix for publishing which now works.

PR #428, opened 2019-01-09
Feature/extra python vertices

christophernorth:
Changes to the Python code to simplify dependencies and add the ability to handle non vertex/const parameters.


keanu-python/keanu/vertex/vertex_helpers.py
```diff
@@ -0,0 +1,54 @@
...
+from py4j.java_collections import JavaArray
+
+k = KeanuContext()
+
+
+def do_vertex_cast(vertex_ctor: Callable, value: vertex_constructor_param_types) -> Vertex:
+    return value if isinstance(value, Vertex) else vertex_ctor(value)
+
+
+def do_generic_vertex_cast(ctors: Dict[type, Callable], value: vertex_constructor_param_types) -> Vertex:
```
icantsayrural: This method name makes it seem like we are specifically casting `GenericVertex` (which I suppose `Vertex` kind of is...? haha)

codegen/src/main/resources/generated.py.ftl
```diff
@@ -1,17 +1,46 @@
...
+def cast_to_integer_vertex(input: vertex_constructor_param_types) -> Vertex:
+    return do_vertex_cast(ConstantInteger, input)
+
+
+def cast_to_boolean_vertex(input: vertex_constructor_param_types) -> Vertex:
+    return do_vertex_cast(ConstantBoolean, input)
+
+
+def cast_to_vertex(input: vertex_constructor_param_types) -> Vertex:
+    return do_generic_vertex_cast({bool: ConstantBoolean, int: ConstantInteger, float: ConstantDouble}, input)
```
icantsayrural: Nice!!

codegen/src/test/java/io/improbable/keanu/codegen/python/GetNonExportedVertices.java
```diff
@@ -0,0 +1,34 @@
...
+
+import io.improbable.keanu.annotation.ExportVertexToPythonBindings;
+import io.improbable.keanu.vertices.Vertex;
+import org.reflections.Reflections;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Modifier;
+import java.util.Set;
+
+public class GetNonExportedVertices {
```
icantsayrural: I guess we are not running this in tests to avoid being to strict? Useful tool!

keanu-python/keanu/vartypes.py
```diff
@@ -16,7 +16,7 @@
 pandas_types = Union[Series, DataFrame]
 numpy_types = Union[ndarray]
 
-tensor_arg_types = Union[primitive_types, pandas_types, numpy_types]
+tensor_arg_types = Union[primitive_types, pandas_types, numpy_types, JavaObjectWrapper, str]
```
migwellian: What's going on here? How/when would one coerce a `JavaObjectWrapper` or string into a tensor?

keanu-python/keanu/vertex/const.py
```diff
@@ -1,49 +1,8 @@
...
-def __infer_const_ctor_from_scalar(scalar: np.generic) -> Callable:
-    if isinstance(scalar, runtime_bool_types):
-        return ConstantBoolean
-    elif isinstance(scalar, runtime_int_types):
-        return ConstantInteger
-    elif isinstance(scalar, runtime_float_types):
-        return ConstantDouble
-    else:
-        raise NotImplementedError("Generic types in an ndarray are not supported. Was given {}".format(type(scalar)))
+    return infer_type_and_execute(t, {bool: ConstantBoolean, int: ConstantInteger, float: ConstantDouble})
```
migwellian: looks like a good refactor

keanu-python/keanu/vertex/vertex_helpers.py
```diff
@@ -0,0 +1,54 @@
+from typing import Callable, Dict, Any, Collection
```
migwellian: the name `vertex_helpers` is very generic - I'd rather call this `vertex_casting` to avoid it becoming a catch-all for miscellaneous static functions.

keanu-python/keanu/vertex/base.py
```diff
@@ -169,9 +169,7 @@ def __parse_args(args: Tuple[Union[vertex_constructor_param_types, shape_types],
 
     @staticmethod
     def __parse_arg(arg: Union[vertex_constructor_param_types, shape_types]) -> JavaObject:
-        if isinstance(arg, runtime_tensor_arg_types):
-            return kn.vertex.const.Const(arg).unwrap()
-        elif isinstance(arg, runtime_wrapped_java_types):
+        if isinstance(arg, runtime_wrapped_java_types):
             return arg.unwrap()
         elif isinstance(arg, collections.Collection) and all(isinstance(x, runtime_primitive_types) for x in arg):
```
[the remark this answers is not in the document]
christophernorth: There is a test that currently checks that a shape can be passed to the Vertex constructor (this is the only way to pass a shape through for random variables at the moment) and I didn't want to remove that ability.  I think we should get rid of this once we have "real" support for specifying a shape as part of the vertex constructors?

keanu-python/keanu/context.py
```diff
@@ -115,6 +115,12 @@ def to_java_array(self, l: Collection[Any], klass: JavaClass = None) -> JavaArra
     def to_java_long_array(self, l: Collection[Any]) -> JavaArray:
         return self.to_java_array(l, self._gateway.jvm.long)
 
+    def to_java_int_array(self, l: Collection[Any]) -> JavaArray:
+        return self.to_java_array(l, self._gateway.jvm.int)
+
+    def to_java_vertex_array(self, l: Collection[Any]) -> JavaArray:
+        return self.to_java_array(l, self._gateway.jvm.Vertex)
```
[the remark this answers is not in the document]
christophernorth: Silly mistake - my test for this didn't get committed so hasn't run (and of course, this doesn't work).  I'll get it fixed.

keanu-python/keanu/vertex/vertex_helpers.py
christophernorth: Yeah - one of the little warts on our class hierarchy.  I'll rename.

codegen/src/main/resources/generated.py.ftl
christophernorth: 👍

codegen/src/test/java/io/improbable/keanu/codegen/python/GetNonExportedVertices.java
christophernorth: Yeah - at the moment,  we don't enforce anything about all vertices having to be exported to Python like we do for Saveability.  It certainly helped when it came time to annotate the extra ones we wanted to export though!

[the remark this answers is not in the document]
christophernorth: It was a useful tool for me to be able to print those vertices that aren't exported to Python.  I'd say it's worth having in the repo just in case people want to do the same in future (it's certainly quicker than checking by hand if you're looking at more than a couple of vertices).  Later I'd envisage this moving to be a full test case in the same way we have one for Saveability.

keanu-python/keanu/vertex/const.py
christophernorth: 👍

keanu-python/keanu/vertex/vertex_helpers.py
christophernorth: I sort of wanted it to be a catch all as it happens (I find bigger modules have less tendency to cause issues with the dependency graph over time :) but happy to rename.

codegen/src/main/java/io/improbable/keanu/codegen/python/VertexProcessor.java
```diff
@@ -114,15 +116,37 @@ private static String toCastedPythonParams(String[] pythonParameters, Class<?>[]
...
+            return "cast_to_string(" + pythonParameter + ")";
+        } else if (Long[].class.isAssignableFrom(parameterType) || long[].class.isAssignableFrom(parameterType)) {
+            return "cast_to_long_array(" + pythonParameter + ")";
+        } else if (Integer[].class.isAssignableFrom(parameterType) || int[].class.isAssignableFrom(parameterType)) {
+            return "cast_to_int_array(" + pythonParameter + ")";
+        } else if (Vertex[].class.isAssignableFrom(parameterType)) {
+            return "cast_to_vertex_array(" + pythonParameter + ")";
         } else {
-            return pythonParameter;
+            throw new IllegalArgumentException("Failed to Encode " + pythonParameter + " of type: " + parameterType);
```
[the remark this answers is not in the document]
christophernorth: As discussed - there's a class hierarchy thing going on at the moment so we can't just use a plain old map without us having an unwieldy maintenance job of putting every vertex in there.  We could write a smarter walker that checks the map - but I'd have to convince myself that would definitely work in all cases before I'd want to make the change (and it would potentially reduce the readability a bit anyway).  I'll leave as is for now, and have a think about how we could better handle this (since we have similar code in the Protobuf stuff too).

keanu-python/keanu/vartypes.py
christophernorth: This is basically here to try and convince mypy to behave tbh in the least impactful way possible.  I'll chat to you offline.

christophernorth: Resolved in a new way (by adding a bit more code) so that it's hopefully a lot cleaner now.  Let me know what you think.

PR #429, opened 2019-01-10
Feature/sampling takes interface

migwellian:
Sampling algorithms no longer take a `BayesianNetwork` and now take a ` ProbabilisticModel`.

- Builds on the work done with Optimizers, specifically the 2 new interfaces that were introduced there:
  - `Variable` (A very stripped down version of Vertex which has no concept of living inside a graph)
  - `ProbabilisticModel` (formerly `ProbabilisticGraph` - actually it is not necessarily a graph). This is able to compute `logProb` and `logLikelihood` across its self but has no graph concepts (e.g: cascading)
- BayesNet-specific logic has been removed from Metropolis Hastings by injecting some objects that implement new interfaces: `LogProbCalculationStrategy`, `ProposalApplicationStrategy` and `ProposalRejectionStrategy`
- A new interface `VariableState` whose purpose is to abstract the concept of state. For a `Vertex` this state comprises two things: `value` and `isObserved`.
- `Hamiltonian` has been deleted

This touches a lot of files but when reviewing you can focus on the following bits:
- `ProbabilisticModel` and its concrete implementation `KeanuProbabilisticModel`
- Everything in the `mcmc` package (in particular `MetropolisHastings` and `NUTS`) to see how the `ProbabilisticModel` and `ProposalApplication` and `ProposalRejection` strategies are used to embed graph concepts (cascading) outside of `ProbabilisticModel`


keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/LambdaSectionOptimizedLogProbCalculator.java
```diff
@@ -0,0 +1,23 @@
...
+
+    private final LambdaSectionSnapshot lambdaSectionSnapshot;
+
+    public LambdaSectionOptimizedLogProbCalculator(List<Vertex> latentVertices) {
+        lambdaSectionSnapshot = new LambdaSectionSnapshot(latentVertices);
+    }
+
+    @Override
+    public double calculate(ProbabilisticModel model, Set<Variable> variables) {
+        return lambdaSectionSnapshot.logProb(variables);
```
migwellian: The `model` argument is unused here. That's because the `latentVertices` passed in the constructor should be those that came from the model. This is not ideal but I haven't thought of a nicer way to do it.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
```diff
@@ -62,113 +57,51 @@ public static MetropolisHastings withDefaultConfig(KeanuRandom random) {
...
-    private boolean useCacheOnRejection = DEFAULT_USE_CACHE_ON_REJECTION;
+    private ProposalRejectionStrategy rejectionStrategy = DEFAULT_REJECTION_STRATEGY;
+
+    @Getter
+    @Setter
+    @Builder.Default
+    private LogProbCalculationStrategy logProbCalculationStrategy = DEFAULT_LOG_PROB_CALCULATION_STRATEGY;
+
+    @Getter
+    @Setter
```
migwellian: I think we should remove `@Setter` from all of the variables in this class. The object can be immutable.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
```diff
@@ -62,113 +57,51 @@ public static MetropolisHastings withDefaultConfig(KeanuRandom random) {
...
+            proposalApplicationStrategy,
             random
         );
 
-        double logProbabilityBeforeStep = bayesianNetwork.getLogOfMasterP();
-
-        return new MetropolisHastingsSampler(latentVertices, verticesToSampleFrom, mhStep, variableSelector, logProbabilityBeforeStep);
-    }
-
-    public static class MetropolisHastingsSampler implements SamplingAlgorithm {
```
migwellian: NB: this has been pulled out into its own file

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
```diff
@@ -1,38 +1,32 @@
...
  */
 @Builder
-@Slf4j
 public class MetropolisHastings implements PosteriorSamplingAlgorithm {
 
     private static final ProposalDistribution DEFAULT_PROPOSAL_DISTRIBUTION = ProposalDistribution.usePrior();
     private static final MHStepVariableSelector DEFAULT_VARIABLE_SELECTOR = SINGLE_VARIABLE_SELECTOR;
-    private static final boolean DEFAULT_USE_CACHE_ON_REJECTION = true;
+    public static final CascadeOnRejection DEFAULT_REJECTION_STRATEGY = new CascadeOnRejection();
+    private static final LogProbCalculationStrategy DEFAULT_LOG_PROB_CALCULATION_STRATEGY = new SimpleLogProbCalculationStrategy();
```
migwellian: These two lines mean that the default has changed. This is because we'd need to know the latent variables to pass into the constructors if we were to have the same defaults (`RollBackOnRejection` and `LambdaSectionOptimizedLogProbCalculator`).

I would like to add/change the factory methods `withDefaultConfig` so that downstream behaviour is as before.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/ProposalApplicationStrategy.java
```diff
@@ -0,0 +1,21 @@
...
+
+public interface ProposalApplicationStrategy {
+
+    /**
+     * Any actions taken after after applying the provided proposal
+     * from a sampling algorithm to its variables.
+     *
+     * No action will be taken in most cases.
+     *
+     * @param proposal  the proposal that has been applied
```
migwellian: `proposal` is never used... have you put it in with a view to what information someone might need if they wanted to write a new one?

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/LambdaSectionSnapshot.java
```diff
@@ -0,0 +1,68 @@
...
+import io.improbable.keanu.vertices.ProbabilityCalculator;
+import io.improbable.keanu.vertices.Vertex;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class LambdaSectionSnapshot {
```
migwellian: NB: this logic was factored out of `MetropolisHastingsStep`

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/ProbabilisticModel.java
```diff
@@ -20,4 +22,5 @@ default double logLikelihood() {
 
     List<? extends Variable> getLatentVariables();
 
+    List<? extends Variable<DoubleTensor, ?>> getContinuousLatentVariables();
```
migwellian: NB: this is called in one place, `NUTS#setupSampler`

migwellian: @GeorgeNash should it be on `ProbabilisticModelWithGradient` instead?

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/Variable.java
```diff
@@ -1,10 +1,19 @@
...
     VariableReference getReference();
 
-    T getValue();
+    VALUE getValue();
+
+    void setValue(VALUE value);
 
     long[] getShape();
+
+    STATE getState();
```
migwellian: NB: This is only called in one place: the ctor for `NetworkSnapshot`

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/Variable.java
```diff
@@ -1,10 +1,19 @@
...
+
+    void setState(STATE state);
```
migwellian: NB: This is only called in one place: `NetworkSnapshot#apply`

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/Variable.java
```diff
@@ -1,10 +1,19 @@
...
+import io.improbable.keanu.vertices.Samplable;
+
+public interface Variable<VALUE, STATE extends VariableState> extends Samplable<VALUE> {
 
     VariableReference getReference();
 
-    T getValue();
+    VALUE getValue();
+
+    void setValue(VALUE value);
```
migwellian: NB: this is called in several places:
* `Proposal#apply` and `Proposal#reject`
* `Leapfrog#stepPosition`
* `NUTS#resetVariableValue`

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/Variable.java
```diff
@@ -1,10 +1,19 @@
 package io.improbable.keanu.algorithms.variational.optimizer;
 
-public interface Variable<T> {
+import io.improbable.keanu.network.VariableState;
+import io.improbable.keanu.vertices.Samplable;
+
+public interface Variable<VALUE, STATE extends VariableState> extends Samplable<VALUE> {
```
migwellian: Note that this now extends `Samplable`. The `sample` method is only called in one place: `PriorProposalDistribution#setFor`.

We could probably eliminate it, although I think conceptually it is valid for a `Variable` to be `Samplable`.

migwellian: NB: the purpose of `VariableState` is to abstract the concept of state. For a `Vertex` its state comprises two things: `value` and `isObserved`.

keanu-project/src/main/java/io/improbable/keanu/model/PredictiveModel.java
```diff
@@ -1,6 +1,6 @@
 package io.improbable.keanu.model;
 
-public interface Model<INPUT, OUTPUT> {
+public interface PredictiveModel<INPUT, OUTPUT> {
```
migwellian: NB: renamed to distinguish this from `ProbabilisticModel` (the two are not related)

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/nuts/NUTS.java
```diff
@@ -85,47 +85,45 @@ public static NUTS withDefaultConfig(KeanuRandom random) {
...
      * @return Samples taken with NUTS
      */
     @Override
-    public NetworkSamples getPosteriorSamples(final BayesianNetwork bayesNet,
-                                              final List<? extends Vertex> sampleFromVertices,
+    public NetworkSamples getPosteriorSamples(final ProbabilisticModel model,
+                                              final List<? extends Variable> variablesToSampleFrom,
                                               final int sampleCount) {
-        return generatePosteriorSamples(bayesNet, sampleFromVertices)
+        return generatePosteriorSamples((ProbabilisticModelWithGradient) model, variablesToSampleFrom)
```
migwellian: This cast is icky... one way to avoid it would be to parametrise `PosteriorSamplingAlgorithm` on `<? extends ProbabilisticModel>` but perhaps that's over-engineering the problem.

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/KeanuProbabilisticModel.java
```diff
@@ -22,25 +25,31 @@
...
-        this.vertexLookup = bayesianNetwork.getLatentVertices().stream()
+    public KeanuProbabilisticModel(BayesianNetwork bayesianNetwork) {
+        this.vertexLookup = bayesianNetwork.getLatentOrObservedVertices().stream()
             .collect(toMap(Vertex::getId, v -> v));
 
         this.latentVertices = ImmutableList.copyOf(bayesianNetwork.getLatentVertices());
         this.observedVertices = ImmutableList.copyOf(bayesianNetwork.getObservedVertices());
         this.latentOrObservedVertices = ImmutableList.copyOf(bayesianNetwork.getLatentOrObservedVertices());
+
+        cascadeObservations();
```
migwellian: @GeorgeNash you were gonna rename this

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/ProposalApplicationStrategy.java
```diff
@@ -0,0 +1,21 @@
+package io.improbable.keanu.algorithms.mcmc;
+
+import io.improbable.keanu.algorithms.mcmc.proposal.Proposal;
+import io.improbable.keanu.algorithms.variational.optimizer.Variable;
+
+import java.util.Set;
+
+public interface ProposalApplicationStrategy {
```
icantsayrural: I wonder if we should rename this? We are using the word `apply` for both a proposal and sampling algorithms.

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/KeanuProbabilisticModel.java
```diff
@@ -49,7 +58,31 @@ public double logLikelihood(Map<VariableReference, ?> inputs) {
...
+
+    private void checkBayesNetInHealthyState() {
+        if (latentOrObservedVertices.isEmpty()) {
+            throw new IllegalArgumentException("Cannot sample from a completely deterministic BayesNet");
+        } else if (ProbabilityCalculator.isImpossibleLogProb(this.logProb())) {
+            throw new IllegalArgumentException("Cannot start optimizer on zero probability network");
+        }
+    }
+
+    private void resetModelToObservedState() {
```
icantsayrural: Why was this added?

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
```diff
@@ -1,37 +1,32 @@
...
  * Metropolis Hastings is a Markov Chain Monte Carlo method for obtaining samples from a probability distribution
  */
 @Builder
-@Slf4j
 public class MetropolisHastings implements PosteriorSamplingAlgorithm {
 
     private static final ProposalDistribution DEFAULT_PROPOSAL_DISTRIBUTION = ProposalDistribution.usePrior();
     private static final MHStepVariableSelector DEFAULT_VARIABLE_SELECTOR = SINGLE_VARIABLE_SELECTOR;
-    private static final boolean DEFAULT_USE_CACHE_ON_REJECTION = true;
+    public static final CascadeOnRejection DEFAULT_REJECTION_STRATEGY = new CascadeOnRejection();
```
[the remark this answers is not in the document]
migwellian: good spot

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/ProposalApplicationStrategy.java
migwellian: I don't think there's anything wrong with the interface name `ProposalApplicationStrategy`, i.e. it is the strategy for what to do when a proposal is applied.

We could rename the method `ProposalApplicationStrategy#apply`.

Unless you mean something else?

icantsayrural: Yeah, the method name change could also do

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastingsStep.java
```diff
@@ -1,89 +1,87 @@
...
 @Slf4j
-class MetropolisHastingsStep {
+public class MetropolisHastingsStep {
 
     //Temperature for standard MH step accept/reject calculation
     private static final double DEFAULT_TEMPERATURE = 1.0;
 
+    private final ProbabilisticModel model;
     private final ProposalDistribution proposalDistribution;
-    private final boolean useCacheOnRejection;
```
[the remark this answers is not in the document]
migwellian: That's not quite accurate. We always used `LambdaSectionOptimizedLogProbCalculator`, regardless of the value of the `useCacheOnRejection` flag.

The `LambdaSectionOptimizedLogProbCalculator` is just an optimisation, so the answer to your question is yes, it is valid to mix and match.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/SimulatedAnnealing.java
```diff
@@ -60,70 +55,76 @@ public static SimulatedAnnealing withDefaultConfig(KeanuRandom random) {
...
-                setSamplesAsMax(maxSamplesByVertex, latentVertices);
+                setSamplesAsMax(maxSamplesByVariable, latentVariables);
             }
         }
 
-        return new SimpleNetworkState(maxSamplesByVertex);
+        return new SimpleNetworkState(mapByVariableReference(maxSamplesByVariable));
+    }
+
+    private <T> Map<VariableReference, T> mapByVariableReference(Map<VariableReference, T> legacyMap) {
```
icantsayrural: Isn't the input already mapped by `VariableReference`?

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/SimpleLogProbCalculationStrategy.java
```diff
@@ -0,0 +1,13 @@
...
+
+import io.improbable.keanu.algorithms.variational.optimizer.ProbabilisticModel;
+import io.improbable.keanu.algorithms.variational.optimizer.Variable;
+
+import java.util.Set;
+
+public class SimpleLogProbCalculationStrategy implements LogProbCalculationStrategy {
+    @Override
+    public double calculate(ProbabilisticModel model, Set<Variable> variables) {
+        return model.logProb();
```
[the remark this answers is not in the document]
migwellian: No, it is valid to calculate the `logProb` of the entire graph and ignore the `variables` argument. Note that we do this same calculation before and after applying the proposal.

Tagging @gordoncaleb to sanity check that what I am saying here is true.

docs/current_docs/autocorrelation.md
```diff
@@ -35,9 +35,9 @@ calculating autocorrelation on samples.
 With a network defined, we can get the autocorrelation vertex A. The result is 
 a tensor containing the autocorrelation at varying lags.
 ```java
-NetworkSamples posteriorSamples = MetropolisHastings.withDefaultConfig().getPosteriorSamples(
-    bayesNet,
-    bayesNet.getLatentVertices(),
+NetworkSamples posteriorSamples = KeanuMetropolisHastings.withDefaultConfigFor(model).getPosteriorSamples(
```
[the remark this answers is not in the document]
migwellian: we now have both and `MetropolisHastings` (which, as before, implements the `PosteriorSamplingAlgorithm` algorithm but has no dependency on the `io.improbable.keanu.vertices` package) `KeanuMetropolisHastings` (a Keanu-specific utility class that just contains two factory methods).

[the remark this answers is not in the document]
migwellian: I like your idea of `Keanu.Sampling...` as an easy lookup of all the factory methods.

docs/src/test/java/io/improbable/snippet/WetGrass.java
```diff
@@ -55,8 +55,9 @@ public void doesFindCorrectProbability() {
 
         //What does that observation say about the probability that it rained or that
         //the sprinkler was on?
-        NetworkSamples posteriorSamples = MetropolisHastings.withDefaultConfig().getPosteriorSamples(
-            new BayesianNetwork(wetGrass.getConnectedGraph()),
+        KeanuProbabilisticModel model = new KeanuProbabilisticModel(wetGrass.getConnectedGraph());
+        NetworkSamples posteriorSamples = KeanuMetropolisHastings.withDefaultConfigFor(model).getPosteriorSamples(
+            model,
```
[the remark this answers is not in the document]
migwellian: In the factory method, all we need is `model.getLatentVertices()` so we could pass that in instead if you like?

In the call to `getPosteriorSamples` - that is defined by the API and I don't think we want to change that.

keanu-project/src/main/java/io/improbable/keanu/algorithms/NetworkSamples.java
```diff
@@ -24,67 +24,67 @@
...
         samplesForVertex.add(value);
     }
 
     public int size() {
         return this.size;
     }
 
-    public <T> VertexSamples<T> get(Vertex<T> vertex) {
-        return get(vertex.getId());
+    public <T> VertexSamples<T> get(Variable<T, ?> variable) {
```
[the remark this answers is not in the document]
migwellian: Good idea... or even just `Samples`? It's parametrised, not even specific to `Variable`

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
```diff
@@ -1,47 +1,35 @@
...
 import static io.improbable.keanu.algorithms.mcmc.proposal.MHStepVariableSelector.SINGLE_VARIABLE_SELECTOR;
 
 /**
  * Metropolis Hastings is a Markov Chain Monte Carlo method for obtaining samples from a probability distribution
  */
 @Builder
-@Slf4j
 public class MetropolisHastings implements PosteriorSamplingAlgorithm {
 
-    private static final ProposalDistribution DEFAULT_PROPOSAL_DISTRIBUTION = ProposalDistribution.usePrior();
```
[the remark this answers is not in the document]
migwellian: Because the alternative, `PriorProposalDistribution` now requires a list of latent vertices when you construct it.

Perhaps it would be better to have no default.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastingsStep.java
```diff
@@ -1,89 +1,87 @@
...
      * @return the log probability of the network after either accepting or rejecting the sample
      */
-    public StepResult step(final Set<Vertex> chosenVertices,
+    public StepResult step(final Set<Variable> chosenVariables,
                            final double logProbabilityBeforeStep,
                            final double temperature) {
 
-        log.trace(String.format("Chosen vertices: %s", chosenVertices.stream()
-            .map(Vertex::toString)
+        log.trace(String.format("Chosen variables: %s", chosenVariables.stream()
```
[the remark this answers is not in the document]
migwellian: Good call - I will remove it.

How do you feel about the (simpler) trace logging on lines 98-102?

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/KeanuProbabilisticModel.java
```diff
@@ -49,7 +58,31 @@ public double logLikelihood(Map<VariableReference, ?> inputs) {
...
+    public List<? extends Variable<DoubleTensor, ?>> getContinuousLatentVariables() {
+        return getLatentVariables().stream()
+            .filter(v -> v.getValue() instanceof DoubleTensor)
+            .map(v -> (Variable<DoubleTensor, ?>) v)
+            .collect(Collectors.toList());
+    }
+
+    private void checkBayesNetInHealthyState() {
+        if (latentOrObservedVertices.isEmpty()) {
+            throw new IllegalArgumentException("Cannot sample from a completely deterministic BayesNet");
```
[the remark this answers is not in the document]
GeorgeNash: Good spot

keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/LambdaSectionSnapshot.java
```diff
@@ -0,0 +1,68 @@
...
+    private static final boolean USE_CACHE_ON_REJECTION = true;
```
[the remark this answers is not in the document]
migwellian: 👍

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastingsStep.java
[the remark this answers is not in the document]
migwellian: I have removed both

keanu-project/src/main/java/io/improbable/keanu/algorithms/NetworkSamples.java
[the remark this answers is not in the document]
GeorgeNash: Renamed to `Samples`

docs/src/test/java/io/improbable/snippet/WetGrass.java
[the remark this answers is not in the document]
gordoncaleb: Hmm I guess configuring a sampler and getting samples are two different things. What happens if we configure it for a one model and then sample from another?

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/MetropolisHastings.java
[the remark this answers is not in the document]
gordoncaleb: I think it makes sense to be able to default to one or the other. Being forced to default to on instead of the other is a smell that we're building the proposal algorithm before we have enough information.

PR #430, opened 2019-01-10
Add Print Vertex

alex-remedios-improbable:
Open Q's
1. Added a test library which I'm a big fan of, any objections?
2. Have not implemented save, would be cool to talk through how that stuff works
3. Have not seen a new python class show up yet, need to follow up on this
4. This is only for DoubleVertexes, I have not noticed a simple way of making it more generic and am open to suggestions

SOL-2955


keanu-project/build.gradle
```diff
@@ -38,6 +38,7 @@ dependencies {
     //testing
     testCompile 'junit:junit:4.12'
     testCompile 'org.mockito:mockito-core:2.19.1'
+    testCompile 'org.assertj:assertj-core:3.11.1'
```
gordoncaleb: We’ve been using Hamcrest, which is a slightly older assertion lib with similar functionality. They have a very similar api most the time but assertj does a nicer api for more complex assertions. I’ve used both libs before and I could go either way with it if we didnt already have quite a few tests that use Hamcrest. You could make an argument to have both but for simplicity sake I’d say just go with one.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
```diff
@@ -49,6 +49,7 @@
 import io.improbable.keanu.vertices.dbl.nonprobabilistic.operators.unary.TakeVertex;
 import io.improbable.keanu.vertices.dbl.nonprobabilistic.operators.unary.TanVertex;
 
+import io.improbable.keanu.vertices.utility.PrintVertex;
```
gordoncaleb: The verticies package is organized by data type and not function. This print would apply to all types so I’d probably put it in the generic package.

keanu-project/src/main/java/io/improbable/keanu/vertices/utility/PrintVertex.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Override
+    public DoubleTensor sample(KeanuRandom random) {
+        return parent.sample();
+    }
+
+    @Override
+    public DoubleTensor calculate() {
+        final String label = parent.getLabel() == null ? "<no label>" : parent.getLabel().getUnqualifiedName();
+        final String statement = String.format("Calculated Vertex: (label: %s, data: %s)", label,  parent.getValue());
```
gordoncaleb: I think the print shouldn’t have an opinion on the string format. Or at least the default opinion should be minimal. Maybe just the label and value? The user should also be able to specify (somehow) exactly how the print happens.

keanu-project/src/main/java/io/improbable/keanu/vertices/utility/PrintVertex.java
```diff
@@ -0,0 +1,42 @@
...
+        printStream.println(statement);
```
gordoncaleb: I’d go for just .print(...) and add a new line to the string being printed. That way if you had the ability to configure what’s printed you could also control whether there’s even a new line added.

keanu-project/src/main/java/io/improbable/keanu/vertices/utility/PrintVertex.java
```diff
@@ -0,0 +1,42 @@
...
+import com.google.common.base.Preconditions;
+import io.improbable.keanu.annotation.ExportVertexToPythonBindings;
+import io.improbable.keanu.tensor.dbl.DoubleTensor;
+import io.improbable.keanu.vertices.NonProbabilistic;
+import io.improbable.keanu.vertices.NonSaveableVertex;
+import io.improbable.keanu.vertices.dbl.DoubleVertex;
+import io.improbable.keanu.vertices.dbl.KeanuRandom;
+import java.io.PrintStream;
+
+public class PrintVertex extends DoubleVertex implements NonProbabilistic<DoubleTensor>, NonSaveableVertex {
```
gordoncaleb: I think this can be a generic vertex. You don’t care about the data type of your parent. This also means that the .print() method might make sense to be on the base vertex class, which is interesting and we might need to have a think/chat about that.

keanu-project/src/test/java/io/improbable/keanu/vertices/utility/PrintVertexTest.java
```diff
@@ -0,0 +1,85 @@
...
+    @Mock
+    PrintStream printStream;
+
+    @Before
+    public void setup() {
+        MockitoAnnotations.initMocks(this);
+    }
+
+    @Test
+    public void testWhenCreatedThenParentIsSet() {
```
gordoncaleb: This might be in our codebase in some places but I like to avoid prepending every test name with “test”. I used to try to name them using the Given When Then approach but that’s not always practical in camelCase. https://martinfowler.com/bliki/GivenWhenThen.html

keanu-project/src/test/java/io/improbable/keanu/vertices/utility/PrintVertexTest.java
```diff
@@ -0,0 +1,85 @@
...
+    public void testWhenParentHasLabelThenItIsPrinted() {
+        final DoubleVertex parent = new ConstantDoubleVertex(30);
+        String label = "my vertex";
+        parent.setLabel(label);
+        final PrintVertex pv = new PrintVertex(parent, printStream);
+
+        pv.calculate();
+
+        final String expected = "Calculated Vertex: (label: my vertex, data: {\ndata = [30.0]\nshape = []\n})";
+        verify(printStream).println(expected);
```
gordoncaleb: Is one of the requirements that .println(...) is called? You could use a print stream that writes to a String instead of a mock and then assert what that string should be.

keanu-project/build.gradle
alex-remedios-improbable: sounds good, switching to hamcrest

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
```diff
@@ -58,6 +59,10 @@ public DoubleVertex(long[] initialShape) {
         super(initialShape);
     }
 
+    public DoubleVertex print() {
```
migwellian: We should also have a `print(PrintStream)` method here

keanu-project/src/main/java/io/improbable/keanu/vertices/utility/PrintVertex.java
```diff
@@ -0,0 +1,42 @@
...
+    private PrintStream printStream = System.out;
+    private final DoubleVertex parent;
+
+    @ExportVertexToPythonBindings
+    public PrintVertex(DoubleVertex parent, final PrintStream printStream) {
```
migwellian: I think the superclass `OutputStream` would be more appropriate than `PrintStream` here.

`PrintStream` is a `FilterOutputStream`. From its javadoc:

> These streams sit on top of an already existing output stream (the <i>underlying</i> output stream) which it uses as its basic sink of data, but possibly transforming the data along the way or providing additional functionality.

We don't need/want this, we just want to write some text to an `OutputStream`.

keanu-project/src/test/java/io/improbable/keanu/vertices/utility/PrintVertexTest.java
alex-remedios-improbable: 👍

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
alex-remedios-improbable: cool, moved to 
`io.improbable.keanu.vertices.generic.nonprobabilistic`

keanu-project/src/main/java/io/improbable/keanu/vertices/utility/PrintVertex.java
alex-remedios-improbable: made this configurable

keanu-project/src/main/java/io/improbable/keanu/vertices/utility/PrintVertex.java
alex-remedios-improbable: made generic as discussed offline

alex-remedios-improbable:
Hopefully we can wrap up most comments at this stage...couple of TODOs remaining

1. Test coverage on `print()` may require duplicate some tests from `DoubleVertex` to Integer/Bool vertices. Otherwise reworking this a little bit
2. CI style check is failing
3. Save/Load missing string param implementation

keanu-project/src/main/java/io/improbable/keanu/vertices/generic/nonprobabilistic/PrintVertex.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Override
+    public T sample(KeanuRandom random) {
+        return parent.sample();
+    }
+
+    @Override
+    public T calculate() {
+        final String dataOutput = printData ? parent.getValue().toString() + "\n" : "";
+        printStream.print(message + dataOutput);
```
wesleyw72: I think we could print out the vertex label for the parent as well.

keanu-project/src/test/java/io/improbable/keanu/vertices/generic/nonprobabilistic/PrintVertexTest.java
```diff
@@ -0,0 +1,144 @@
...
+        new PrintVertex<>(temperature);
+
+        final BayesianNetwork bayesNet = new BayesianNetwork(temperature.getConnectedGraph());
+
+        final int nSamples = 10000;
+        MetropolisHastings
+            .withDefaultConfig()
+            .getPosteriorSamples(bayesNet, bayesNet.getLatentVertices(), nSamples);
+
+        verify(printStream, atLeast(nSamples)).print(any(String.class));
```
wesleyw72: Can use `anyString()` from `org.mockito.ArgumentMatchers.anyString;`.

keanu-project/src/main/java/io/improbable/keanu/vertices/generic/nonprobabilistic/PrintVertex.java
```diff
@@ -0,0 +1,74 @@
...
+        setParents(parent);
+    }
+
+    public PrintVertex(Vertex<T> parent) {
+        this(parent, "Calculated Vertex:\n", true);
+    }
+
+    @Override
+    public T sample(KeanuRandom random) {
+        return parent.sample();
```
wesleyw72: Pass through the `KeanuRandom` to the `parent.sample()`.

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
```diff
@@ -221,6 +222,16 @@ public void addParent(Vertex<?> parent) {
         return parents;
     }
 
+    public Vertex<T> print() {
+        new PrintVertex<>(this);
+        return this;
+    }
+
+    public Vertex print(final String message, final boolean printData) {
```
wesleyw72: Should this have the generic type `Vertex<T>`?

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
```diff
@@ -221,6 +222,16 @@ public void addParent(Vertex<?> parent) {
         return parents;
     }
 
+    public Vertex<T> print() {
```
wesleyw72: If you do the following you can avoid overriding on the subclasses.
```
public <V extends Vertex<T>> V print() {
        new PrintVertex<>(this);
        return (V) this;
    }
```

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/BoolVertex.java
```diff
@@ -30,6 +30,18 @@ public BoolVertex(long[] initialShape) {
         super(initialShape);
     }
 
+    @Override
+    public BoolVertex print() {
```
wesleyw72: This has the same functionality as the print on the `Vertex` class so I think you don't need to override. See my comment on `Vertex.java` for something that should work.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
wesleyw72: This has the same functionality as the print on the `Vertex` class so I think you don't need to override. See my comment on `Vertex.java` for something that should work.

keanu-project/src/main/java/io/improbable/keanu/vertices/intgr/IntegerVertex.java
```diff
@@ -39,6 +40,18 @@ public IntegerVertex(long[] shape) {
         super(shape);
     }
 
+    @Override
+    public IntegerVertex print() {
```
wesleyw72: This has the same functionality as the print on the `Vertex` class so I think you don't need to override. See my comment on `Vertex.java` for something that should work.

keanu-project/src/test/java/io/improbable/keanu/vertices/generic/nonprobabilistic/PrintVertexTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+    @Test
+    public void whenRunningMetropolisHastingsThenSamplesArePrinted() {
+        final Vertex<DoubleTensor> temperature = new UniformVertex(20., 30.);
+
+        new PrintVertex<>(temperature);
+
+        final BayesianNetwork bayesNet = new BayesianNetwork(temperature.getConnectedGraph());
+
+        final int nSamples = 10000;
```
wesleyw72: Can you lower this sample amount? We can verify the functionality with 10 samples and it will mean the tests run quicker.

keanu-project/src/test/java/io/improbable/keanu/vertices/generic/nonprobabilistic/PrintVertexTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+        assertThat(sut.getParents(), equalTo(ImmutableSet.of(parent)));
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void whenCreatedWithNullParentThenFails() {
+        new PrintVertex<>(null);
+    }
+
+    @Test(expected = NullPointerException.class)
```
wesleyw72: See above re: `ExpectedException.

keanu-project/src/test/java/io/improbable/keanu/vertices/generic/nonprobabilistic/PrintVertexTest.java
```diff
@@ -0,0 +1,144 @@
...
+    public void whenCreatedWithNullPrefixParentThenFails() {
+        final DoubleVertex parent = new ConstantDoubleVertex(30);
+        new PrintVertex<>(parent, null, false);
+    }
+
+    @Test(expected = NullPointerException.class)
```
wesleyw72: See above re: `ExpectedException.

keanu-project/src/test/java/io/improbable/keanu/vertices/generic/nonprobabilistic/PrintVertexTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+        final int nSamples = 10000;
+        MetropolisHastings
+            .withDefaultConfig()
+            .getPosteriorSamples(bayesNet, bayesNet.getLatentVertices(), nSamples);
+
+        verify(printStream, atLeast(nSamples)).print(any(String.class));
+    }
+
+    private static class PlusOneOp extends UnaryOpVertex<IntegerTensor, IntegerTensor> implements NonSaveableVertex {
```
wesleyw72: Why do you need to wrap the plus one in a new vertex?

keanu-project/src/main/java/io/improbable/keanu/vertices/generic/nonprobabilistic/PrintVertex.java
alex-remedios-improbable: nice one! Fixed + new test added

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
alex-remedios-improbable: Ah great, have removed implementation from subclasses

keanu-project/src/test/java/io/improbable/keanu/vertices/generic/nonprobabilistic/PrintVertexTest.java
alex-remedios-improbable: I'm still trying to comprehend the class hierarchies and am struggling to find an existing vertex which allows a basic arithmetic operation on a `GenericVertex` (which `PrintVertex` extends)

keanu-project/src/main/java/io/improbable/keanu/vertices/generic/nonprobabilistic/PrintVertex.java
alex-remedios-improbable: I think, as Caleb noted, there's value in keeping this vertex relatively unopinionated. The caller is free to put the vertex label into the print message, I believe the label is unlikely to change throughout the trace so it would work okay.

I'm pretty on the fence but would rather start minimal and add more things if enough people would find it useful.

keanu-project/src/test/java/io/improbable/keanu/vertices/generic/nonprobabilistic/PrintVertexTest.java
wesleyw72: Ah I see. Yeah I don't think there is one. So what you've done makes sense.

keanu-project/src/main/java/io/improbable/keanu/vertices/generic/nonprobabilistic/PrintVertex.java
wesleyw72: Yeah that makes sense.

keanu-project/src/main/java/io/improbable/keanu/vertices/utility/PrintVertex.java
migwellian: ok fair enough

keanu-project/src/test/java/io/improbable/keanu/vertices/generic/nonprobabilistic/PrintVertexTest.java
```diff
@@ -0,0 +1,196 @@
...
+
+    @Test
+    public void whenPrintStreamIsNullThenFails() {
+        thrown.expect(NullPointerException.class);
+        PrintVertex.setPrintStream(null);
+    }
+
+    @Test
+    public void whenOnlyParentIsSpecifiedThenDefaultPresentationIsUsed() {
+        final Vertex<DoubleTensor> parent = new ConstantDoubleVertex(30);
```
migwellian: This is a good place to use a mock. If the implementation of `ConstantDoubleVertex.toString()` changed (which is quite likely) then this test would fail, but we don't actually care about `ConstantDoubleVertex` in this test. It's enough to mock a `Vertex` and stub its `toString` method.

PR #432, opened 2019-01-11
Feature/vertex state

migwellian:
* This PR introduces a new class VertexState which is an abstraction of the state we currently store in a Vertex (i.e. value and "is observed"). 
* I use it in NetworkSnapshot. This is part of the refactoring work we are doing to make the sampling algorithms take an interface instead of the concrete Keanu Vertex class.
* This also enabled me to delete a couple of classes, Observation and NotObservable, which simplifies the code.


keanu-project/src/main/java/io/improbable/keanu/algorithms/variational/optimizer/Variable.java
```diff
@@ -1,10 +1,16 @@
...
 
     VariableReference getReference();
 
     T getValue();
 
     long[] getShape();
+
+    VariableState getState();
+
+    void setState(VariableState variableState);
```
gordoncaleb: I was hoping to keep Variable very functionally limited. Putting set here opens some complicated questions around setting and cascading or just setting. If I do a set on this variable does it affect other variables? If not, then this might get confusing and if so then this implies some sort of value propagation algorithm operating on the new state, which is basically the Vertex object we already have.

keanu-project/src/main/java/io/improbable/keanu/network/NetworkSnapshot.java
```diff
@@ -1,44 +1,36 @@
...
+import java.util.Map;
+import java.util.Set;
 
 /**
  * Saves the state (value and observed) of a specified collection of vertices.
  */
 public class NetworkSnapshot {
 
-    public static NetworkSnapshot create(Set<Vertex> vertices) {
+    public static NetworkSnapshot create(Set<? extends Variable> vertices) {
```
gordoncaleb: Can we chat about this offline? This makes sense but we might be able to avoid needing to go this route.

keanu-project/src/main/java/io/improbable/keanu/network/NetworkSnapshot.java
```diff
@@ -1,44 +1,36 @@
...
         return new NetworkSnapshot(vertices);
     }
 
-    private final Map<Vertex, Object> values;
-    private final Set<Vertex> observed;
+    private final Map<Variable, VariableState> values;
```
GeorgeNash: nit: rename this to `states` or `vertexState` etc

keanu-project/src/main/java/io/improbable/keanu/network/VariableState.java
```diff
@@ -0,0 +1,4 @@
+package io.improbable.keanu.network;
+
+public interface VariableState {
```
GeorgeNash: I'm not a big fan of marker interfaces, could we create a `VariableState` annotation and add it to the `VertexState` class?

keanu-project/src/main/java/io/improbable/keanu/vertices/Observable.java
```diff
@@ -14,17 +14,11 @@
 
     boolean isObserved();
 
-    static <T> Observable<T> observableTypeFor(Class<? extends Vertex> v) {
-
+    static boolean isObservable(Class<? extends Vertex> v) {
```
GeorgeNash: Feels like `isObservable` should live on `Vertex` rather than on `Observable`

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexState.java
```diff
@@ -0,0 +1,24 @@
...
+import lombok.Value;
+
+import java.util.Optional;
+
+@Value
+public class VertexState<T> implements VariableState {
+    private T value;
+    private boolean isObserved;
+
+    public static final <T> VertexState<T> nullState() {
```
GeorgeNash: what does `final` do when it's on a `static`?

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
```diff
@@ -102,17 +102,32 @@ public final boolean isProbabilistic() {
...
+        return hasValue() ? state.getValue() : lazyEval();
+    }
+
+    @Override
+    public VertexState getState() {
+        return state;
+    }
+
+    @Override
+    public void setState(VariableState newState) {
```
GeorgeNash: Will anyone ever pass in a `VariableState`? Can we remove the need for:

```
public void setState(VertexState<T> newState) {
        state = newState;
    }
```

If you changed the parameter signature of `setState` to: 
```
<U extends VariableState> void setState(U variableState);
```

keanu-project/src/test/java/io/improbable/keanu/network/NetworkSnapshotTest.java
```diff
@@ -0,0 +1,115 @@
...
+        assertThat(vertex.getValue(), equalTo(originalValue));
+        assertThat(vertex.isObserved(), is(true));
+
+        NetworkSnapshot snapshot = NetworkSnapshot.create(ImmutableSet.of(vertex));
+
+        vertex.unobserve();
+        assertThat(vertex.getValue(), equalTo(originalValue));
+        assertThat(vertex.isObserved(), is(false));
+
+        vertex.setValue(otherValue);
```
GeorgeNash: nit: you're already testing the changing of the value in the previous 2 tests. You only need the check on `isObserved` here

keanu-project/src/main/java/io/improbable/keanu/vertices/Observable.java
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/network/VariableState.java
migwellian: done

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexState.java
migwellian: nothing... that was a hangover from when I originally had it as a field rather than a method.

removed.

keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
migwellian: done

migwellian:
closing in lieu of https://github.com/improbable-research/keanu/pull/429

PR #439, opened 2019-01-16
Feature/python nuts params

christophernorth:
Made NUTS parameters settable from the python API (and slightly changed how the python "sample" API works so that it now takes an optional PosteriorSamplingAlgorithm object that represents the constructed sampler.

At the moment if the user doesn't specify an explicit sampler we just use metropolis hastings with default params, but this should probably change to use something a little more automagic (ie based on  differentiability).

Note - I haven't added customisable options for Hamiltonian just because it may go away.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/PosteriorSamplingAlgorithm.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/PosteriorSamplingAlgorithm.java
@@ -1,5 +1,6 @@
 package io.improbable.keanu.algorithms;
 
+import io.improbable.keanu.algorithms.mcmc.NetworkSamplesGenerator;
 import io.improbable.keanu.network.BayesianNetwork;
 import io.improbable.keanu.vertices.Vertex;
 
@@ -21,5 +22,8 @@ public interface PosteriorSamplingAlgorithm {
     default NetworkSamples getPosteriorSamples(BayesianNetwork bayesianNetwork, int sampleCount) {
         return getPosteriorSamples(bayesianNetwork, bayesianNetwork.getTopLevelLatentVertices(), sampleCount);
     }
+
+    NetworkSamplesGenerator generatePosteriorSamples(final BayesianNetwork bayesianNetwork,
+                                                     final List<? extends Vertex> verticesToSampleFrom);
 
 }
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/PosteriorSamplingAlgorithm.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/PosteriorSamplingAlgorithm.java
@@ -15,9 +15,17 @@ public interface PosteriorSamplingAlgorithm {
         return getPosteriorSamples(bayesianNetwork, Collections.singletonList(vertexToSampleFrom), sampleCount);
     }
 
-    NetworkSamples getPosteriorSamples(BayesianNetwork bayesNet,
-                                       List<? extends Vertex> verticesToSampleFrom,
-                                       int sampleCount);
+    /**
+     * @param bayesNet      a bayesian network containing latent vertices
+     * @param verticesToSampleFrom the vertices to include in the returned samples
+     * @param sampleCount          number of samples to take using the algorithm
+     * @return Samples for each vertex ordered by MCMC iteration
+     */
+    default NetworkSamples getPosteriorSamples(BayesianNetwork bayesNet,
+                                               List<? extends Vertex> verticesToSampleFrom,
+                                               int sampleCount) {
+        return generatePosteriorSamples(bayesNet, verticesToSampleFrom).generate(sampleCount);
+    }
 
     default NetworkSamples getPosteriorSamples(BayesianNetwork bayesianNetwork, int sampleCount) {
         return getPosteriorSamples(bayesianNetwork, bayesianNetwork.getTopLevelLatentVertices(), sampleCount);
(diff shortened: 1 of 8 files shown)


keanu-python/tests/test_sampling.py
```diff
@@ -161,28 +155,17 @@ def test_can_track_acceptance_rate_when_iterating(net: BayesNet) -> None:
...
-    with pytest.raises(TypeError) as excinfo:
-        sample(
-            net=net,
-            sample_from=net.get_latent_vertices(),
-            algo="hamiltonian",
-            proposal_listeners=[AcceptanceRateTracker()],
-            drop=3)
-    assert str(excinfo.value) == "Only Metropolis Hastings supports the proposal_listeners parameter"
 
+def test_can_specify_nuts_params(net: BayesNet):
```
wesleyw72: nit: Can you add `-> None`

christophernorth: 👍

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -17,31 +16,90 @@
...
+        self._sampler = sampler
+
+    def get_sampler(self) -> JavaObject:
+        return self._sampler
+
+
+class MetropolisHastingsSampler(PosteriorSamplingAlgorithm):
+
+    def __init__(self,
+                 proposal_distribution: str = None,
```
GeorgeNash: I know that the `.builder()` sets all of the defaults defined in Java but they feel very hidden from a python user. What're your thoughts on adding all the defaults to the arguments here too? Means we have another thing to monitor across both API's but it'll be clearer for python users

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -17,31 +16,90 @@
...
+
+class HamiltonianSampler(PosteriorSamplingAlgorithm):
+
+    def __init__(self) -> None:
+        super().__init__(k.jvm_view().Hamiltonian.withDefaultConfig())
+
+
+class NUTSSampler(PosteriorSamplingAlgorithm):
+
+    def __init__(self,
```
GeorgeNash: same comment about defaults

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -17,31 +16,90 @@
...
            draws: int = 500,
            drop: int = 0,
            down_sample_interval: int = 1,
            plot: bool = False,
            ax: Any = None) -> sample_types:
 
-    sampling_algorithm: JavaObject = build_sampling_algorithm(algo, proposal_distribution, proposal_distribution_sigma,
-                                                              proposal_listeners)
+    if sampling_algorithm is None:
+        sampling_algorithm = MetropolisHastingsSampler()
```
GeorgeNash: add this to the `sample` method?

christophernorth: Only if we want to change the semantics and share the "default" MetropolisHastingsSampler across all calls (which confuses some of the tests).  Also, it's generally considered bad form to create "mutable" objects as part of a default parameter in Python so I'd prefer to keep as is.

keanu-project/src/main/java/io/improbable/keanu/algorithms/PosteriorSamplingAlgorithm.java
```diff
@@ -22,4 +23,7 @@ default NetworkSamples getPosteriorSamples(BayesianNetwork bayesianNetwork, int
         return getPosteriorSamples(bayesianNetwork, bayesianNetwork.getTopLevelLatentVertices(), sampleCount);
     }
 
+    NetworkSamplesGenerator generatePosteriorSamples(final BayesianNetwork bayesianNetwork,
+                                                     final List<? extends Vertex> verticesToSampleFrom);
+
```
migwellian: now you can give the other method a default version:

    default NetworkSamples getPosteriorSamples(BayesianNetwork bayesNet,
                                       List<? extends Vertex> verticesToSampleFrom,
                                       int sampleCount) {
        return generatePosteriorSamples(bayesNet, verticesToSampleFrom)
            .generate(sampleCount);
    }

christophernorth: Good catch - will update

keanu-python/keanu/algorithm/sampling.py
christophernorth: As discussed - we should raise a ticket to document these things in Python better.  Deferring that till then.

keanu-python/keanu/algorithm/sampling.py
christophernorth: As above.

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -17,31 +16,90 @@
...
+                type_=proposal_distribution, sigma=proposal_distribution_sigma, listeners=proposal_listeners)
+            builder = builder.proposalDistribution(proposal_distribution_object.unwrap())
+
+        if use_cache_on_rejection is not None:
+            builder.useCacheOnRejection(use_cache_on_rejection)
+
+        super().__init__(builder.build())
+
+
+class HamiltonianSampler(PosteriorSamplingAlgorithm):
```
gordoncaleb: We can just remove Hamilitonian. Might want to do it in another PR if it causes much extra effort.

christophernorth: Yeah - I think it's best as its own PR.  Thanks for the review.

PR #454, opened 2019-01-22
Feature/python if syntax

migwellian:
enables Python users to write `If(predicate, thn, els)` and the type (Double/Integer/Boolean) gets inferred


keanu-python/keanu/vertex/iff.py
```diff
@@ -0,0 +1,39 @@
+from typing import Union
+
+from keanu.infer_type import get_type_of_value
+from .base import Boolean
+from .vertex_casting import do_inferred_vertex_cast
+from .base import Vertex, vertex_constructor_param_types
+from .generated import BooleanIf, DoubleIf, IntegerIf, ConstantBoolean, ConstantInteger, ConstantDouble
+
+
+def __cast_to_vertex(input: vertex_constructor_param_types) -> Vertex:
```
migwellian: @christophernorth - this is duplicated from `generated.py`. Before moving it somewhere else I wanted to check with you about the work you did to simplify the dependency tree.

christophernorth: Let's chat - but if you depend on .generated - you should be able to just use the version from there rather than copying it.

keanu-python/keanu/vertex/iff.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+from keanu.vartypes import tensor_arg_types
+
+
+def If(predicate: Union[tensor_arg_types, Vertex], thn: Union[tensor_arg_types, Vertex],
+       els: Union[tensor_arg_types, Vertex]) -> Vertex:
+    predicate = __cast_to_vertex(predicate)
+    thn = __cast_to_vertex(thn)
+    els = __cast_to_vertex(els)
```
christophernorth: Do you need this?  Won't just calling the constructor for the If Vertex make this happen anyway?

keanu-python/keanu/vertex/iff.py
```diff
@@ -0,0 +1,39 @@
...
+
+    if type(predicate) != Boolean:
+        raise TypeError("Predicate must be boolean: got {}".format(type(predicate)))
+
+    then_type = get_type_of_value(thn.get_value())
+    else_type = get_type_of_value(els.get_value())
+    if (then_type != else_type):
+        raise TypeError('The "then" and "else" clauses must be of the same datatype: {} vs {}'.format(
+            then_type, else_type))
```
christophernorth: I'm assuming you cast to a vertex first so you can unconditionally call the get_value() function and then pass that to infer_type.  We could just skip this step by having get_type_of_value handle vertices natively?

keanu-python/keanu/vertex/iff.py
```diff
@@ -0,0 +1,39 @@
...
+    predicate = __cast_to_vertex(predicate)
+    thn = __cast_to_vertex(thn)
+    els = __cast_to_vertex(els)
+
+    if type(predicate) != Boolean:
+        raise TypeError("Predicate must be boolean: got {}".format(type(predicate)))
+
+    then_type = get_type_of_value(thn.get_value())
+    else_type = get_type_of_value(els.get_value())
+    if (then_type != else_type):
```
christophernorth: I'm concerned that users could fall in to a trap using this and in the end it won't actually make the code more usable.

As an example of what I'm thinking, imagine the scenario:
a = Gaussian(0.0, 1.0)
pred = a >= 0
ifVertex = If(pred, a, 0)

At first glance, that looks like a perfectly valid thing to do, but would throw an exception here.  If the user instead did a DoubleIf explicitly then it would just work as they'd expect as we'd know the types to coerce to.

It's the same argument I have against the "Const" utility function - both of these functions don't understand the type the user *actually* wants to produce, and given they *have* to care about this with Keanu we're better off not hiding it from them IMO.

PR #456, opened 2019-01-23
NUTS stepsize adaption issue

gordoncaleb:
The stepsize adaption in the NUTS paper uses the concept of stepsize frozen and just stepsize. The frozen bit was dropped in a refactor a while ago.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/nuts/NUTS.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/nuts/NUTS.java
@@ -105,6 +105,7 @@ public class NUTS implements PosteriorSamplingAlgorithm {
         Map<VertexId, DoubleTensor> gradient = logProbGradientCalculator.getJointLogProbGradientWrtLatents();
 
         double initialLogOfMasterP = ProbabilityCalculator.calculateLogProbFor(probabilisticVertices);
+        Map<VertexId, Object> initialSample = takeSample(sampleFromVertices);
 
         double startingStepSize = (initialStepSize == null) ? Stepsize.findStartingStepSize(position,
             gradient,
@@ -121,9 +122,9 @@ public class NUTS implements PosteriorSamplingAlgorithm {
             adaptCount
         );
 
-        resetVertexValue(sampleFromVertices, position);
+        resetVertexValue(latentVertices, position);
 
-        Tree tree = Tree.createInitialTree(position, momentum, gradient, initialLogOfMasterP, takeSample(sampleFromVertices));
+        Tree tree = Tree.createInitialTree(position, momentum, gradient, initialLogOfMasterP, initialSample);
 
         return new NUTSSampler(
             sampleFromVertices,
--- a/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/nuts/Stepsize.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/nuts/Stepsize.java
@@ -31,6 +31,7 @@ class Stepsize implements SaveStatistics {
     private double stepsize;
     private double averageAcceptanceProb;
     private double logStepSizeFrozen;
+    private double logStepSize;
 
     /**
      * @param stepsize             the step size
@@ -41,7 +42,8 @@ class Stepsize implements SaveStatistics {
         this.targetAcceptanceProb = targetAcceptanceProb;
         this.stepsize = stepsize;
         this.averageAcceptanceProb = 0;
-        this.logStepSizeFrozen = Math.log(stepsize);
+        this.logStepSize = Math.log(stepsize);
+        this.logStepSizeFrozen = Math.log(1);
         this.adaptCount = adaptCount;
         this.shrinkageTarget = Math.log(10 * stepsize);
     }
@@ -105,10 +107,10 @@ class Stepsize implements SaveStatistics {
      */
     public double adaptStepSize(Tree tree, int sampleNum) {
 
-        double logStepSize = logStepSizeFrozen;
-
         if (sampleNum < adaptCount) {
             logStepSize = updateLogStepSize(tree, sampleNum);
+        } else {
+            logStepSize = logStepSizeFrozen;
         }
 
         stepsize = Math.exp(logStepSize);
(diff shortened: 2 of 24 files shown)


keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/nuts/NUTSTest.java
```diff
@@ -98,13 +94,12 @@ public void samplesGaussian() {
     @Test
     public void samplesContinuousPrior() {
```
christophernorth: It's not *immediately* obvious what this test is actually verifying (or why what it's verifying is correct).  A bit of exposition might be nice to explain why we think the test is valid?

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/nuts/NUTS.java
```diff
@@ -121,9 +122,9 @@ private NUTSSampler setupSampler(final BayesianNetwork bayesNet,
             adaptCount
         );
 
-        resetVertexValue(sampleFromVertices, position);
+        resetVertexValue(latentVertices, position);
```
christophernorth: This looks like a good fix!

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/nuts/Stepsize.java
```diff
@@ -146,10 +149,11 @@ private double updateLogStepSize(Tree tree, int sampleNum) {
         //(1-m^-k) * log(epsilon_bar_m-1)
         double increasedStepSizeFrozen = (1 - tendToZero) * logStepSizeFrozen;
 
+        //log(epsilon_bar_m) = m^-k * log(epsilon_m) + (1 - m^-k) * log(epsilon_bar_m-1)
+        logStepSizeFrozen = reducedStepSize + increasedStepSizeFrozen;
         averageAcceptanceProb = updatedAverageAcceptanceProb;
 
-        //log(epsilon_bar_m) = m^-k * log(epsilon_m) + (1 - m^-k) * log(epsilon_bar_m-1)
-        return reducedStepSize + increasedStepSizeFrozen;
+        return updatedLogStepSize;
```
christophernorth: These all look like reasonable changes - but I'll read through the paper again and compare to be absolutely sure.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/nuts/Stepsize.java
```diff
@@ -40,7 +41,9 @@
     Stepsize(double stepsize, double targetAcceptanceProb, int adaptCount) {
         this.targetAcceptanceProb = targetAcceptanceProb;
         this.stepsize = stepsize;
-        this.logStepSizeFrozen = Math.log(stepsize);
+        this.averageAcceptanceProb = 0;
```
GeorgeNash: this is auto initialised to zero

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/nuts/Stepsize.java
```diff
@@ -104,10 +107,10 @@ public static double findStartingStepSize(Map<VertexId, DoubleTensor> position,
      */
     public double adaptStepSize(Tree tree, int sampleNum) {
 
-        double logStepSize = logStepSizeFrozen;
-
         if (sampleNum < adaptCount) {
             logStepSize = updateLogStepSize(tree, sampleNum);
+        } else {
```
GeorgeNash: Good spot. I thought we were meant to return the last calculated value of `logStepSize` here

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/nuts/Stepsize.java
gordoncaleb: True but for clarity's sake.

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/nuts/Stepsize.java
gordoncaleb: Yes, please. The paper is a beast if you've never looked at it. Let me know if you'd like a summary before you dig in.

christophernorth: I've read it once before and thought I'd understood it, but I'll grab you if I'm struggling!  I just want to walk through side by side with the code to make sure I agree it's all looking good.

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/nuts/NUTSTest.java
```diff
@@ -60,7 +56,7 @@ public void canRecordStatisticsFromSamples() {
 
         Statistics statistics = nuts.getStatistics();
 
-        List<Double> stepSize = statistics.get(NUTS.Metrics.STEPSIZE);
```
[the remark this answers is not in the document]
gordoncaleb: Didn't mean to change this. Happy to revert.

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/nuts/NUTSTest.java
gordoncaleb: For the gaussian and the sum of gaussian tests, it's asserting that the average sample is equal to the MAP. This is true because we know the posterior distribution is symmetric about the MAP. We know it's symmetric because we can analytically determine it due to it just being the same as the Prior (no observations), which is either a multivariate gaussian with a diagonal covariance matrix or just a simple gaussian. I'll drop this explanation in the test.

christophernorth: Yep - agreed.  Thanks!

keanu-project/src/test/java/io/improbable/keanu/algorithms/mcmc/nuts/NUTSTest.java
```diff
@@ -98,13 +94,12 @@ public void samplesGaussian() {
     @Test
     public void samplesContinuousPrior() {
 
-        BayesianNetwork bayesNet = MCMCTestDistributions.createSumOfGaussianDistribution(20.0, 1.0, 46., 18.0);
+        BayesianNetwork bayesNet = MCMCTestDistributions.createSumOfGaussianDistribution(20.0, 1.0, 46., 15.0);
```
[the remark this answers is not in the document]
gordoncaleb: I changed it to 15 in order to make it *more* challenging to the sampler. Starting at 15 gives us a pretty bad starting point, which requires more samples but it's a better test for robustness.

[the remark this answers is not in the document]
christophernorth: Awesome - magic numbers are bad though I guess still - a few quick lines of description would probably still be good to explain why we've chosen them?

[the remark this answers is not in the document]
gordoncaleb: It *should* work for any number. In this case 15 is just a more difficult starting point. Can you look at the sum of Gaussian distribution code and let me know what you would like clarified?

PR #458, opened 2019-01-23
Improved Python overloading

wesleyw72:
Improves the Python overloading support. This should allow users to use the operator overloads on Integer vertices and get the correct vertex created.

Also performs casting where possible to mimic the behaviour of numpy. e.g `Integer a * Double b = Double(a) * b = Double`


keanu-python/keanu/vertex/base.py
```diff
@@ -202,6 +211,94 @@ class Integer(Vertex):
...
+            except KeyError:
+                return super().__array_ufunc__(ufunc, method, input0, input1)
+        else:
+            raise NotImplementedError("NumPy ufunc method %s not implemented" % method)
+
+    def __add__(self, other: vertex_operation_param_types) -> 'Vertex':
+        if is_floating_type(other):
+            return kn.vertex.generated.Addition(kn.vertex.generated.CastToDouble(self), other)
+
+        return kn.vertex.generated.IntegerAddition(self, other)
```
migwellian: nit: I would prefer `else return`

(applies to lots of places in this file)

keanu-python/keanu/vertex/base.py
```diff
@@ -202,6 +211,94 @@ class Integer(Vertex):
...
+        if is_floating_type(other):
+            return kn.vertex.generated.Power(kn.vertex.generated.CastToDouble(self), other)
+
+        return kn.vertex.generated.IntegerPower(self, other)
+
+    def __rpow__(self, other: vertex_operation_param_types) -> 'Vertex':
+        if is_floating_type(other):
+            return kn.vertex.generated.Power(other, kn.vertex.generated.CastToDouble(self))
+
+        return kn.vertex.generated.IntegerPower(other, self)
```
migwellian: can you refactor out the common logic? Something like:

    return self.construct(other, kn.vertex.generated.Power, n.vertex.generated.IntegerPower, right_op=true)

keanu-python/keanu/vertex/base.py
```diff
@@ -202,6 +211,94 @@ class Integer(Vertex):
...
+
+    def __truediv__(self, other: vertex_operation_param_types) -> 'Vertex':
+        other = cast_to_double_vertex_if_integer_vertex(other)
+        return kn.vertex.generated.Division(kn.vertex.generated.CastToDouble(self), other)
```
migwellian: I would find this logic easier to understand:

    return kn.vertex.generated.CastToDouble(self).__truediv__(other)

similarly for `rtruediv`.

keanu-python/keanu/vertex/base.py
```diff
@@ -94,40 +94,49 @@ def __array_ufunc__(self, ufunc: np.ufunc, method: str, input0: numpy_types, inp
             raise NotImplementedError("NumPy ufunc method %s not implemented" % method)
 
     def __add__(self, other: vertex_operation_param_types) -> 'Vertex':
+        other = cast_to_double_vertex_if_integer_vertex(other)
```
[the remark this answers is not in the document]
wesleyw72: For those cases, when the `other` is an `Integer` vertex then the left side operation (`add`, `sub`, etc) will be called on the `Integer` vertex therefore the logic isn't needed in the right sided operations on these.

keanu-python/keanu/vertex/base.py
wesleyw72: Yeah that's nice. I can do that in pretty much all of the integer ones so have done that.

keanu-python/tests/test_vertex_operations.py
```diff
@@ -135,126 +138,168 @@ def test_can_do_less_than_or_equal_to_with_vertex_on_rhs(lhs: Union[Vertex, nump
...
 
 
 # yapf: disable
+@pytest.mark.parametrize("lhs_constructor", types_to_test)
+@pytest.mark.parametrize("rhs_constructor", types_to_test)
 @pytest.mark.parametrize("lhs, rhs, expected_result", [
-    (Const(np.array([3., 2.])), Const(np.array([5., 7.])), np.array([15, 14])),
-    (Const(np.array([3., 2.])),       np.array([5., 7.]) , np.array([15, 14])),
-    (Const(np.array([3., 2.])),                 5.       , np.array([15, 10])),
+    (         np.array([10, 20]), np.array([1, 2]) , np.array([11, 22])),
```
GeorgeNash: Shouldn't we be creating different combinations of lhs and rhs with `integer` and `double`? Looks like we're always doing integer here

GeorgeNash: Oh that's done in:

```
@pytest.mark.parametrize("lhs_constructor", types_to_test)
@pytest.mark.parametrize("rhs_constructor", types_to_test)
```

keanu-python/tests/test_vertex_operations.py
```diff
@@ -135,126 +138,168 @@ def test_can_do_less_than_or_equal_to_with_vertex_on_rhs(lhs: Union[Vertex, nump
...
+                            rhs: arithmetic_op_value_types, expected_result: arithmetic_op_value_types) -> None:
+    lhs_casted = lhs_constructor(lhs)
+    rhs_casted = rhs_constructor(rhs)
+
+    if result_should_be_vertex(lhs_casted, rhs_casted):
+        result: arithmetic_op_value_types = lhs_casted - rhs_casted
+        assert_is_correct_vertex_type_and_expected_value(lhs_casted, rhs_casted, result, expected_result)
+
+        result = rhs_casted - lhs_casted
+        assert_is_correct_vertex_type_and_expected_value(lhs_casted, rhs_casted, result, -expected_result)
```
GeorgeNash: why the minus sign on `-expected_result`?

keanu-python/tests/test_vertex_operations.py
```diff
@@ -135,126 +138,168 @@ def test_can_do_less_than_or_equal_to_with_vertex_on_rhs(lhs: Union[Vertex, nump
 
 ### Arithmetic
 
+# Included the identity lambda to test the interoperability between vertexes and np/primitive types
+types_to_test = [generated.ConstantDouble, generated.ConstantInteger, lambda x: x]
```
GeorgeNash: nit: `vertex_construcors_to_test`

keanu-python/tests/test_vertex_operations.py
```diff
@@ -137,123 +139,176 @@ def test_can_do_less_than_or_equal_to_with_vertex_on_rhs(lhs: Union[Vertex, nump
...
+    (  Const(np.array([10, 20])),                 Const(2.), np.array([12, 22]), Double),
+    (  Const(np.array([10, 20])),                        2., np.array([12, 22]), Double),
+    (         np.array([10, 20]),                  Const(2), np.array([12, 22]), Integer),
+    (       np.array([10., 20.]),                  Const(2), np.array([12, 22]), Double),
+    (                        20.,                  Const(2),                 22, Double),
 ])
 # yapf: enable
-def test_can_do_addition(lhs: Union[Vertex], rhs: Union[Vertex, numpy_types, float],
-                         expected_result: numpy_types) -> None:
+def test_can_do_addition(lhs: Union[Vertex], rhs: Union[Vertex, numpy_types, float], expected_result: numpy_types,
```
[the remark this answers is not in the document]
wesleyw72: Updated the types.

keanu-python/tests/test_vertex_operations.py
wesleyw72: Hmm, but the lambda wouldn't construct a vertex?

keanu-python/tests/test_vertex_operations.py
wesleyw72: This is the rhs - lhs which is testing the other way round so the expected answer is negated.

keanu-python/keanu/vertex/base.py
```diff
@@ -202,6 +212,91 @@ class Integer(Vertex):
...
+            return kn.vertex.generated.IntegerPower(self, other)
+
+    def __rpow__(self, other: vertex_operation_param_types) -> 'Vertex':
+        if is_floating_type(other):
+            return kn.vertex.generated.CastToDouble(self).__rpow__(other)
+        else:
+            return kn.vertex.generated.IntegerPower(other, self)
+
+    def __truediv__(self, other: vertex_operation_param_types) -> 'Vertex':
+        return kn.vertex.generated.CastToDouble(self).__truediv__(other)
```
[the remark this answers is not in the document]
wesleyw72: true div should always be floating, so don't need to do this check.

keanu-python/keanu/vertex/base.py
```diff
@@ -155,7 +164,8 @@ def __abs__(self) -> 'Vertex':
     def __round__(self, ndigits: Optional[int] = 0) -> 'Vertex':
         if ndigits != 0:
             raise NotImplementedError("Keanu only supports rounding to 0 digits")
-        return kn.vertex.generated.Round(self)
+        else:
+            return kn.vertex.generated.Round(self)
```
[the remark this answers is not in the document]
wesleyw72: Ah good spot, that was some poor copy and pasting. Will remove

keanu-python/keanu/vertex/base.py
wesleyw72: Good idea, have done this into `__op_based_on_other_type`

keanu-python/tests/test_vertex_operations.py
GeorgeNash: Ok - no worries

keanu-python/keanu/vertex/base.py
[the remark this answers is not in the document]
migwellian: got it 👍

keanu-python/keanu/vertex/base.py
```diff
@@ -202,6 +211,87 @@ class Integer(Vertex):
...
+            other, lambda casted_to_double: casted_to_double.__mul__(other),
+            lambda other_holder: kn.vertex.generated.IntegerMultiplication(self, other_holder))
+
+    def __rmul__(self, other: vertex_operation_param_types) -> 'Vertex':
+        return self.__op_based_on_other_type(
+            other, lambda casted_to_double: casted_to_double.__rmul__(other),
+            lambda other_holder: kn.vertex.generated.IntegerMultiplication(other_holder, self))
+
+    def __pow__(self, other: vertex_operation_param_types) -> 'Vertex':
+        return self.__op_based_on_other_type(other, lambda casted_to_double: casted_to_double.__pow__(other),
```
migwellian: nit: format the same as the others (CR before `other`)

keanu-python/keanu/vertex/base.py
```diff
@@ -202,6 +211,87 @@ class Integer(Vertex):
...
+                                             lambda other_holder: kn.vertex.generated.IntegerPower(self, other_holder))
+
+    def __rpow__(self, other: vertex_operation_param_types) -> 'Vertex':
+        return self.__op_based_on_other_type(other, lambda casted_to_double: casted_to_double.__rpow__(other),
```
migwellian: nit: format the same as the others (CR before `other`)

keanu-python/tests/test_vertex_operations.py
migwellian: nice

keanu-python/keanu/vertex/base.py
wesleyw72: That fails the `yapf` formatting checks

PR #459, opened 2019-01-25
Reverse the ordering of parameters in VertexLabel constructor

icantsayrural:
Before, for `io.improbable.keanu`, a user had to pass `VertexLabel("keanu", "improbable", "io")`. Now they can pass `VertexLabel("io", "improbable", "keanu")`.


keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
```diff
@@ -37,18 +46,18 @@ public boolean isInNamespace(String... namespace) {
...
-        List<String> reducedNamespace = namespace.subList(0, namespace.size() - 1);
-        return new VertexLabel(this.name, reducedNamespace);
+        List<String> reducedNamespace = namespace.subList(1, namespace.size());
+        return new VertexLabel(reducedNamespace, this.name);
     }
 
     public Optional<String> getOuterNamespace() {
         try {
-            return Optional.of(namespace.get(namespace.size() - 1));
+            return Optional.of(namespace.get(0));
```
wesleyw72: You can use `return namespace.stream().findFirst();` instead of the try catch block.

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
```diff
@@ -2,24 +2,33 @@
...
     private static final char NAMESPACE_SEPARATOR = '.';
-    private final String name;
     private final List<String> namespace;
+    private final String name;
 
-    public VertexLabel(String name, String... namespace) {
-        this(name, ImmutableList.copyOf(namespace));
+    public VertexLabel(String outer, String... inner) {
+        if (inner.length <= 0) {
+            this.namespace = ImmutableList.of();
```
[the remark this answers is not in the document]
icantsayrural: Hmm.. I kind of like `IndexOutOfBoundsException` thrown instead of `IllegalArgumentException` thrown from non-guava list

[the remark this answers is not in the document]
icantsayrural: (from my previous comment: I was specifically talking about a case like `VertexLabel#itThrowsIfYouDiminishTheNamespaceButThereIsNone` test)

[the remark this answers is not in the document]
wesleyw72: Fair enough, happy for you to merge this.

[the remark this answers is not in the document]
icantsayrural: But until we upgrade to guava 21, I agree I should use `Collections.emptyList`!

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
```diff
@@ -2,24 +2,34 @@
...
+        if (inner.length <= 0) {
+            this.namespace = Collections.emptyList();
+            this.name = outer;
+        } else {
+            this.namespace = ImmutableList.<String>builder()
+                .add(outer)
+                .add(Arrays.copyOf(inner, inner.length - 1))
+                .build();
+            this.name = inner[inner.length - 1];
+        }
```
migwellian: I think this logic is easier to understand:

    ImmutableList<String> allTerms = ImmutableList.<String>builder().add(outer).add(inner).build();
    int lastPosition = allTerms.size() - 1;
    this.name = allTerms.get(lastPosition);
    this.namespace = allTerms.subList(0, lastPosition);

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
```diff
@@ -37,30 +47,26 @@ public boolean isInNamespace(String... namespace) {
...
+        return new VertexLabel(reducedNamespace, this.name);
     }
 
     public Optional<String> getOuterNamespace() {
-        try {
-            return Optional.of(namespace.get(namespace.size() - 1));
-        } catch (IndexOutOfBoundsException e) {
-            return Optional.empty();
-        }
+        return namespace.stream().findFirst();
```
migwellian: nice

keanu-project/src/test/java/io/improbable/keanu/vertices/VertexLabelTest.java
```diff
@@ -132,44 +132,50 @@ public void youCanDiminishTheNamespace() {
...
         String name = "foo";
-        VertexLabel foo = new VertexLabel(name, innerNamespace, outerNamespace);
+        VertexLabel foo = new VertexLabel(outerNamespace, innerNamespace, name);
         VertexLabel newFoo = foo.withoutOuterNamespace();
-        assertThat(newFoo, equalTo(new VertexLabel(name, innerNamespace)));
+        assertThat(newFoo, equalTo(new VertexLabel(innerNamespace, name)));
     }
 
-    @Test(expected = IndexOutOfBoundsException.class)
+    @Test(expected = IllegalArgumentException.class)
```
migwellian: NB you still get an IndexOutOfBoundsException if you do:

        VertexLabel foo = new VertexLabel("foo", "bar").withoutOuterNamespace();
        foo.withoutOuterNamespace();

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
icantsayrural: (As pointed out by Wes) I like the use of `Collections.emptyList()` because it just reuses a singleton immutable empty list for `VertexLabel` when there is no namespace. This is good because in most cases users won't specify a namespace.

I think using `subList` to set `namespace` might be a small memory leak because it's just a view of `allTerms` and will still hold onto the last element?

keanu-project/src/test/java/io/improbable/keanu/vertices/VertexLabelTest.java
icantsayrural: Yeah... the returned list from `withoutOuterNamespace` will be guava so it'll throw `IndexOutOfBoundsException` instead - but I think that's OK since the error message will be self-explanatory. I'll add a test for it though.

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexLabel.java
migwellian: That's interesting about the view, although I don't think it's a memory leak, because `allTerms` would get garbage collected when `this.namespace` gets garbage collected, right?

Anyway if you prefer to use `Collections.emptyList()` that's fine.

Another way you could make the logic clearer is to get rid of the `if` by splitting your `then` and `else` into 2 different constructors:

    public VertexLabel(String name) {
        this.namespace = Collections.emptyList();
        this.name = name;
    }

    public VertexLabel(String outer, String... inner) {
        this.namespace = ImmutableList.<String>builder()
            .add(outer)
            .add(Arrays.copyOf(inner, inner.length - 1))
            .build();
        this.name = inner[inner.length - 1];
    }

keanu-project/src/test/java/io/improbable/keanu/vertices/VertexLabelTest.java
migwellian: 👍

PR #463, opened 2019-01-28
Add __version__ attribute 

wesleyw72:
Allow users to query the version of keanu they are using by `keanu.__version__`. Also changes `setup.py` to use the same version string for a single location of storing the version.


keanu-python/setup.py
```diff
@@ -1,23 +1,27 @@
...
 import shutil
+import os
 
 
 def readme() -> str:
     with open('README.rst', "r") as f:
         return f.read()
 
 
-version_string = '0.0.15.dev1'
```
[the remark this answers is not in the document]
wesleyw72: It was set as part of the release process prior to doing a release. I think from now on though we can keep it up to date with the version we are working towards like the rest of the version strings.

release_notes.txt
```diff
@@ -1,4 +1,5 @@
 ### Version 0.0.19 ###
 
-Added a permute vertex
-Added the release notes text file to the repo
\ No newline at end of file
+* Added a permute vertex
+* Added the release notes text file to the repo
+* Added __version__ attribute to Python
```
icantsayrural: I think the newest change goes to top (assuming adding the release notes text file was the first addition). Small nit though.

wesleyw72:
@migwellian Good idea, will add that test.

I think that one lining the exec statement will make it a bit of a complex line because mine has a join etc. I did have a look at how numerous other repos did versioning, and they all seem to do it in a slightly different way and as you said slightly overkill at the moment. We can look at how we do versioning of the Python stuff better in the future (taking it from the tag).

release_notes.txt
wesleyw72: Good point, I will swap it around

PR #466, opened 2019-01-29
Refactor dot file generation

chris-major-improbable:
This removes the `DotSaver` class - and replaces it with the concept of a `AbstractGraph` (which represents _any_ directed graph structure) an implementation called `VertexGraph` (which is created from keanu vertices) and a `GraphToDot` class which writes any `AbstractGraph` to a `.dot` file.

This means we can reuse the Dot file syntax for other graph types - as well as modify and customise the output. The `VertexGraph` provides some built in methods to customise the output (e.g. coloring nodes by the data type or reducing the complexity by hiding deterministic nodes).


keanu-project/src/main/java/io/improbable/keanu/util/graph/AbstractGraph.java
```diff
@@ -0,0 +1,84 @@
...
+    public Collection<E> getEdges() {
+        return edges;
+    }
+
+    /**
+     * This converts a java color to a .dot format
+     * @param c The color to convert
+     * @return the colot as a 6 digit hex string (e.g. #FF00FF)
+     */
+    public static final String formatColorForDot(Color c) {
```
gordoncaleb: Does this DOT related method have to live on this very general AbstractGraph class?

chris-major-improbable: I think it's misnamed - it's actually formatting a color for HTML and DOT - and I think it makes sense to have a standard way to store colors as strings in the AbstractGraph.
I'll rename it.

PR #472, opened 2019-01-31
Feature/remove sample function

GeorgeNash:
- `sample` taken off `Vertex`
- `Probabilistic` extends `Samplable`
- Can no longer call `sample` on operations


keanu-project/src/main/java/io/improbable/keanu/algorithms/particlefiltering/ParticleFilter.java
```diff
@@ -132,7 +134,7 @@ private void addObservedVertexToParticles(List<Particle> particles,
     }
 
     private <T> void sampleValueAndAddToParticle(Vertex<T> vertex, Particle particle) {
-        T sample = vertex.sample(random);
+        T sample = ((Probabilistic<T>) vertex).sample(random);
```
GeorgeNash: This looks unsafe but the `vertex` is coming from a set inside `sort` which extracts only latent variables

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
```diff
@@ -209,13 +210,16 @@ public static void setFromSampleAndCascade(List<? extends Vertex> vertices) {
 
     public static void setFromSampleAndCascade(List<? extends Vertex> vertices, KeanuRandom random) {
```
gordoncaleb: Can we just take a List<? extends Probabilistic>?

keanu-project/src/main/java/io/improbable/keanu/vertices/LogProbGraph.java
```diff
@@ -47,10 +46,6 @@ public DoubleTensor calculate() {
             return getPlaceholderVertexValue(this);
         }
 
-        @Override
-        public DoubleTensor sample(KeanuRandom random) {
```
gordoncaleb: Nice! RIP `sample()`

keanu-project/src/main/java/io/improbable/keanu/network/BayesianNetwork.java
gordoncaleb: Talked about this offline. OK...

PR #473, opened 2019-02-01
Feature/mir save load

christophernorth:
Diffs to introduce MIR based save load to Keanu.  At the moment I've left the existing Protobuf stuff in there, but this can be removed at some point in the future if we want to.

Couple of small tasks left to do, but I thought I'd put this up ASAP so people can take a look at the real meat.

Reminder that this is the very basic stuff to just get us using Save/Load.  We'll need to do more work as we add more features to detect more invalid input conditions/open up some of the restrictions currently placed (like only reading graphs that we generated inside keanu).


keanu-project/src/main/java/io/improbable/keanu/util/io/MIRLoader.java
```diff
@@ -0,0 +1,69 @@
...
+
+    private static void checkProperties(MIR.Model model) {
+        MIR.ModelProperties properties = model.getProperties();
+
+        if (properties.getMirVersion() != MIR.VersionNumber.VERSION_1) {
+            throw new IllegalArgumentException("Keanu only supports Version 1 of MIR");
+        }
+
+        if (properties.getRecursiveTracesCount() != 0) {
+            throw new IllegalArgumentException("Keanu Does not support Recursive models");
```
joncoello83: Capital D in 'Does' should be lower case. Possible capital R in 'Recursive' too. Maybe saying 'Keanu does not support models with recursive function calls' would be slightly clearer.

keanu-project/src/main/java/io/improbable/keanu/util/io/MIRLoader.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        if (properties.getLoopMetadata().getDimensionGenerating()
+            || properties.getRecursionMetadata().getDimensionGenerating()) {
+            throw new IllegalArgumentException("Keanu does not support Dimension Generation");
```
joncoello83: I think this exception could be clearer. Maybe 'Keanu does not support models with random number generation in loops' or something like that?

keanu-project/src/main/java/io/improbable/keanu/util/io/MIRLoader.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        if (properties.getLoopMetadata().getIteration() != MIR.IterationType.NONE
+            || properties.getRecursionMetadata().getIteration() != MIR.IterationType.NONE) {
+            throw new IllegalArgumentException("Keanu does not support iteration");
```
joncoello83: Do you think iteration or looping is clearer? I think looping, or models with loops, but I'm not sure.

keanu-project/src/main/java/io/improbable/keanu/util/io/MIRSaver.java
```diff
@@ -0,0 +1,70 @@
...
+import io.improbable.mir.SavedBayesNet;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.HashMap;
+import java.util.Map;
+
+public class MIRSaver extends ProtobufSaver {
+
+    public final static String MODEL_NAME = "Saved Keanu Graph";
```
joncoello83: Do we want to give the user the option to provide their own model name?

keanu-project/src/main/java/io/improbable/keanu/util/io/MIRLoader.java
```diff
@@ -0,0 +1,69 @@
...
+
+    private static void checkFunction(MIR.Model model) {
+        if (model.getFunctionsByNameMap().get(MIRSaver.ENTRY_POINT_NAME) == null) {
+            throw new IllegalArgumentException("Expected Entry Point not found");
+        }
+    }
+
+    private static void checkEntryPoint(MIR.Model model) {
+        if (!model.getEntryPointName().equals(MIRSaver.ENTRY_POINT_NAME)) {
+            throw new IllegalArgumentException("Keanu only supports loading Keanu generated Graphs");
```
joncoello83: Do we need this rule? Provided that the graph being loaded is a BN, it shouldn't matter where the graph came from.

keanu-project/src/main/java/io/improbable/keanu/util/io/MIRLoader.java
```diff
@@ -0,0 +1,69 @@
...
+        MIR.Model model = MIR.Model.parseFrom(input);
+        checkModelIsCompatible(model);
+
+        return extractGraphAndLoadNetwork(model);
+    }
+
+    private static void checkModelIsCompatible(MIR.Model model) {
+        checkProperties(model);
+        checkEntryPoint(model);
+        checkFunction(model);
```
joncoello83: Could check that the model only contains one function? That is implicitly covered by the rule that it can only load Keanu models at the moment, but I'd expect the 'one function' rule to apply for longer than the 'only Keanu models' rule. The 'one function' rule may actually apply indefinitely if function in-lining is carried out by an intermediate tool.

joncoello83: We could also check that the one function only contains one graph too. That seems like the set of checks that we'll do to verify that a model can be loaded once we've build the intermediate Bayesnetifier.

keanu-project/src/main/java/io/improbable/keanu/util/io/MIRSaver.java
```diff
@@ -0,0 +1,70 @@
...
+    public final static String ENTRY_POINT_NAME = "Keanu Graph";
```
joncoello83: Could we call the entry point 'Bayesian Network' or something else that's not Keanu specific?

keanu-project/src/main/java/io/improbable/keanu/util/io/MIRSaver.java
christophernorth: added (+ setting of Metadata)

keanu-project/src/main/java/io/improbable/keanu/util/io/MIRLoader.java
christophernorth: We certainly could leave it configurable, but until we start adding features I like that we're being explicit that you should only really use it to save and load stuff keanu itself generated.

keanu-project/src/main/java/io/improbable/keanu/util/io/MIRSaver.java
christophernorth: As above - I'd prefer for this to be Keanu specific - it is a graph that Keanu has saved and loaded after all.

keanu-project/src/test/java/io/improbable/keanu/util/io/MIRTest.java
```diff
@@ -0,0 +1,53 @@
...
+        assertThat(output.size(), greaterThan(0));
+        ByteArrayInputStream input = new ByteArrayInputStream(output.toByteArray());
+
+        MIRLoader loader = new MIRLoader();
+        BayesianNetwork readNet = loader.loadNetwork(input);
+
+        assertThat(readNet.getLatentVertices().size(), is(1));
+        assertThat(readNet.getLatentVertices().get(0), instanceOf(GaussianVertex.class));
+        GaussianVertex latentGaussianVertex = (GaussianVertex) readNet.getLatentVertices().get(0);
+        GaussianVertex labelGaussianVerted = (GaussianVertex) readNet.getVertexByLabel(new VertexLabel(gaussianLabel));
```
joncoello83: typo in 'Verted'

keanu-project/src/test/java/io/improbable/keanu/util/io/MIRTest.java
```diff
@@ -0,0 +1,53 @@
...
+        assertThat(latentGaussianVertex, equalTo(labelGaussianVerted));
+        assertThat(latentGaussianVertex.getMu().getValue(0), closeTo(3.0, 1e-10));
+        assertThat(labelGaussianVerted.getMu().getValue(2), closeTo(5.0, 1e-10));
```
joncoello83: typo in 'Verted'

keanu-project/src/main/java/io/improbable/keanu/util/io/MIRLoader.java
christophernorth: I've added those checks now.

keanu-project/src/test/java/io/improbable/keanu/util/io/MIRTest.java
christophernorth: done

keanu-project/src/test/java/io/improbable/keanu/util/io/MIRTest.java
christophernorth: done

PR #475, opened 2019-02-01
Remove need to pass latent vertices to prior proposal distribution

gordoncaleb:
This PR removes the need for our PriorProposalDistribution to know anything about the model at construction time. It relies on only ever being used with Keanu vertices, which was already the case. I've added defaults for the MH and SimulatedAnnealing builders but those will probably go when we add something other than Keanu Vertex for computation.

This PR is a response to an issue faced where the user didn't know how to build a new ProposalDistribution and ended up passing it the wrong vertices at construction time.


migwellian:
What was the issue? What vertices did they pass at construction time.

release_notes.md
```diff
@@ -6,10 +6,6 @@
   * `Keanu.Sampling.SimulatedAnnealing`
   * `Keanu.Sampling.MCMC` (to automatically choose between `MetropolisHastings` and `NUTS`)
 * The `PosteriorSamplingAlgorithm` objects are now immutable: you cannot set their values after construction.
-* When you choose a custom configuration of `MetropolisHastings` or `SimulatedAnnealing`, you must specify:
```
GeorgeNash: Can we leave this in but change it slightly. As people can still customise the proposal and rejection strategies.

`you can specify (defaults are prepared)... `

and: ```the proposal distribution - the default option is PriorProposalDistribution. Options are `PriorProposalDistribution` and `GaussianProposalDistribution```

```a proposal rejection strategy - the default option is RollBackToCachedValuesOnRejection. Options are RollBackToCachedValuesOnRejection and RollbackAndCascadeOnRejection```

gordoncaleb: 👍

PR #479, opened 2019-02-05
Lower number of samples

GeorgeNash:
This runs for bleedin' ages - as it's not currently working as expected, let's lower the number of samples.

### DIFF
--- a/keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHierarchicalRegression.java
+++ b/keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHierarchicalRegression.java
@@ -136,7 +136,7 @@ public class RadonHierarchicalRegression {
             .maxTreeHeight(5)
             .adaptEnabled(true)
             .build()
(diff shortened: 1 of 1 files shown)


keanu-project/src/test/java/io/improbable/keanu/e2e/regression/RadonHierarchicalRegression.java
```diff
@@ -136,7 +136,7 @@ private NetworkSamples NUTSSample(BayesianNetwork bayesianNetwork) {
             .maxTreeHeight(5)
             .adaptEnabled(true)
             .build()
-            .getPosteriorSamples(probabilisticModel, Arrays.asList(muAlpha, muBeta, sigmaAlpha, sigmaBeta), 1000)
+            .getPosteriorSamples(probabilisticModel, Arrays.asList(muAlpha, muBeta, sigmaAlpha, sigmaBeta), 10)
```
alex-remedios-improbable: perhaps leave a handy comment here

PR #480, opened 2019-02-05
Use `match` parameter in calls to `pytest.raises`

migwellian:
We don't need to inspect `excinfo.value`


migwellian:
The `r` doesn't mean regex exactly, it means "raw". If you use `\` in your regex and don't use a raw string, you get a warning.

But to answer your question - there is a `message` parameter but it's deprecated in favour of `matches`, so we're stuck with that.

PR #484, opened 2019-02-05
Feature/unwrap java exceptions

migwellian:
Py4J doesn't do a great job of presenting the information about an exception that's emanated from Java: the `Py4JJavaError` contains a `java_exception` object but it's of type `JavaObject`.

This PR provides a utility function to turn a `Py4JJavaError` into a simple struct that has a `type` and a `message`.

Obviously there's more to Java's Exception API than that, but for most/all of our cases we only care about the type and message.


icantsayrural:
What is this for? For asserts in tests?

keanu-python/keanu/context.py
```diff
@@ -122,6 +123,18 @@ def to_java_vertex_array(self, l: Collection[Any]) -> JavaArray:
...
+
+        def __init__(self, type: str, message: str):
+            self.type = type
+            self.message = message
+
+        def __repr__(self) -> str:
+            return "{}: {}".format(self.type, self.message)
+
+    def from_java_exception(self, e: Py4JJavaError) -> 'KeanuContext.JavaException':
+        return KeanuContext.JavaException(e.java_exception.getClass().getName(), e.java_exception.getMessage())
```
icantsayrural: I wonder if `JavaException` should be the one unwrapping the Java exception class? (i.e. take raw Java exception in its constructor)

migwellian: That's a good idea. In fact I realised it would be much neater if I implemented it as a JavaObjectWrapper. Please take another look!

migwellian: I also made it implement `Exception` so you can catch and rethrow Py4J's exception. See the tests.

PR #486, opened 2019-02-06
implemented Java Iterators in Python

migwellian:
A little something that I need for Plates.

Python iterators don't have a concept of `hasNext()` so I needed to implement that.

### DIFF
--- a/keanu-python/keanu/functional/iterator.py
+++ b/keanu-python/keanu/functional/iterator.py
@@ -1,0 +1,35 @@
+from typing import TypeVar, Optional, Iterator
+
+T = TypeVar('T')
+
+class JavaIterator:
+    """
+    Wraps a python iterator in an object that implements Java's Iterator interface.
+    The main difference is that Python has no hasNext() method - it just raises a StopIteration exception when it's finished.
+    We implement hasNext() by using a field `next_item` which caches the next item to be returned.
+    """
+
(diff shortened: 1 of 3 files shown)


icantsayrural:
Can't we just do this?:
```
>>> a = iter("hi")
>>>print(next(a, None))
h
>>>print(next(a, None))
i
>>>print(next(a, None))
None
```

migwellian:
I'm not sure what you're asking for... `next` is, I'm guessing, just some syntactic sugar to catch the `StopIteration` exception and return what you told it to, i.e. `None`.

To illustrate the point:

	>>> a = iter("hi")
	>>> a.__next__()
	'h'
	>>> a.__next__()
	'i'
	>>> a.__next__()
	Traceback (most recent call last):
	  File "<stdin>", line 1, in <module>
	StopIteration

keanu-python/keanu/functional/iterator.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def __init__(self, python_iterator: Iterator[T]) -> None:
+        self.python_iterator = python_iterator
+        self.next_item: Optional[T] = None
+
+    def next(self) -> T:
+        if self.next_item is None:
+            return self.python_iterator.__next__()
+        else:
+            to_return = self.next_item
```
GeorgeNash: This is really well done - I had to write a test to make sure that `hasNext` wasn't iterating but it looks great

PR #487, opened 2019-02-07
[Python] Faster samples to ndarray

wesleyw72:
Users were experiencing a large amount of time between sampling finishing (progress bar at 100%) and the samples being returned. This was due to the poor performance of sending lots of array elements and then putting them into an ndarray. 

This PR changes the approach to send a byte array. This is much more efficient as it is sent together instead of individual elements. Numpy can also construct an ndarray straight from this array. 

In an example problem the time for the sample function to return decreased from 80 seconds to 5 seconds for 1000 samples.

See [here](https://stackoverflow.com/questions/39095994/fast-conversion-of-java-array-to-numpy-array-py4j) and [here](https://github.com/bartdag/py4j/issues/317) for more info.


keanu-project/src/main/java/io/improbable/keanu/util/Py4jUtils.java
```diff
@@ -0,0 +1,34 @@
...
+
+import lombok.experimental.UtilityClass;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+@UtilityClass
+public class Py4jUtils {
+    private static final int JAVA_DOUBLE_SIZE = 8;
+    private static final int JAVA_INT_SIZE = 4;
```
migwellian: you can use `Double.SIZE` and `Integer.SIZE` - just need to divide by 8 since they're measured in bits.

keanu-project/src/main/java/io/improbable/keanu/util/Py4jUtils.java
```diff
@@ -0,0 +1,43 @@
...
+        // Java defaults to BIG_ENDIAN. LITTLE_ENDIAN is what is expected by numpy to construct an ndarray.
+        integerBuffer.order(ByteOrder.LITTLE_ENDIAN);
+        for (int element : integerArray) {
+            integerBuffer.putInt(element);
+        }
+        byte[] byteArray = integerBuffer.array();
+        return byteArray;
+    }
+
+    public byte[] toByteArray(Boolean[] booleanArray) {
```
migwellian: nit - for consistency with the other 2 methods

```suggestion
    public byte[] toByteArray(boolean[] booleanArray) {
```

keanu-project/src/main/java/io/improbable/keanu/util/Py4jUtils.java
```diff
@@ -0,0 +1,43 @@
+package io.improbable.keanu.util;
+
+import lombok.experimental.UtilityClass;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+@UtilityClass
+public class Py4jUtils {
```
GeorgeNash: I don't like `Util` very much in the name of classes, they normally become a graveyard for all sorts of stuff. Can we give it a specific name like `Py4jByteConverter`

keanu-project/src/main/java/io/improbable/keanu/util/Py4jUtils.java
```diff
@@ -0,0 +1,43 @@
...
+        }
+        byte[] byteArray = integerBuffer.array();
+        return byteArray;
+    }
+
+    public byte[] toByteArray(boolean[] booleanArray) {
+        ByteBuffer booleanBuffer = ByteBuffer.allocate(booleanArray.length);
+        booleanBuffer.order(ByteOrder.LITTLE_ENDIAN);
+        for (boolean element : booleanArray) {
+            byte booleanByte = element ? (byte) 0x01 : 0x00;
```
GeorgeNash: does this do the cast on the return of the conditional? that's cool. I thought you'd have to do `(byte) 0x01 : (byte) 0x00`

release_notes.md
```diff
@@ -1,5 +1,8 @@
 ### Version 0.0.20 ###
 
+## Python
+* Improved performance of getting samples.
```
GeorgeNash: by using byte streams

keanu-project/src/main/java/io/improbable/keanu/util/Py4jUtils.java
wesleyw72: So did I, and then did the first one and it stopped complaining

PR #488, opened 2019-02-07
Add labels to constant vertices

wesleyw72:
When debugging a large model using the DOT format. Users would like to be able to label constant vertices. This is especially useful when identifying constants that share the same value in a model.


keanu-project/src/main/java/io/improbable/keanu/util/io/VertexDotLabel.java
```diff
@@ -32,23 +33,33 @@ public void setValue(String value) {
...
-            return dotLabel + DOT_LABEL_CLOSING;
+            StringBuilder dotLabel = new StringBuilder();
+            dotLabel.append(vertex.getId().hashCode() + DOT_LABEL_OPENING + value);
+            extraLabelInfoForVertexWithValue().ifPresent(info -> dotLabel.append(" (" + info + ")"));
+            return dotLabel.append(DOT_LABEL_CLOSING).toString();
         }
         return vertex.getId().hashCode() + DOT_LABEL_OPENING + getDescriptiveInfo() + DOT_LABEL_CLOSING;
     }
 
+    private Optional<String> extraLabelInfoForVertexWithValue() {
```
icantsayrural:
```suggestion
    private Optional<String> getDescriptiveInfoForVertexWithValue() {
```

keanu-project/src/main/java/io/improbable/keanu/util/io/VertexDotLabel.java
```diff
@@ -32,23 +33,33 @@ public void setValue(String value) {
     public String inDotFormat() {
         // Output value if value is set, but also add some descriptive info for non-constant vertices.
```
GeorgeNash: I guess this changes now to: `also add some descriptive info for all vertices`

keanu-project/src/test/resources/dotFiles/ConstantScalarIntNodeOutput.dot
```diff
@@ -1,7 +1,7 @@
 digraph BayesianNetwork {
 <91> -> <92> [label=right]
 <90> -> <92> [label=left]
-90[label="42"]
+90[label="42 (SomeLabel)"]
```
migwellian: Have you checked with users that this is the desired format? For example they might prefer `SomeLabel: 42` or `SomeLabel = 42`

keanu-project/src/main/java/io/improbable/keanu/util/io/VertexDotLabel.java
```diff
@@ -32,23 +33,33 @@ public void setValue(String value) {
...
+            return dotLabel.append(DOT_LABEL_CLOSING).toString();
         }
         return vertex.getId().hashCode() + DOT_LABEL_OPENING + getDescriptiveInfo() + DOT_LABEL_CLOSING;
     }
 
+    private Optional<String> getDescriptiveInfoForVertexWithValue() {
+        if (!vertexLabel.isEmpty()) {
+            return Optional.of(vertexLabel);
+        }
+        if (!(vertex instanceof ConstantVertex)) {
```
migwellian: I would prefer `else if` here - it doesn't change the functionality (because of the `return`s but it makes it clearer to the reader that the blocks are mutually exclusive.

(And I know the rest of this file doesn't adhere to this rule, but I would prefer it throughout)

keanu-project/src/main/java/io/improbable/keanu/util/io/VertexDotLabel.java
```diff
@@ -32,23 +33,33 @@ public void setValue(String value) {
...
+            return Optional.of(getAnnotationIfPresentElseSimpleName());
+        }
+        return Optional.empty();
```
migwellian: as per above, I'd prefer `else { return ... }`

keanu-project/src/main/java/io/improbable/keanu/util/io/VertexDotLabel.java
```diff
@@ -32,23 +33,33 @@ public void setValue(String value) {
...
     private String getDescriptiveInfo() {
         if (!vertexLabel.isEmpty()) {
             return vertexLabel;
         }
-        if (!annotation.isEmpty()) {
-            return annotation;
-        }
-        return vertex.getClass().getSimpleName();
+        return getAnnotationIfPresentElseSimpleName();
+    }
```
migwellian: We could combine `getDescriptiveInfo` and `getDescriptiveInfoForVertexWithValue` into one method like this. I think this is easier (although still not easy!) to understand overall since it puts all the branching into one place.

    private Optional<String> getDescriptiveInfo() {
        if (!vertexLabel.isEmpty()) {
            return Optional.of(vertexLabel);
        } else if (value.isEmpty() || !(vertex instanceof ConstantVertex)) {
            return Optional.of(getAnnotationIfPresentElseSimpleName());
        } else {
            return Optional.empty();
        }
    }

wesleyw72: Thank you! This is nicer

keanu-project/src/test/resources/dotFiles/ConstantScalarIntNodeOutput.dot
wesleyw72: Discussed offline, this isn't a change introduced in this PR.

PR #490, opened 2019-02-08
Feature/forward sampler

GeorgeNash:
- Sampler for sampling from the prior


keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/ForwardSampler.java
```diff
@@ -0,0 +1,114 @@
...
+     * Non probabilistic variables are computed second.
+     *
+     * @param network the network to sample from
+     * @param fromVariables the variables to sample from
+     * @param sampleCount the number of samples to take
+     * @return sampling samples of a computable graph
+     */
+    public NetworkSamples sample(BayesianNetwork network,
+                                        List<? extends Variable> fromVariables,
+                                        int sampleCount) {
```
christophernorth: Nit: align the start column for parameters

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/ForwardSampler.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class ForwardSampler {
```
christophernorth: Is there a reason this doesn't implement the PosteriorSamplingAlgorithm interface? (I know given the restrictions as they are currently the posterior == prior, but we're missing sample generation now (ie you can't stream samples) and if we relax the restrictions below the posterior could be different to the prior.

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/ForwardSampler.java
```diff
@@ -0,0 +1,114 @@
...
+    private void takeSamples(Map<VariableReference, List> samples, Set<? extends Variable> fromVariables) {
+        fromVariables.forEach(variable -> addVariableValue(variable, samples));
+    }
+
+    private void addVariableValue(Variable variable, Map<VariableReference, List> samples) {
+        List samplesForVariable = samples.computeIfAbsent(variable.getReference(), v -> new ArrayList<>());
+        samplesForVariable.add(variable.getValue());
+    }
+
+    private void upstreamOfProbabilisticDoesNotContainProbabilistic(Set<Vertex> vertices) {
```
christophernorth: This seems overly strict? (and certainly wouldn't work for the model that initially prompted ChrisM's implementation).

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/ForwardSampler.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    private NetworkSamples sample(ComputableModel graph,
+                                        Set<? extends Variable> probabilisticFromVariables,
+                                        Set<? extends Variable> nonProbabilisticFromVariables,
+                                        ProposalApplicationStrategy proposalApplicationStrategy,
+                                        int sampleCount,
+                                        KeanuRandom random) {
+
+        Set<? extends Variable> allSampleFromVariables = Sets.union(probabilisticFromVariables, nonProbabilisticFromVariables);
```
christophernorth: Do we need to generate this again?  Couldn't we just pass it through from the function above?

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/ForwardSampler.java
```diff
@@ -0,0 +1,114 @@
...
+        Map<VariableReference, List> samplesByVariable = new HashMap<>();
+
+        for (int sampleNum = 0; sampleNum < sampleCount; sampleNum++) {
+            Proposal probabilisticVariableProposal = PRIOR_PROPOSAL.getProposal(probabilisticFromVariables, random);
+            proposalApplicationStrategy.apply(probabilisticVariableProposal);
+            graph.compute(Collections.EMPTY_MAP, nonProbabilisticFromVariables.stream().map(Variable::getReference).collect(Collectors.toList()));
+            takeSamples(samplesByVariable, allSampleFromVariables);
+        }
+
+        ArrayList<Double> logProb = new ArrayList<>(Collections.nCopies(sampleCount, LOG_PROB_OF_PRIOR));
```
christophernorth: Is this correct?  Surely samples from the Prior still have a logProb for the likelihood?

keanu-python/keanu/algorithm/sampling.py
```diff
@@ -31,6 +32,12 @@ def get_sampler(self) -> JavaObject:
         return self._sampler
 
 
+class ForwardSampler(PosteriorSamplingAlgorithm):
```
christophernorth: This will cause issues if people try and use the sample generation stuff?

keanu-project/src/test/java/io/improbable/keanu/algorithms/sampling/ForwardSamplerTest.java
```diff
@@ -0,0 +1,165 @@
...
+    @Before
+    public void setup() {
+        random = new KeanuRandom(1);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void throwIfSampleFromHasUpstreamRandomVariables() {
+        GaussianVertex A = new GaussianVertex(100.0, 1);
+        DoubleVertex B = A.plus(ConstantVertex.of(5.));
+        GaussianVertex C = new GaussianVertex(B, 1.);
```
christophernorth: If we can't even handle a simple model like this, does the forward sampler really have that much value?

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/ForwardSampler.java
gordoncaleb: The `likelihood` is the probability of the observations. Do we have a use case where we need the logProb? I think this sampler violates our previous assumption that all samples from the posterior need a logProb. We only use this logProb of the samples for determining if the samples have converged on the posterior but in this case we know we're sampling directly from the posterior. We would need the logProb for VI though so maybe make it optional?

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/ForwardSampler.java
GeorgeNash: Relaxed the constraint now that we're happy with conditionals

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/ForwardSampler.java
GeorgeNash: Because it's tied to a Bayesian Network and we can't use a probabilistic model. I agree that the benefits of having it implement `PosteriorSamplingAlgorithm` are huge so I've made it implement it with the constraint you must pass Vertices. It gives us streaming for free!

keanu-project/src/test/java/io/improbable/keanu/algorithms/sampling/ForwardSamplerTest.java
GeorgeNash: As discussed offline - we now allow conditional

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/Forward.java
```diff
@@ -0,0 +1,110 @@
...
+        Set<Vertex> upstream = new HashSet<>();
+        for (Vertex vertex : fromVertices) {
+            LambdaSection upstreamLambdaSection = LambdaSection.getUpstreamLambdaSection(vertex, true);
+            Set<Vertex> vertexUpsteam = upstreamLambdaSection.getAllVertices();
+            upstream = Sets.union(upstream, vertexUpsteam);
+        }
+        return upstream;
+    }
+
+    private void assertUpstreamOfObservedDoesNotContainProbabilistic(List<Vertex> vertices) {
```
icantsayrural: nit:
```suggestion
    private void checkUpstreamOfObservedDoesNotContainProbabilistic(List<Vertex> vertices) {
```
"assert" is usually not used for checking errors in user input

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/ForwardSampler.java
```diff
@@ -0,0 +1,52 @@
...
+        this.topologicallySortedVertices = topologicallySortedVertices;
+        this.random = random;
+    }
+
+    @Override
+    public void step() {
+        for (Vertex vertex : topologicallySortedVertices) {
+            if (vertex instanceof Probabilistic) {
+                vertex.setValue(((Probabilistic) vertex).sample(random));
+            } else if (vertex instanceof NonProbabilistic){
```
icantsayrural:
```suggestion
            } else if (vertex instanceof NonProbabilistic) {
```

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/ForwardSampler.java
```diff
@@ -0,0 +1,52 @@
...
+import io.improbable.keanu.vertices.NonProbabilistic;
+import io.improbable.keanu.vertices.Probabilistic;
+import io.improbable.keanu.vertices.Vertex;
+
+import java.util.List;
+import java.util.Map;
+
+public class ForwardSampler implements SamplingAlgorithm {
+
+    private static final double LOG_PROB_OF_PRIOR = 0.;
```
icantsayrural: Why is this 0?

GeorgeNash: See conversation between Chris and Caleb above

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/ForwardSampler.java
christophernorth: Yeah - I believe this was initially added for KL divergence calculations and the like.  As you say given this sampling algorithm doesn't need to calculate the logProb as part of the operation it would be best to make this optional (and probably follow through with the other sampling algorithms too).  Perhaps something to raise a ticket for?

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/Forward.java
```diff
@@ -0,0 +1,110 @@
...
+    public NetworkSamplesGenerator generatePosteriorSamples(ProbabilisticModel model, List<? extends Variable> variablesToSampleFrom) {
+        return new NetworkSamplesGenerator(setupSampler(model, variablesToSampleFrom), StatusBar::new);
+    }
+
+    private SamplingAlgorithm setupSampler(ProbabilisticModel model, List<? extends Variable> variablesToSampleFrom) {
+        List<? extends Variable> latentVariables = model.getLatentVariables();
+        Preconditions.checkArgument(latentVariables.size() > 0, "Your model must contain latent variables in order to forward sample.");
+
+        Variable latent = latentVariables.get(0);
+        Preconditions.checkArgument(latent instanceof Vertex, "The Forward Sampler only works for Variables of type Vertex.");
```
christophernorth: Is it sufficient to check only the first variable? (ie if the first is a vertex are we guaranteed that all will be, and that this will always be true?).

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/Forward.java
```diff
@@ -0,0 +1,110 @@
...
+        for (Vertex vertex : fromVertices) {
+            LambdaSection upstreamLambdaSection = LambdaSection.getUpstreamLambdaSection(vertex, true);
+            Set<Vertex> vertexUpsteam = upstreamLambdaSection.getAllVertices();
+            upstream = Sets.union(upstream, vertexUpsteam);
+        }
+        return upstream;
+    }
+
+    private void checkUpstreamOfObservedDoesNotContainProbabilistic(List<Vertex> vertices) {
+        for (Vertex vertex : vertices) {
```
christophernorth: Probably not a huge pain at the moment, but obviously this could cause some performance issues in some large graphs with lots of observed variables.  If so we'd probably want to enhance how the Lambda section stuff works.

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/Forward.java
```diff
@@ -0,0 +1,110 @@
...
+                copy.remove(vertex);
+            }
+        }
+        return copy;
+    }
+
+    private Set<Vertex> allUpstreamVertices(List<Vertex> fromVertices) {
+        Set<Vertex> upstream = new HashSet<>();
+        for (Vertex vertex : fromVertices) {
+            LambdaSection upstreamLambdaSection = LambdaSection.getUpstreamLambdaSection(vertex, true);
```
christophernorth: This code is super clear - good job.  I'm a bit worried that this could cause pretty serious performance issues for larger graphs though (and yes - it only gets run once, but for interactive sampling that cost will still bite us).

I definitely think that this section provides a very strong incentive for enhancing the Lambda section code to improve performance for cases like this though.  Happy to help out with that if it would be useful.

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/Forward.java
```diff
@@ -0,0 +1,110 @@
...
+        sortedVertices = removeNonProbabilisticVerticesBeforeTheFirstProbabilistic(sortedVertices);
+
+        return new ForwardSampler(verticesToSampleFrom, sortedVertices, random);
+    }
+
+    private List<Vertex> removeNonProbabilisticVerticesBeforeTheFirstProbabilistic(List<Vertex> topologicalVertices) {
+        List<Vertex> copy = new ArrayList<>(topologicalVertices);
+
+        for (Vertex vertex : topologicalVertices) {
+            if (vertex instanceof Probabilistic) {
```
christophernorth: If something has been observed - we can treat it as non-probabilistic for the sake of this algorithm, right?

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/Forward.java
```diff
@@ -0,0 +1,110 @@
...
+                break;
+            } else {
+                copy.remove(vertex);
+            }
+        }
```
christophernorth: This could be pretty expensive again.  Couple of thoughts:
1) Removing elements from the start of an ArrayList is expensive as every item has to be copied over to the left - LinkedList or queue might be better if we want to handle this in this way.
2) The .remove(Object) version we're using here will be slower than just passing the index value directly.  Given we're just popping stuff off the front remove(0) would do here?

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/ForwardSampler.java
christophernorth: As discussed - would be good to raise a ticket to make tracking of sample log probs optional across the board and implement the true log prob calculation here if someone turns it on.

keanu-project/src/test/java/io/improbable/keanu/algorithms/sampling/ForwardSamplerTest.java
```diff
@@ -0,0 +1,233 @@
...
+        }
+
+        @Override
+        protected DoubleTensor op(DoubleTensor l, DoubleTensor r) {
+            ids.add(this.getId());
+            return DoubleTensor.ZERO_SCALAR;
+        }
+    }
+
+}
```
christophernorth: Really nice tests.

keanu-python/tests/test_sampling.py
```diff
@@ -229,16 +231,18 @@ def test_iter_returns_same_result_as_sample() -> None:
 
 
 def test_iter_with_live_plot(net: BayesNet) -> None:
-    _, ax = plt.subplots(3, 1, squeeze=False)
+    KeanuRandom.set_default_random_seed(1)
```
christophernorth: This line should no longer be required given it is run before every test now.

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/Forward.java
GeorgeNash: Fair point - will check them all

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/ForwardSampler.java
migwellian: it needs some explanation in the code, especially since below you do `logOfMasterPForEachSample.add(LOG_PROB_OF_PRIOR)`

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/ForwardSampler.java
```diff
@@ -0,0 +1,54 @@
...
+    public void sample(Map<VariableReference, List<?>> samples, List<Double> logOfMasterPForEachSample) {
+        step();
+        SamplingAlgorithm.takeSamples(samples, variablesToSampleFrom);
+        logOfMasterPForEachSample.add(LOG_PROB_OF_PRIOR);
+    }
+
+    @Override
+    public NetworkSample sample() {
+        step();
+        return new NetworkSample(SamplingAlgorithm.takeSample((List<? extends Variable<Object, ?>>) variablesToSampleFrom), LOG_PROB_OF_PRIOR);
```
migwellian: you don't need this cast

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/ForwardSampler.java
```diff
@@ -0,0 +1,54 @@
...
+        this.variablesToSampleFrom = variablesToSampleFrom;
+        this.topologicallySortedVertices = topologicallySortedVertices;
+        this.random = random;
+    }
+
+    @Override
+    public void step() {
+        for (Vertex vertex : topologicallySortedVertices) {
+            if (vertex instanceof Probabilistic) {
+                if (!(vertex.isObserved())) {
```
migwellian: you don't need this check, it's checked inside `setValue`

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/ForwardSampler.java
```diff
@@ -0,0 +1,54 @@
...
+                    vertex.setValue(((Probabilistic) vertex).sample(random));
+                }
+            } else if (vertex instanceof NonProbabilistic) {
+                vertex.setValue(((NonProbabilistic) vertex).calculate());
+            }
```
migwellian: else throw?

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/ForwardSampler.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Override
+    public void step() {
+        for (Vertex vertex : topologicallySortedVertices) {
+            if (vertex instanceof Probabilistic) {
+                if (!(vertex.isObserved())) {
+                    vertex.setValue(((Probabilistic) vertex).sample(random));
+                }
+            } else if (vertex instanceof NonProbabilistic) {
+                vertex.setValue(((NonProbabilistic) vertex).calculate());
```
migwellian: the `setValue` is the only reason this class needs to know about `Vertex` rather than just dealing with `Variable`s. Is this the best design choice? I can't think of a better way...

keanu-project/src/main/java/io/improbable/keanu/network/LambdaSection.java
```diff
@@ -121,4 +143,57 @@ public static LambdaSection getDownstreamLambdaSection(Vertex<?> aVertex, boolea
...
+    /**
+     * @param vertices       vertices to start propagation from
+     * @param nextVertices The next vertices to move to given a current vertex. E.g getChildren for downstream or
+     *                     getParents for upstream.
+     * @param shouldAdd    true when a given vertex should be included in the result, false otherwise
+     * @return A Set of vertices that are in the direction implied by nextVertices and filtered by shouldAdd
+     */
+    public static Set<Vertex> getVerticesCollection(List<Vertex> vertices,
+                                                    Function<Vertex, Collection<Vertex>> nextVertices,
+                                                    Predicate<Vertex> shouldAdd) {
```
migwellian: there's a lot of duplication between this and `getVertices` - I expect it's quite easy to roll them into one method

migwellian: ... and once you've done that, it should be easy to remove the duplication between  `getUpstreamLambdaSectionForCollection` and `getUpstreamLambdaSection` too

keanu-project/src/test/java/io/improbable/keanu/algorithms/sampling/ForwardSamplerTest.java
```diff
@@ -0,0 +1,233 @@
...
+            .average().getAsDouble();
+
+        assertEquals(0.0, averageA, 0.1);
+    }
+
+    public static class IDTrackerVertex extends DoubleVertex implements Differentiable, ProbabilisticDouble {
+
+        private final List<VertexId> ids;
+
+        public IDTrackerVertex(@LoadVertexParam("left")DoubleVertex left, @LoadVertexParam("right")DoubleVertex right, List<VertexId> ids) {
```
migwellian: it seems like over-engineering to support save/load in a test class!

keanu-project/src/test/java/io/improbable/keanu/algorithms/sampling/ForwardSamplerTest.java
```diff
@@ -0,0 +1,233 @@
...
+public class ForwardSamplerTest {
+
+    private KeanuRandom random;
+
+    @Before
+    public void setup() {
+        random = new KeanuRandom(1);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
```
migwellian: if you use `ExpectedException` you can assert something about the exception message

keanu-project/src/test/java/io/improbable/keanu/algorithms/sampling/ForwardSamplerTest.java
```diff
@@ -0,0 +1,233 @@
...
+    public void throwIfSampleFromHasObservationsDownstreamOfLatents() {
+        GaussianVertex A = new GaussianVertex(100.0, 1);
+        DoubleVertex B = A.plus(ConstantVertex.of(5.));
+        GaussianVertex C = new GaussianVertex(B, 1.);
+        C.observe(100.);
+
+        BayesianNetwork network = new BayesianNetwork(A.getConnectedGraph());
+        ProbabilisticModel model = new KeanuProbabilisticModel(network);
```
migwellian: you can do these 2 lines in one:

    ProbabilisticModel model = new KeanuProbabilisticModel(A.getConnectedGraph());

keanu-project/src/test/java/io/improbable/keanu/algorithms/sampling/ForwardSamplerTest.java
```diff
@@ -0,0 +1,233 @@
...
+        GaussianVertex A = new GaussianVertex(0, 1);
+        TestGraphGenerator.PassThroughVertex B = new TestGraphGenerator.PassThroughVertex(A, opCount, null, id -> log.info("OP on id:" + id));
+
+        BayesianNetwork network = new BayesianNetwork(A.getConnectedGraph());
+        ProbabilisticModel model = new KeanuProbabilisticModel(network);
+
+        Keanu.Sampling.Forward.withDefaultConfig(random).getPosteriorSamples(model, Arrays.asList(A, B), 100);
+
+        assertEquals(100, opCount.get());
+    }
```
migwellian: suggestion - How to do this with mocking instead of the `PassThroughVertex`:

    @Test
    public void nonProbabilisticVerticesAreRecomputedDuringForwardSample() {
        GaussianVertex A = new GaussianVertex(0, 1);
        Vertex B = mock(Vertex.class);
        when(B.getParents()).thenReturn(Collections.singleton(A));
        when(B.getId()).thenReturn(new VertexId());

        ProbabilisticModel model = new KeanuProbabilisticModel(A.getConnectedGraph());

        Keanu.Sampling.Forward.withDefaultConfig(random).getPosteriorSamples(model, Arrays.asList(A, B), 100);

        verify(B, times(100)).setValue(any());
    }

keanu-project/src/test/java/io/improbable/keanu/algorithms/sampling/ForwardSamplerTest.java
```diff
@@ -0,0 +1,233 @@
...
+
+    @Test
+    public void samplesInTopologicalOrder() {
```
migwellian: You can do all of this with mocks too - I think you can use `InOrder`: 

https://stackoverflow.com/a/21901415/741789

keanu-project/src/test/java/io/improbable/keanu/network/LambdaSectionTraversalTest.java
```diff
@@ -70,4 +71,24 @@ public void doesGetAllUpstreamVertices() {
...
         assertThat(verticesDepthFirst, containsInAnyOrder(A, B, aPlusB, cSigma, C));
     }
+
+    @Test
+    public void doesGetUpstreamProbabilisticVerticesOfCollection() {
+
+        LambdaSection lambdaSection = LambdaSection.getUpstreamLambdaSectionForCollection(Arrays.asList(C, cSigma), false);
+        Set<Vertex> verticesDepthFirst = lambdaSection.getLatentAndObservedVertices();
+
+        assertEquals(3, verticesDepthFirst.size());
```
migwellian: you don't need to check the size, as that's done implicitly in `containsInAnyOrder` on the next line

migwellian: ... and then I'd suggest you inline `verticesDepthFirst` too, and your test is only 2 lines:

    @Test
    public void doesGetUpstreamProbabilisticVerticesOfCollection() {
        LambdaSection lambdaSection = LambdaSection.getUpstreamLambdaSectionForCollection(Arrays.asList(C, cSigma), false);
        assertThat(lambdaSection.getLatentAndObservedVertices(), containsInAnyOrder(A, B, C));
    }

migwellian: ditto for every test in this file

keanu-project/src/test/java/io/improbable/keanu/algorithms/sampling/ForwardSamplerTest.java
GeorgeNash: Nice - done

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/Forward.java
```diff
@@ -0,0 +1,103 @@
...
+        List<Vertex> verticesToSampleFrom = variablesToSampleFrom.stream().map(v -> (Vertex) v).collect(Collectors.toList());
+        Set<Vertex> allUpstreamVertices = allUpstreamVertices(verticesToSampleFrom);
+
+        List<Vertex> sortedVertices = TopologicalSort.sort(allUpstreamVertices);
+        sortedVertices = removeNonProbabilisticVerticesBeforeTheFirstProbabilistic(sortedVertices);
+
+        return new ForwardSampler(verticesToSampleFrom, sortedVertices, random);
+    }
+
+    private List<Vertex> removeNonProbabilisticVerticesBeforeTheFirstProbabilistic(List<Vertex> topologicalVertices) {
```
gordoncaleb: Downstream of many instead with option to turn off.

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/Forward.java
gordoncaleb: This will return a LinkedList  as a List for later traversal btw. This could be a problem generally but especially if some other code starts randomly accessing elements.

keanu-project/src/main/java/io/improbable/keanu/network/LambdaSection.java
```diff
@@ -117,6 +160,7 @@ public static LambdaSection getDownstreamLambdaSection(Vertex<?> aVertex, boolea
                     queued.add(next);
                 }
             }
+            visited.add(visiting);
```
gordoncaleb: If you don't add the visiting until the end here you miss adding observed and probabilistic to the visited set.

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/Forward.java
christophernorth: +1

keanu-project/src/main/java/io/improbable/keanu/network/LambdaSection.java
```diff
@@ -64,45 +89,63 @@ public static LambdaSection getUpstreamLambdaSection(Vertex<?> aVertex, boolean
...
-        Deque<Vertex> stack = new ArrayDeque<>(initialNext);
         Set<Vertex> result = new HashSet<>();
-        result.add(vertex);
+        result.add(firstVertex);
+        result.addAll(verticesCopy);
 
         while (!stack.isEmpty()) {
             Vertex visiting = stack.pop();
 
+            if (visited.contains(visiting)) {
```
gordoncaleb: Talked about this offline. Also, should add another test to LambdaSectionEfficiencyTest.

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/Forward.java
```diff
@@ -0,0 +1,103 @@
...
+            }
+        }
+        return copy;
+    }
+
+    private Set<Vertex> allUpstreamVertices(List<Vertex> fromVertices) {
+        return LambdaSection.getUpstreamLambdaSectionForCollection(fromVertices, true).getAllVertices();
+    }
+
+    private void checkUpstreamOfObservedDoesNotContainProbabilistic(List<Vertex> vertices) {
```
gordoncaleb: Use upstream of collection?

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/Forward.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        Variable latent = latentVariables.get(0);
+        BayesianNetwork network = new BayesianNetwork(((Vertex) latent).getConnectedGraph());
+
+        List<Vertex> observedVertices = network.getObservedVertices();
+        checkUpstreamOfObservedDoesNotContainProbabilistic(observedVertices);
+
+        List<Vertex> verticesToSampleFrom = variablesToSampleFrom.stream().map(v -> (Vertex) v).collect(Collectors.toList());
+        Set<Vertex> allDownstreamVertices = allDownstreamVertices(network.getLatentVertices());
```
christophernorth: As discussed offline:
1) I don't think Lambda Sections always work here as we need to re-evaluate every node in the tree back to the root. (We'd be looking for a transitive closure kind of thing)
2) For the set of vertices we actually need to re-evaluate we want the intersection of (downstream of all latents) & (upstream from the nodes we're sampling from).

keanu-project/src/test/java/io/improbable/keanu/algorithms/sampling/ForwardSamplerTest.java
```diff
@@ -0,0 +1,268 @@
...
+import java.util.Set;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@Slf4j
+public class ForwardSamplerTest {
```
gordoncaleb: Could you add a test that samples from some sort of hierarchical gaussian model. Like A -> B where A and B are Gaussian and A is the mu of B?

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/ForwardSampler.java
GeorgeNash: We can't get rid of `setValue` for this use case unfortunately

keanu-project/src/test/java/io/improbable/keanu/algorithms/sampling/ForwardSamplerTest.java
GeorgeNash: I'd rather keep it explicit as a testing class as it becomes quite complex too mock

docs/legacy_docs/0_0_21/autocorrelation.md
```diff
@@ -0,0 +1,79 @@
+---
```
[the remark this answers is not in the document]
GeorgeNash: Nope - reverted

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/Forward.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    private Set<Vertex> allDownstreamVertices(List<Vertex> randomVertices) {
+        return LambdaSection.getDownstreamLambdaSectionForCollection(randomVertices, true).getAllVertices();
+    }
+
+    private void checkUpstreamOfObservedDoesNotContainProbabilistic(List<Vertex> observedVertices) {
+        LambdaSection upstreamLambdaSection = LambdaSection.getUpstreamLambdaSectionForCollection(observedVertices, false);
+        Set<Vertex> upstreamRandomVariables = upstreamLambdaSection.getAllVertices();
+        Preconditions.checkArgument(upstreamRandomVariables.size() == 1 || upstreamRandomVariables.size() == 0, "Forward sampler cannot be ran if observed variables have a random variable in their upstream lambda section");
```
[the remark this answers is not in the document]
GeorgeNash: Changed to illegal argument

keanu-project/src/test/java/io/improbable/keanu/algorithms/sampling/ForwardSamplerTest.java
GeorgeNash: I do that in `samplesFromRandomVariablesThatAreInUpstreamOfSampleFrom` and `samplesInTopologicalOrder`

keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/SamplingUtil.java
```diff
@@ -0,0 +1,22 @@
+package io.improbable.keanu.algorithms.mcmc;
+
+import io.improbable.keanu.algorithms.Variable;
+import io.improbable.keanu.algorithms.VariableReference;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class SamplingUtil {
```
migwellian: this can be a `@UtilityClass`

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/Forward.java
```diff
@@ -0,0 +1,131 @@
...
+        Set<Vertex> intersection = Sets.intersection(allDownstreamVertices, transitiveClosureSampleFrom);
+
+        List<Vertex> sortedVertices = TopologicalSort.sort(intersection);
+
+        return new ForwardSampler(network, verticesToSampleFrom, sortedVertices, random, calculateSampleProbability);
+    }
+
+    private BayesianNetwork checkSampleFromVariablesComeFromConnectedGraph(List<? extends Variable> variablesToSampleFrom) {
+        Variable variable = variablesToSampleFrom.get(0);
+        Set connectedGraph = ((Vertex) variable).getConnectedGraph();
```
migwellian: `Set<Vertex>`

keanu-project/src/main/java/io/improbable/keanu/algorithms/sampling/Forward.java
```diff
@@ -0,0 +1,131 @@
...
+        return new NetworkSamplesGenerator(setupSampler(model, variablesToSampleFrom), StatusBar::new);
+    }
+
+    private SamplingAlgorithm setupSampler(ProbabilisticModel model, List<? extends Variable> variablesToSampleFrom) {
+        List<? extends Variable> latentVariables = model.getLatentVariables();
+        Preconditions.checkArgument(latentVariables.size() > 0, "Your model must contain latent variables in order to forward sample.");
+
+        for (Variable variable : variablesToSampleFrom) {
+            Preconditions.checkArgument(variable instanceof Vertex, "The Forward Sampler only works for Variables of type Vertex. Received : " + variable);
+        }
```
migwellian: I think it would be cleaner to cast all your `Variable`s to `Vertex` here, just once, instead of multiple times further down.

keanu-project/src/main/java/io/improbable/keanu/network/Propagation.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Collections;
+import java.util.Deque;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+public class Propagation {
```
migwellian: `@UtilityClass`

PR #491, opened 2019-02-08
Feature/vertex label namespaces in python

migwellian:


keanu-python/keanu/base.py
```diff
@@ -13,9 +13,14 @@ def __repr__(self) -> str:
...
 
     def __getattr__(self, k: str) -> Callable:
+        self.__check_if_constructed_without_error(k)
         self.__check_if_unwrapped(k)
         raise AttributeError("{} has no attribute {}".format(self.__class__, k))
 
+    def __check_if_constructed_without_error(self, k: str) -> None:
+        if k in ("_val", "_class"):
+            raise ValueError("Object did not get properly constructed - probably due to an earlier unhandled Error.")
+
```
migwellian: this is unrelated to this PR, but it sometimes bites me - typically when a ctor throws an exception, and in the exception there is some kind of error string that ends up calling `__getattr___`

keanu-python/tests/test_base.py
```diff
@@ -27,7 +27,7 @@ def test_you_can_call_a_java_method_on_the_unwrapped_object(java_list_wrapper) -
 
 
 def test_you_cannot_call_a_java_method_with_snake_case_on_the_unwrapped_object(java_list_wrapper) -> None:
-    with pytest.raises(Py4JError, match="Method is_empty\(\[\]\) does not exist"):
+    with pytest.raises(Py4JError, match=r"Method is_empty\(\[\]\) does not exist"):
```
migwellian: unrelated to this PR - just to remove a warning in the tests

keanu-python/tests/test_labels.py
```diff
@@ -0,0 +1,26 @@
...
+
+def test_there_is_a_helper_method_to_build_the_string_from_a_list() -> None:
+    label = _VertexLabel.create_from_list("outer", "inner", "foo")
+    assert label.unwrap().getUnqualifiedName() == "foo"
+    assert label.get_name() == "outer.inner.foo"
+
+
+def test_the_repr_method_makes_it_clear_what_the_namespace_is() -> None:
+    label = _VertexLabel("outer.inner.foo")
+    assert str(label) == "['outer', 'inner', 'foo']"
```
GeorgeNash: what's this test doing?

keanu-python/keanu/vertex/label.py
```diff
@@ -9,6 +10,24 @@
 
 class _VertexLabel(JavaObjectWrapper):
```
icantsayrural: We might want to write a comment saying this is for internal use, so people don't try to pass it to `Vertex`.

keanu-python/keanu/vertex/label.py
```diff
@@ -9,6 +10,24 @@
...
         super(_VertexLabel, self).__init__(java_object)
+
+    def get_name(self) -> str:
+        return self.unwrap().getQualifiedName()
+
+    def __repr__(self) -> str:
+        return self.get_name().split(".").__repr__()
+
+    @staticmethod
+    def create_from_list(*names: str) -> '_VertexLabel':
```
icantsayrural: What happens when this list is empty?

keanu-python/keanu/base.py
icantsayrural: How does `_val` and `_class` get called by `__getattr__` in the first place? These are actually valid fields of `JavaObjectWrapper` that we set ourselves.

keanu-python/tests/test_labels.py
icantsayrural: Checking `__repr__`

keanu-python/keanu/base.py
migwellian: This happens if there's an exception in a constructor of a class that inherits from `JavaObjectWrapper`. In such a case, `_val` and `_class` have not been set.

migwellian:
> Why does `VertexLabel` feel so separate from `Vertex` in python? I expected to see it as a member variable in the `Vertex` class and we don't seem to call any of the methods from `Vertex`?

That's a good observation. We don't need to call it from `Vertex` because:
* for `set_label`, we can pass a string as the label
* for `get_label`, we call a Java method that returns a string.

This is only really used behind the scenes in places like `lambda_model` and (coming soon) `plates`.

keanu-python/keanu/vertex/label.py
migwellian: I've added a docstring:

        class _VertexLabel(JavaObjectWrapper):
        """
        This class is intended for private use - as a user you should only need to specify string values.
        It's used behind the scenes e.g. in :class:`keanu.vertex.lambda_model.LambdaModel`
        """

keanu-python/keanu/vertex/label.py
migwellian: good call - I've made it raise an exception in that case

PR #500, opened 2019-02-12
Deterministic Python Testing

christophernorth:
Python tests should now all automatically be deterministic (ie start with a known random seed and vertex ID values)

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
@@ -17,7 +17,7 @@ import java.util.concurrent.atomic.AtomicLong;
 @EqualsAndHashCode
 public class VertexId implements Comparable<VertexId>, VariableReference {
 
-    public static final AtomicLong ID_GENERATOR = new AtomicLong(0L);
+    private static final AtomicLong ID_GENERATOR = new AtomicLong(0L);
     private static final int TOP_LEVEL_ARRAY_SIZE = 1;
 
     private long[] idValues = new long[TOP_LEVEL_ARRAY_SIZE];
@@ -83,5 +83,9 @@ public class VertexId implements Comparable<VertexId>, VariableReference {
 
     public long[] getValue() {
         return Arrays.copyOf(idValues, idValues.length);
+    }
+
+    public static void resetIdGenerator() {
+        ID_GENERATOR.set(0);
     }
 }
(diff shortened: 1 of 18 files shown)


keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
```diff
@@ -84,4 +84,8 @@ public int getIndentation() {
     public long[] getValue() {
         return Arrays.copyOf(idValues, idValues.length);
     }
+
+    public static void resetIdGenerator() {
+        ID_GENERATOR.set(0);
```
GeorgeNash: Were we setting this to `1` for any particular reason before?

keanu-python/keanu/keanu_random.py
```diff
@@ -22,3 +23,8 @@ def next_double(self) -> float:
     @staticmethod
     def set_default_random_seed(seed: int) -> None:
         k.jvm_view().KeanuRandom.setDefaultRandomSeed(seed)
+
+
+def set_deterministic_state():
```
GeorgeNash: should this be a static method?

keanu-python/tests/conftest.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+
+from keanu import set_deterministic_state
+
+
+@pytest.fixture(autouse=True)
```
GeorgeNash: nice

keanu-python/keanu/keanu_random.py
```diff
@@ -22,3 +23,8 @@ def next_double(self) -> float:
...
+    KeanuRandom.set_default_random_seed(0)
+    k.jvm_view().VertexId.resetIdGenerator()
```
GeorgeNash: can we write a quick test to make sure it resets the Id as expected? Or have we got one on the Java side that covers it

keanu-project/src/main/java/io/improbable/keanu/vertices/VertexId.java
christophernorth: Not really - it was just a good known value - I've decided to set it 0 everywhere just as that's the value we start at normally so it seems more consistent.

keanu-python/keanu/keanu_random.py
christophernorth: I personally think it's better just being a function given it's not specifically related to the KeanuRandom object - it's just something that stops things being non-deterministic and interfaces with the KeanuRandom class to do that.  Happy to hear more thoughts though.

keanu-python/keanu/keanu_random.py
christophernorth: Good suggestion - I'll knock up a super quick test

keanu-python/tests/conftest.py
christophernorth: 👍

PR #501, opened 2019-02-12
DotSaver saves disconnected vertices

icantsayrural:


keanu-project/src/main/java/io/improbable/keanu/util/io/DotSaver.java
```diff
@@ -59,45 +61,64 @@ public DotSaver(BayesianNetwork network) {
...
+     * @param vertices   vertices around which the subGraphs will be centered
      * @param degree     degree of connections to be visualized; for instance, if the degree is 1,
-     *                   only connections between the vertex and its parents and children will be written out to the DOT file.
+     *                   only connections between the vertices and its parents and children will be written out to the DOT file.
      * @param saveValues specify whether you want to output values of non-constant scalar vertices
      * @throws IOException Any errors that occur during saving to the output stream
      */
-    public void save(OutputStream output, Vertex vertex, int degree, boolean saveValues) throws IOException {
-        save(output, vertex, degree, saveValues, null);
+    public void save(OutputStream output, List<Vertex> vertices, int degree, boolean saveValues) throws IOException {
```
GeorgeNash: Add a helper `save` method that takes a single Vertex and creates a `Collections.singletonList()` and calls into this method. It's annoying in your tests having to make lists all the time for the use case of a single vertex

keanu-project/src/test/java/io/improbable/keanu/util/io/DotSaverTest.java
```diff
@@ -150,6 +152,52 @@ public void metadataIsWrittenOut() throws IOException {
...
+        BayesianNetwork bayesianNetwork = new BayesianNetwork(Arrays.asList(v1, v2, gamma1, gaussian1, v3, v4, gamma2, gaussian2));
+        DotSaver dotSaver = new DotSaver(bayesianNetwork);
+        dotSaver.save(outputWriter, Arrays.asList(gaussian1, gaussian2), 1, true);
+        String expectedOutputWithValues = readFileToString(VERTEX_DEGREE1__OUTPUT_WITH_DISCONNECTED_VERTICES_FILENAME);
+        checkDotFilesMatch(outputWriter.toString(), expectedOutputWithValues);
+    }
+
+    @Test
+    public void dotSaveShowsAllDisconnectedVertices() throws IOException {
+        DoubleVertex v1 = new ConstantDoubleVertex(0.);
```
GeorgeNash: Rather than doing this twice can we do it in a:

```
@Before
public void setupDisconnectedBayesNet() {
}
```

keanu-project/src/main/java/io/improbable/keanu/util/io/DotSaver.java
```diff
@@ -48,56 +46,35 @@ public DotSaver(BayesianNetwork network) {
     }
```
migwellian: `bayesianNetwork` is only used once, to `getAllVertices`.

If you replace the `BayesianNetwork` field with a `Set<Vertex>` field, then the user still has the option to construct the `DotSaver` with `bayesianNetwork.getSubgraph(vertex, degree)` if they wish.

migwellian: You can also keep the constructor `DotSaver(BayesianNetwork)` as a convenience if you like.

    public DotSaver(BayesianNetwork network) {
        this(network.getAllVertices);
    }

PR #504, opened 2019-02-13
getProbabilisticParents added to Vertex.

icantsayrural:


keanu-python/tests/test_vertex.py
```diff
@@ -367,3 +367,28 @@ def test_java_vertex_to_python_vertex_persists_label() -> None:
...
+    assert [p1.get_id()] == v_parents
+
+
+def test_get_children() -> None:
+    p1 = Const(0.)
+    p2 = Const(1.)
+    v = Gaussian(p1, p2)
+    p1_children = [c.get_id() for c in p1.get_children()]
+    p2_children = [c.get_id() for c in p2.get_children()]
+    assert [v.get_id()] == p1_children == p2_children
```
migwellian: nit: split this into 2 lines.

Firstly because, although it's valid in Python (which surprised me), it's not obvious that the logic is correct - I expected it to evaluate the first equality as `True` and then evaluating the second equality as `True == p2_children`. 

Secondly because if it failed, the error message wouldn't tell you which of the 3 terms was unequal to the other 2.

release_notes.md
```diff
@@ -1,9 +1,12 @@
 ### Version 0.0.20 ###
 
 ## Common
+* Added `get_probabilistic_parents` to bayesian network
```
migwellian: nit: we should use camel case in the "Common" section

PR #508, opened 2019-02-14
 Use pip3 and python3

wesleyw72:
We use Python 3 so we should be using pip3 and python3 commands. It's not a guarantee that pip links through to pip3 on all systems so we should explicitly call the correct version. Specifically hitting this issue on Ubuntu 16.04 build agents.

### DIFF
--- a/keanu-python/build.gradle
+++ b/keanu-python/build.gradle
@@ -28,7 +28,7 @@ task pythonVersionInfo {
         }
         exec {
             // NB: the pip version inside pipenv is not necessarily the same as the version outside pipenv
-            commandLine 'pipenv', 'run', 'pip', '--version'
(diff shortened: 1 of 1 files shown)


icantsayrural:
Do we also have to change line `commandLine 'pipenv', 'run', 'pip', '--version'` ?

GeorgeNash:
And change the `python` commands to `python3`?

wesleyw72:
Yeah both good comments, will do some testing to make sure that works (on my machine and double check on a build agent).

migwellian:
The `pip` inside your `pipenv` can be different to the `pip` outside you `pipenv`.

I expect that inside your `pipenv` you don't need to specify `pip3`, but it would still be better to be explicit and use `pip3` throughout.

PR #512, opened 2019-02-14
Python methods returning generators renamed to `iter_*`.

icantsayrural:
More specifically, methods calling `_to_generator`.


migwellian:
Heads up when you merge `develop` into this branch: you'll need to change `get_parents` and `get_children` too

icantsayrural:
> Is `iter_` a Python convention, or something we've made up?

It's Python convention

PR #513, opened 2019-02-14
Ignore generated python from spotless

wesleyw72:
## Problem description
Currently when building on Windows, the `spotlessPython` step will pass the first time the code is run however on further runs it will fail due to formatting errors in the generated python files.

## Cause
Spotless checks that the line endings are the ones that the user's OS uses (unless specified otherwise in gitattributes). For windows this is CRLF and for unix this LF. The files are currently being generated using LF line endings on all OS's.

## Solution
We don't really need to run any formatting checks on the generated files so for the time being they can be ignored.

## Testing
I was hitting this problem when running multiple times on a build agent. This fix stops this problem.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,6 +5,7 @@
 # to native line endings on checkout.
 *.java text
 *.py text
+*.ftl text
 
 # Declare files that will always have CRLF line endings on checkout.
 # *.sln text eol=crlf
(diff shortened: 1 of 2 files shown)


wesleyw72:
Discussed offline, `core.autocrlf true` won't fix this problem. The files are being checkedout by git correctly with the intended line endings. It is the generation of python code which is giving them the wrong line endings.

christophernorth:
As discussed offline - this is probably not what we want long term (as we might want a bit of code to get checked in the generated files (ie that which we've put in the template file itself) but is a reasonable workaround for now until we have time to sort it fully.

wesleyw72:
@christophernorth raised a JIRA ticket for this.

PR #524, opened 2019-02-21
Automated PyPi releasing

wesleyw72:
This PR introduces a few things:
- A script to simplify the release of Keanu Python (from CI and manually)
- Pipeline for automating releases
- Change script naming to conform to company style guide

An example of the pipeline can be seen here: 
https://buildkite.com/improbable/keanu-release-python/builds/15#2d809e8b-1270-4860-956e-694776211eec

It has been tested for releasing manually and through CI to PyPi testing.

The pipeline "should" be triggered on pushes to master.


ci/build_python.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -e -u -o pipefail
+
+if [[ -n "${DEBUG-}" ]]; then
+  set -x
+fi
+
+cd "$(dirname "$0")/../"
+
+./gradlew clean build preparePythonClasspath buildWheelDistribution --info --stacktrace
```
migwellian: doesn't `preparePythonClasspath` happen as part of `build`?

ci/release_python.sh
```diff
@@ -0,0 +1,79 @@
...
+set -e -u -o pipefail
+
+if [[ -n "${DEBUG-}" ]]; then
+  set -x
+fi
+
+cd "$(dirname "$0")/"
+
+PYPI_TEST_SECRET="keanu-eng-test-pypi"
+PYPI_REAL_SECRET="keanu-eng-pypi"
```
migwellian: These variable names suggest that they are secrets in plain text - I assume you mean they are the names of the secrets, that are stored in the vault?

ci/build_python.sh
wesleyw72: Yes. However, if this changes in the future for whatever reason it will break this script without the changer noticing. I'd rather be explicit about what tasks are required and let gradle figure out what what to do.

ci/release_python.sh
wesleyw72: Yeah good point. You're correct they are the names of the secrets, I have changed the naming to reflect this.

ci/build_python.sh
migwellian: I don't think this makes a lot of sense. 

Firstly, because `build` always depends on `test`, and the python tests would fail if you hadn't done `preparePythonClasspath`. 

Secondly, because your order is weird. Gradle will just dumbly execute the tasks in order. So you're telling it first to do a full `build` (it will build every project*, including `python`) and then to do the `preparePythonClasspath` (again). 

So at the very least, I'd suggest swapping the order of `preparePythonClassPath` and `build`. But even so, I don't think it's necessary.

(*) P.S. Maybe as an optimisation you want `:keanu-python:build` instead of `build`? So that it doesn't waste time on unrelated tasks e.g. `:docs:build`.

wesleyw72: Yeah fair enough.

Oh really? I thought gradle would be smart enough to know that `preparePythonClasspath` had been done.

I'll change it to be `clean preparePythonClasspath buildWheelDistribution`.

PR #525, opened 2019-02-21
throw an error if concatenation vertex is given operands with ranks less than or equal to dimension

icantsayrural:


keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShapeValidation.java
```diff
@@ -198,6 +198,10 @@ public static void checkShapesMatch(long[] actual, long[] expected) {
         long[] concatShape = Arrays.copyOf(shapes[0], shapes[0].length);
 
         for (int i = 1; i < shapes.length; i++) {
+            if (shapes[i].length <= dimension) {
+                throw new IllegalArgumentException("Cannot concat shapes at a dimension greater than or equal to their ranks");
```
migwellian: include `i` in the message

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShapeValidation.java
```diff
@@ -198,6 +198,10 @@ public static void checkShapesMatch(long[] actual, long[] expected) {
         long[] concatShape = Arrays.copyOf(shapes[0], shapes[0].length);
 
         for (int i = 1; i < shapes.length; i++) {
+            if (shapes[i].length <= dimension) {
+                throw new IllegalArgumentException(String.format("Cannot concat operand %d because dimension is greater than or equal to its rank", i));
```
migwellian: you should include `shapes[i].length` and `dimension` in the error message

PR #532, opened 2019-02-26
Support for vector-vector and matrix-vector product.

icantsayrural:
Note that the reshape is not done at the tensor level, because that will interfere with auto diff.

The new behaviour of matrix multiply was copied from numpy's `matmul`.

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
@@ -100,7 +100,8 @@ public abstract class DoubleVertex extends Vertex<DoubleTensor> implements Doubl
      * Matrix product of two vertices
      *
      * @param that  a double vertex to matrix multiply
-     * @return a vertex that represents the matrix multiplication of two vertices.
+     * @return a vertex that represents the matrix multiplication of two vertices. Note that this operation
+     * does that always return a MatrixMultiplicationVertex.
      * - If both left and right operands are rank 2, they are multiplied like conventional matrices.
      * - If both left and right operands are rank 1, they are promoted to a matrix by prepending a 1 to its dimensions.
      *   After matrix multiplication, it is reshaped to be a scalar. This is essentially a dot product.
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
@@ -96,6 +96,26 @@ public abstract class DoubleVertex extends Vertex<DoubleTensor> implements Doubl
         return new MultiplicationVertex(this, that);
     }
 
+    /*
+        - If both arguments are rank 2 they are multiplied like conventional matrices.
+        - If the first argument has rank 1, it is promoted to a matrix by prepending a 1 to its dimensions.
+          After matrix multiplication the prepended 1 is removed.
+        - If the second argument has rank1 1, it is promoted to a matrix by appending a 1 to its dimensions.
+          After matrix multiplication the appended 1 is removed.
+        - If both arguments are rank 1, dot product is applied.
+     */
+
+    /**
+     * Matrix product of two vertices
+     *
+     * @param that  a double vertex to matrix multiply
+     * @return a vertex that represents the matrix multiplication of two vertices.
+     * - If both left and right operands are rank 2, they are multiplied like conventional matrices.
+     * - If both left and right operands are rank 1, they are promoted to a matrix by prepending a 1 to its dimensions.
+     *   After matrix multiplication, it is reshaped to be a scalar. This is essentially a dot product.
+     * - If only one of the operands are rank 1, it is promoted to a matrix by prepending a 1 to its dimensions.
+     *   After matrix multiplication, the appended 1 is removed. This is essentially a matrix-vector product.
+     */
     public DoubleVertex matrixMultiply(DoubleVertex that) {
         int leftRank = this.getRank();
         int rightRank = that.getRank();
--- a/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
@@ -96,15 +96,6 @@ public abstract class DoubleVertex extends Vertex<DoubleTensor> implements Doubl
         return new MultiplicationVertex(this, that);
     }
 
-    /*
-        - If both arguments are rank 2 they are multiplied like conventional matrices.
-        - If the first argument has rank 1, it is promoted to a matrix by prepending a 1 to its dimensions.
-          After matrix multiplication the prepended 1 is removed.
-        - If the second argument has rank1 1, it is promoted to a matrix by appending a 1 to its dimensions.
-          After matrix multiplication the appended 1 is removed.
-        - If both arguments are rank 1, dot product is applied.
-     */
-
     /**
      * Matrix product of two vertices
      *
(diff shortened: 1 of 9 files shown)


keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
```diff
@@ -96,8 +96,38 @@ public MultiplicationVertex multiply(DoubleVertex that) {
...
+     * - If both left and right operands are rank 1, they are promoted to a matrix by prepending a 1 to its dimensions.
+     *   After matrix multiplication, it is reshaped to be a scalar. This is essentially a dot product.
+     *   This returns a ReshapeVertex.
+     * - If only one of the operands is rank 1 (and the other operand is rank 2), it is promoted to a matrix by prepending a 1 to its dimensions.
+     *   After matrix multiplication, the appended 1 is removed. This is essentially a matrix-vector product.
+     *   This returns a ReshapeVertex.
+     * - Otherwise, they are multiplied like conventional matrices.
+     *   This returns a MatrixMultiplicationVertex.
+     */
+    public DoubleVertex matrixMultiply(DoubleVertex that) {
```
migwellian: I wonder if it would be better to return a `ReshapeVertex` under all circumstances? That way you don't need all the dodgy casts to `MatrixMultiplicationVertex` in your tests.

(What you really care about is that you get back a `DoubleVertex` that implements `Differentiable`)

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/DoubleVertexTest.java
```diff
@@ -98,4 +111,41 @@ public void canCastToIntegerVertex() {
...
+    public void canMatrixMultiplyVectors() {
+        DoubleVertex vectorsMultiplied = ConstantVertex.of(vectorA).matrixMultiply(ConstantVertex.of(vectorA));
+        DoubleTensor result = vectorsMultiplied.lazyEval();
+        DoubleTensor expectedResult = DoubleTensor.scalar(5.);
+        assertThat(result, valuesAndShapesMatch(expectedResult));
+    }
+
+    @Test
+    public void canMatrixMultiplyVectorAndMatrix() {
+        DoubleVertex vectorsMultiplied = ConstantVertex.of(vectorA).matrixMultiply(ConstantVertex.of(vectorB.reshape(2, 3)));
```
migwellian: I think for clarity it's worth pulling out `vectorB.reshape(2, 3)` as a field named `matrix`

migwellian: nit: `vectorsMultiplied` is an inaccurate name

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/DoubleVertexTest.java
```diff
@@ -2,17 +2,30 @@
...
 
+import static io.improbable.keanu.tensor.TensorMatchers.valuesAndShapesMatch;
+import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 
 public class DoubleVertexTest {
 
+    private DoubleTensor vectorA;
+    private DoubleTensor vectorB;
```
migwellian: Your test code would be more concise if these were vertices, i.e.

    DoubleVertex vectorA = ConstantVertex.of(new double[]{1., 2.}, 2);
    DoubleVertex vectorB = ConstantVertex.of(new double[]{1., 2., 3., 4., 5., 6.}, 6);

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
icantsayrural: Hmmm, I didn't mind it too much because the stuff we are doing in those tests aren't supposed to be used directly by the average user.

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/DoubleVertexTest.java
icantsayrural: 😊

icantsayrural: Good eye, some legacy left from copying the logic from my tensor implementation 😅

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
migwellian: ok, let's leave it

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
```diff
@@ -96,8 +96,38 @@ public MultiplicationVertex multiply(DoubleVertex that) {
...
+        int leftRank = this.getRank();
+        int rightRank = that.getRank();
+
+        DoubleVertex leftMatrix = leftRank == 1 ? this.reshape(1, this.getShape()[0]) : this;
+        DoubleVertex rightMatrix = rightRank == 1 ? that.reshape(that.getShape()[0], 1) : that;
```
migwellian: We should probably throw a helpful message if either rank is zero.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
gordoncaleb: The reshape vertex might have a performance impact if added where it's not actually needed. Although, the performance impact might be worth avoiding the extra code complexity.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
```diff
@@ -96,8 +96,42 @@ public MultiplicationVertex multiply(DoubleVertex that) {
...
+        if (leftRank < 1 || rightRank < 1) {
+            throw new IllegalArgumentException("Matrix multiply for rank 0 is not supported. Use times instead.");
+        }
+
+        DoubleVertex leftMatrix = leftRank == 1 ? this.reshape(1, this.getShape()[0]) : this;
+        DoubleVertex rightMatrix = rightRank == 1 ? that.reshape(that.getShape()[0], 1) : that;
+
+        MatrixMultiplicationVertex result = new MatrixMultiplicationVertex(leftMatrix, rightMatrix);
+
+        if (leftRank == 1 && rightRank == 1) {
```
gordoncaleb: Nice. A bit more complicated but nice.

keanu-project/src/main/java/io/improbable/keanu/vertices/dbl/DoubleVertex.java
```diff
@@ -96,8 +96,42 @@ public MultiplicationVertex multiply(DoubleVertex that) {
...
+            return result.reshape();
```
gordoncaleb: Does this reshape to a scalar? Can you reshape to `new long[0]` so it's a bit more explicit that you're shaping it to a scalar?

keanu-project/src/test/java/io/improbable/keanu/vertices/dbl/nonprobabilistic/operators/binary/MatrixMultiplicationVertexTest.java
```diff
@@ -39,7 +39,7 @@ public void canDoMatrixMultiply2x2() {
         UniformVertex b = new UniformVertex(0, 10);
         b.setValue(DoubleTensor.create(new double[]{5, 6, 7, 8}, 2, 2));
 
-        MatrixMultiplicationVertex c = a.matrixMultiply(b);
+        MatrixMultiplicationVertex c = (MatrixMultiplicationVertex) a.matrixMultiply(b);
```
gordoncaleb: Do you need this cast? Can it return something that doesn't need to be cast?

icantsayrural: You do if you explicitly call `.forwardModeAutoDiff`, so that the parameter types match.

PR #538, opened 2019-03-05
duplicateKeysAreIgnoredWhenYouStream

migwellian:
Fix for a bug reported by @chris-major-improbable


keanu-project/src/main/java/io/improbable/keanu/algorithms/mcmc/SamplingAlgorithm.java
```diff
@@ -12,6 +12,7 @@
 
     static Map<VariableReference, ?> takeSample(List<? extends Variable> sampleFromVariables) {
         return sampleFromVariables.stream()
```
christophernorth: Given we call this per sample, are we happy we're adding cost for a case of "User Error"?  Could we filter this once at the beginning of of the sampling run?

gordoncaleb: Maybe make it a Set? We can do this distinct on construction at the very least.

PR #541, opened 2019-03-13
Sequences accept interlinked factories

CharlieCrisp:
This PR introduces some changes to the `SequenceBuilder` so that now factories can use proxy input vertices that are defined in other factories. This means factories can provide inputs to other factories.


keanu-python/keanu/sequence.py
```diff
@@ -59,17 +62,30 @@ def __init__(self,
...
 
         if count is not None:
-            function = lambda p: factory(SequenceItem(p))
-            consumer = Consumer(function)
-            builder = builder.count(count).withFactory(consumer)
+            if factories is not None:
+                functions = [Consumer(lambda p, f=f: f(SequenceItem(p))) for f in factories]
+                java_functions = ListConverter().convert(functions, k._gateway._gateway_client)
+                builder = builder.count(count).withFactories(java_functions)
+            elif factory is not None:
```
migwellian: Having 2 code paths here (and on line 84) increases the complexity. Instead you can just turn a singleton `factory` into a list at the start of this function:

    if factory is not None:
        factories = [factory]

migwellian: Or if you implement my suggestion to allow `factories` to be either a `Consumer` or an `Iterable[Consumer]` then you'd do:

    if isinstance(factories, Consumer):
        factories = [factories]

keanu-python/keanu/sequence.py
```diff
@@ -59,17 +62,30 @@ def __init__(self,
...
+                consumer = Consumer(function)
+                builder = builder.count(count).withFactory(consumer)
 
         if data_generator is not None:
-            bifunction = lambda p, data: factory(SequenceItem(p), data)
-            biconsumer = BiConsumer(bifunction)
-            data_generator_java = (k.to_java_map(m) for m in data_generator)
-            builder = builder.fromIterator(JavaIterator(data_generator_java)).withFactory(biconsumer)
+            if factories is not None:
+                bifunctions = [Consumer(lambda p, data: factory(SequenceItem(p), data)) for factory in factories]
```
migwellian: presumably this also suffers from the "Late Binding Closures" gotcha?

keanu-python/keanu/vertex/base.py
```diff
@@ -80,6 +80,12 @@ def get_label(self) -> Optional[str]:
         label = self.unwrap().getLabel()
         return None if label is None else label.getQualifiedName()
 
+    def get_label_without_outer_namespace(self) -> Optional[str]:
+        label = self.unwrap().getLabel()
+        if label is None:
+            return None
+        return label.withoutOuterNamespace().getQualifiedName()
```
migwellian: for consistency with the method above I'd prefer `return None if label is None else ...`

keanu-python/tests/test_sequence.py
```diff
@@ -97,3 +98,80 @@ def create_time_step(sequence_item):
...
+
+    sequence = Sequence(count=5, factories=factories, initial_state=initial_state)
+
+    assert sequence.size() == 5
+
+    for item in sequence:
+        __check_sequence_output_links_to_input(item, x1_input_label, x1_label)
+        __check_sequence_output_links_to_input(item, x2_input_label, x3_label)
+        __check_sequence_output_links_to_input(item, x3_input_label, x2_label)
+        __check_sequence_output_links_to_input(item, x4_input_label, x4_label)
```
migwellian: same comment as for the java test - you can assert about the 4 output values

docs/src/test/java/io/improbable/snippet/SequenceExample.java
```diff
@@ -72,4 +77,47 @@ public Sequence buildSequence(String dataFileName) {
...
+        //%%SNIPPET_START%% SequenceTimeSeries
+        DoubleVertex two = new ConstantDoubleVertex(2);
+
+        // Define the labels of vertices we will use in our Sequence
+        VertexLabel x1Label = new VertexLabel("x1");
+        VertexLabel x2Label = new VertexLabel("x2");
+
+        // Define 'Proxy Vertices' which we use as placeholders when defining our sequence for the equivalent vertex
+        // in the previous sequence.
+        // I.e. these are the inputs to our sequence
```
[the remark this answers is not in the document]
CharlieCrisp: Ah yes - that is what I meant but didn't articulate properly :)

keanu-python/keanu/sequence.py
CharlieCrisp: Yep - have used the `partial` approach on this one now too :)

keanu-python/keanu/sequence.py
```diff
@@ -59,17 +64,30 @@ def __init__(self,
...
 
         if count is not None:
-            function = lambda p: factory(SequenceItem(p))
-            consumer = Consumer(function)
-            builder = builder.count(count).withFactory(consumer)
+
+            def wrap_function_argument_in_sequence_item(f, p):
+                return f(SequenceItem(p))
+
+            functions = [Consumer(partial(wrap_function_argument_in_sequence_item, f)) for f in factories]
```
migwellian: you can use a lambda here - no need for `wrap_function_argument_in_sequence_item`

keanu-python/keanu/sequence.py
```diff
@@ -59,17 +64,30 @@ def __init__(self,
...
+            builder = builder.count(count).withFactories(java_functions)
 
         if data_generator is not None:
-            bifunction = lambda p, data: factory(SequenceItem(p), data)
-            biconsumer = BiConsumer(bifunction)
+
+            def wrap_function_arguments_in_sequence_item(f, p, data):
+                return f(SequenceItem(p), data)
+
+            bifunctions = [BiConsumer(partial(wrap_function_arguments_in_sequence_item, f)) for f in factories]
```
migwellian: you can use a lambda here - no need for `wrap_function_argument_in_sequence_item`

PR #542, opened 2019-03-19
Vertex Descriptions

CharlieCrisp:
This PR adds method `.createDescription()` on vertices which will create a String to describe how this vertex is generated. It does a DFS of parents, stopping at leaves or labelled vertices and constructs equations of the form such as `"This Vertex = that + (Const(3.0) * Four)"`

Note: The operators which have been supported in this method are currently limited and the default is to list parents separated by commas.


keanu-project/src/main/java/io/improbable/keanu/vertices/Vertex.java
```diff
@@ -312,4 +312,69 @@ public void saveValue(NetworkSaver netSaver) {
     public void loadValue(NetworkLoader loader) {
         loader.loadValue(this);
     }
+
```
gordoncaleb: Can we decouple this description from the Vertex class?

keanu-project/src/main/java/io/improbable/keanu/util/DescriptionCreator.java
```diff
@@ -0,0 +1,178 @@
...
+            String valueString = value.isScalar() ? value.scalar().toString() :
+                new StringBuilder("(Tensor with shape: ").append(Arrays.toString(value.getShape())).append(")").toString();
+
+            return new StringBuilder()
+                .append("Const(")
+                .append(valueString)
+                .append(")")
+                .toString();
+        } else {
+            return "<<Unlabelled leaf Vertex>>";
```
christophernorth: In general I'd shy away from logic like this.  In particular, this function makes assumptions about how other functions behave.  Whilst correct at the moment, someone might not recognise there's an implicit dependency between those two functions.

keanu-project/src/main/java/io/improbable/keanu/util/DescriptionCreator.java
```diff
@@ -0,0 +1,178 @@
...
+        booleanBinaryOperators.put(EqualsVertex.class, " == ");
+        booleanBinaryOperators.put(GreaterThanOrEqualVertex.class, " >= ");
+        booleanBinaryOperators.put(GreaterThanVertex.class, " > ");
+        booleanBinaryOperators.put(LessThanOrEqualVertex.class, " <= ");
+        booleanBinaryOperators.put(LessThanVertex.class, " < ");
+        booleanBinaryOperators.put(NotEqualsVertex.class, " != ");
+        booleanBinaryOperators.put(OrBinaryVertex.class, " || ");
+    }
+
+    private static <T extends Tensor> String getBaseDescription(Vertex<T> vertex) {
```
christophernorth: Should we perhaps call this getLeafDescription given that's sort of what we're doing?

keanu-project/src/main/java/io/improbable/keanu/util/DescriptionCreator.java
```diff
@@ -0,0 +1,178 @@
...
+                .append("Const(")
+                .append(valueString)
+                .append(")")
+                .toString();
+        } else {
+            return "<<Unlabelled leaf Vertex>>";
+        }
+    }
+
+    static <T extends Tensor> String createDescriptionAllowingLabels(Vertex<T> vertex) {
```
christophernorth: Your permissions on your methods are a bit inconsistent - is there a reason this is package private, but others are private?

keanu-project/src/main/java/io/improbable/keanu/util/DescriptionCreator.java
```diff
@@ -0,0 +1,178 @@
...
+        builder.append(String.join(delimiter, parentStrings));
+        if (includeBrackets) {
+            builder.append(")");
+        }
+
+        return builder.toString();
+    }
+
+    /**
+     * his method constructs an equation to describe how a vertex is calculated.
```
christophernorth:
```suggestion
     * This method constructs an equation to describe how a vertex is calculated.
```

keanu-project/src/main/java/io/improbable/keanu/util/DescriptionCreator.java
```diff
@@ -0,0 +1,178 @@
...
+import java.util.stream.Stream;
+
+import static io.improbable.keanu.util.DescriptionUtils.createBinomialDescription;
+import static io.improbable.keanu.util.DescriptionUtils.createBooleanBinaryOpDescription;
+import static io.improbable.keanu.util.DescriptionUtils.createIfStringDescription;
+
+public class DescriptionCreator {
+
+    private static Map<Class, String> delimiters = new HashMap<>();
+    private static Map<Class, String> booleanBinaryOperators = new HashMap<>();
```
christophernorth: Should we keep these maps around all the time, even if no one ever makes use of this class?  I know it's not a major memory consumer, but should we just make this an instance property that gets set up on creation?

keanu-project/src/main/java/io/improbable/keanu/util/DescriptionCreator.java
```diff
@@ -0,0 +1,178 @@
...
+        delimiters.put(IntegerMultiplicationVertex.class, " * ");
+        delimiters.put(DivisionVertex.class, " / ");
+        booleanBinaryOperators.put(AndBinaryVertex.class, " && ");
+        booleanBinaryOperators.put(EqualsVertex.class, " == ");
+        booleanBinaryOperators.put(GreaterThanOrEqualVertex.class, " >= ");
+        booleanBinaryOperators.put(GreaterThanVertex.class, " > ");
+        booleanBinaryOperators.put(LessThanOrEqualVertex.class, " <= ");
+        booleanBinaryOperators.put(LessThanVertex.class, " < ");
+        booleanBinaryOperators.put(NotEqualsVertex.class, " != ");
+        booleanBinaryOperators.put(OrBinaryVertex.class, " || ");
```
christophernorth: Why are Boolean binary operators different to the non-bool ones?

keanu-project/src/main/java/io/improbable/keanu/util/DescriptionUtils.java
```diff
@@ -0,0 +1,61 @@
...
+        builder.append(" : ");
+        builder.append(createDescriptionAllowingLabels(els));
+        if (includeBrackets) {
+            builder.append(")");
+        }
+
+        return builder.toString();
+    }
+
+    static String createBinomialDescription(BinomialVertex binomialVertex, boolean includeBrackets) {
```
christophernorth: We could use the save/load annotations for random variables and auto generate reasonable looking things for all of them)?

keanu-project/src/main/java/io/improbable/keanu/util/DescriptionCreator.java
```diff
@@ -0,0 +1,178 @@
...
+        if (irregularDescription != null) {
+            return irregularDescription;
+        }
+
+        Stream<String> parentStream = vertex
+            .getParents()
+            .stream()
+            .map(DescriptionCreator::createDescriptionAllowingLabels);
+
+        String[] parentStrings = parentStream.toArray(String[]::new);
```
christophernorth: Is there a reason you do this in two steps?

keanu-project/src/main/java/io/improbable/keanu/util/DescriptionCreator.java
```diff
@@ -0,0 +1,178 @@
...
+     * @param vertex The vertex to create the description of
+     * @return An String equation describing how this vertex is calculated.<br>
+     * E.g. "This Vertex = that + (three * Const(4))"
+     */
+    public static <T extends Tensor> String createDescription(Vertex<T> vertex) {
+        Collection<Vertex> parents = vertex.getParents();
+
+        if (parents.size() == 0) {
+            if (vertex.getLabel() == null) {
+                return "<<Unlabelled Leaf Vertex>>";
```
christophernorth: Probably best to try and limit the logic so you don't have to check this in multiple places, but if we really need this we should pull out the string to keep it consistent.

keanu-project/src/main/java/io/improbable/keanu/util/DescriptionCreator.java
```diff
@@ -0,0 +1,178 @@
...
+            }
+            return "Vertex with no parents and label: " + vertex.getLabel().toString();
```
christophernorth:
```suggestion

            return "Vertex with no parents and label: " + vertex.getLabel().toString();
```

keanu-project/src/main/java/io/improbable/keanu/util/DescriptionUtils.java
```diff
@@ -0,0 +1,61 @@
...
+import io.improbable.keanu.vertices.Vertex;
+import io.improbable.keanu.vertices.bool.BooleanVertex;
+import io.improbable.keanu.vertices.bool.nonprobabilistic.operators.binary.BooleanBinaryOpVertex;
+import io.improbable.keanu.vertices.intgr.probabilistic.BinomialVertex;
+
+import static io.improbable.keanu.util.DescriptionCreator.createDescriptionAllowingLabels;
+
+class DescriptionUtils {
+
+    static String createIfStringDescription(BooleanVertex predicate, Vertex thn, Vertex els, boolean includeBrackets) {
```
christophernorth: I'm not sure what qualifies a function to move out of DescriptionCreator and in to this file - feels a tiny bit arbitrary?

keanu-project/src/main/java/io/improbable/keanu/util/DescriptionCreator.java
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        String thisLabel = vertex.getLabel() != null ? vertex.getLabel().toString() : "This Vertex";
+
+        return thisLabel + " = " + recursiveDescriptionStep(vertex, false);
+    }
+
+    private static <T extends Tensor> String getDescriptionDelimiter(Vertex<T> vertex) {
+        return delimiters.getOrDefault(vertex.getClass(), ", ");
```
christophernorth: If you use, for instance, unary operators, this will produce a pretty weird string?  Perhaps we could use the vertex class name or something at least?  Or as mentioned below, some of the save/load annotations?

keanu-project/src/main/java/io/improbable/keanu/util/DescriptionCreator.java
CharlieCrisp: @christophernorth I'm not sure I understand - which function is this logic making assumptions about?

keanu-project/src/main/java/io/improbable/keanu/util/DescriptionCreator.java
CharlieCrisp: Yes - for some stupid reason if I inline this then I get a compile error. Don't ask me why because I have no clue. More than happy to show you on my machine if you're interested ;)

keanu-project/src/main/java/io/improbable/keanu/util/DescriptionCreator.java
CharlieCrisp: Yeah good shout :)

keanu-project/src/main/java/io/improbable/keanu/util/DescriptionCreator.java
christophernorth: Ugh - I've hit similar things before, it's a pain!

keanu-project/src/main/java/io/improbable/keanu/util/DescriptionCreator.java
```diff
@@ -0,0 +1,325 @@
...
+            builder.append(")");
+        }
+
+        return Optional.of(builder.toString());
+    }
+
+    private Optional<String> tryCreateDescriptionFromSaveLoadAnnotations(Vertex vertex, boolean includeBrackets) {
+        Optional<Constructor<?>> vertexConstructor = Arrays.stream(vertex.getClass().getConstructors())
+            .filter(constructor -> Arrays.stream(constructor.getParameterAnnotations())
+                .anyMatch(annotationsList -> annotationsList.length > 0))
```
christophernorth: Potentially dangerous I guess given you're not checking for the types?

Instead of using the constructor, could you just use the parameters that we save?

keanu-project/src/main/java/io/improbable/keanu/util/DescriptionCreator.java
```diff
@@ -0,0 +1,325 @@
...
+        String vertexName = vertex.getClass().getSimpleName();
+        builder.append(vertexName);
+        builder.append("(");
+
+        String[] loadVertexParams = Arrays.stream(vertexConstructor.get().getParameterAnnotations())
+            .map(DescriptionCreator::tryGetLoadVertexParamAnnotation)
+            .filter(Optional::isPresent)
+            .map(Optional::get)
+            .map(LoadVertexParam::value)
+            .toArray(String[]::new);
```
christophernorth: nice streaming :-)

keanu-project/src/test/java/io/improbable/keanu/util/DescriptionGeneratorTest.java
```diff
@@ -0,0 +1,90 @@
...
+        BooleanVertex predicate6 = falze.and(falze);
+
+        assertThat(descriptionCreator.createDescription(predicate1), is("This Vertex = Const(2.0) >= three"));
+        assertThat(descriptionCreator.createDescription(predicate2), is("This Vertex = Const(2.0) > three"));
+        assertThat(descriptionCreator.createDescription(predicate3), is("This Vertex = Const(2.0) <= three"));
+        assertThat(descriptionCreator.createDescription(predicate4), is("This Vertex = Const(2.0) < three"));
+        assertThat(descriptionCreator.createDescription(predicate5), is("This Vertex = Const(false) || Const(false)"));
+        assertThat(descriptionCreator.createDescription(predicate6), is("This Vertex = Const(false) && Const(false)"));
+    }
+}
```
christophernorth: Do these tests cover all error cases fully?  Might be a good idea to run this test with coverage and make sure that you have error cases covered too?

keanu-project/src/main/java/io/improbable/keanu/util/DescriptionCreator.java
```diff
@@ -0,0 +1,325 @@
...
+            DoubleTensor constantValue = ((ConstantDoubleVertex) vertex).getConstantValue();
+            if (constantValue.isScalar()) {
+                scalarValue = Optional.of(constantValue.scalar().toString());
+            }
+        } else if (vertex instanceof ConstantIntegerVertex) {
+            IntegerTensor constantValue = ((ConstantIntegerVertex) vertex).getConstantValue();
+            if (constantValue.isScalar()) {
+                scalarValue = Optional.of(constantValue.scalar().toString());
+            }
+        } else if (vertex instanceof ConstantBooleanVertex) {
```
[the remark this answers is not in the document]
CharlieCrisp: Ooh that is cheeky but I like it

keanu-project/src/test/java/io/improbable/keanu/util/DescriptionGeneratorTest.java
CharlieCrisp: What do you mean 'to run this test with coverage?'. 
Also what sort of error cases could we run into here? I feel like this is all sufficiently wrapped that we shouldn't really get any errors, just maybe weird descriptions I guess?

CharlieCrisp: Yep this was a great shout - I've added a couple more test off the back of these results :)

PR #543, opened 2019-03-20
Expose SequenceItem contents

CharlieCrisp:
Exposes a `getContents()` method on `SequenceItem` which I have found useful for the purposes of debugging.
It allows a narrower view into the graph than `vertex.getConnectedGraph()` because it only shows vertices that have been explicitly added and only vertices from this sequence item.

This PR also adds the `get_last_item()` method to the python `Sequence` API to match the Java API

### DIFF
--- a/keanu-project/src/main/java/io/improbable/keanu/templating/SequenceItem.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/templating/SequenceItem.java
@@ -12,6 +12,8 @@ import java.util.Map;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+import static com.google.common.collect.ImmutableMap.copyOf;
+
 public class SequenceItem implements VertexDictionary {
 
     private static final String NAME_PREFIX = "Sequence_Item_";
@@ -61,7 +63,7 @@ public class SequenceItem implements VertexDictionary {
      * added to this sequence item.
      */
     public Map<VertexLabel, Vertex<?>> getContents() {
-        return this.contents;
+        return copyOf(this.contents);
     }
 
     private String getUniqueName() {
--- a/keanu-project/src/main/java/io/improbable/keanu/templating/SequenceItem.java
+++ b/keanu-project/src/main/java/io/improbable/keanu/templating/SequenceItem.java
@@ -56,6 +56,14 @@ public class SequenceItem implements VertexDictionary {
         return v;
     }
 
+    /**
+     * @return Returns a map of vertex labels to vertices, which covers all of the vertices that have been explicitly
+     * added to this sequence item.
+     */
+    public Map<VertexLabel, Vertex<?>> getContents() {
+        return this.contents;
+    }
+
     private String getUniqueName() {
         return NAME_PREFIX + this.hashCode();
     }
--- a/keanu-python/keanu/sequence.py
+++ b/keanu-python/keanu/sequence.py
@@ -31,6 +31,9 @@ class SequenceItem(JavaObjectWrapper):
     def get(self, label: str) -> Vertex:
         return Vertex._from_java_vertex(self.unwrap().get(_VertexLabel(label).unwrap()))
 
+    def get_contents(self) -> Dict[str, Vertex]:
+        return self.unwrap().getContents()
+
 
 class Sequence(JavaObjectWrapper):
     """
--- a/keanu-python/keanu/sequence.py
+++ b/keanu-python/keanu/sequence.py
@@ -32,12 +32,16 @@ class SequenceItem(JavaObjectWrapper):
         return Vertex._from_java_vertex(self.unwrap().get(_VertexLabel(label).unwrap()))
 
     def get_contents(self) -> Dict[str, Vertex]:
+
         def get_unqualified_name_or_proxy_name(key, vertex) -> str:
             if is_instance_of(k._gateway, vertex, "io.improbable.keanu.vertices.ProxyVertex"):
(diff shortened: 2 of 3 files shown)


keanu-python/keanu/sequence.py
```diff
@@ -31,6 +31,9 @@ def add(self, vertex: Vertex, label: Optional[str] = None) -> None:
     def get(self, label: str) -> Vertex:
         return Vertex._from_java_vertex(self.unwrap().get(_VertexLabel(label).unwrap()))
 
+    def get_contents(self) -> Dict[str, Vertex]:
+        return self.unwrap().getContents()
```
icantsayrural: You'll probably have to convert this to python map (of str to python Vertex)? Mypy is probably not complaining because there are no tests for it.

migwellian: +1

keanu-python/keanu/sequence.py
```diff
@@ -31,6 +31,9 @@ def add(self, vertex: Vertex, label: Optional[str] = None) -> None:
     def get(self, label: str) -> Vertex:
         return Vertex._from_java_vertex(self.unwrap().get(_VertexLabel(label).unwrap()))
 
+    def get_contents(self) -> Dict[str, Vertex]:
+        return {k.getUnqualifiedName(): Vertex._from_java_vertex(v) for k, v in self.unwrap().getContents().items()}
```
icantsayrural: I guess we'd only want the unqualified name because the namespace used here is autogenerated?

CharlieCrisp: yep - if you were trying to access vertices you'd have to use the label `Sequence_Item_****hash****.label`. The full label can still be obtained from the vertex itself if necessary.

keanu-project/src/main/java/io/improbable/keanu/templating/SequenceItem.java
```diff
@@ -56,6 +56,14 @@ public SequenceItem() {
         return v;
     }
 
+    /**
+     * @return Returns a map of vertex labels to vertices, which covers all of the vertices that have been explicitly
+     * added to this sequence item.
+     */
+    public Map<VertexLabel, Vertex<?>> getContents() {
+        return this.contents;
```
icantsayrural: Should this be an immutable copy?

CharlieCrisp: What would be the benefit of this?

keanu-python/keanu/sequence.py
```diff
@@ -31,6 +31,18 @@ def add(self, vertex: Vertex, label: Optional[str] = None) -> None:
     def get(self, label: str) -> Vertex:
         return Vertex._from_java_vertex(self.unwrap().get(_VertexLabel(label).unwrap()))
 
+    def get_contents(self) -> Dict[str, Vertex]:
+
+        def get_unqualified_name_or_proxy_name(key, vertex) -> str:
+            if is_instance_of(k._gateway, vertex, "io.improbable.keanu.vertices.ProxyVertex"):
+                return "proxy_for." + key.getUnqualifiedName()
```
icantsayrural: Is this for when a proxy vertex is added to the sequence by the user? Can you add an example in the tests?

keanu-project/src/main/java/io/improbable/keanu/templating/SequenceItem.java
icantsayrural: So a user can't mutate the contents with what's returned

keanu-python/keanu/sequence.py
CharlieCrisp: So with sequences you have to add a proxy vertex which represents the value of a vertex in the last sequence item and then you redeclare (and add to the sequence item) the newer version of that vertex from this sequence. 
This means you will necessarily have two versions of an input/output vertex's label in a sequence item.

For instance see the last test in `test_sequence.py`. Does that make sense @icantsayrural

keanu-python/tests/test_sequence.py
```diff
@@ -191,3 +193,26 @@ def factory2(sequence_item):
...
+def test_last_item_retrieved_correctly() -> None:
+    x_label = "x"
+    x_input_label = Sequence.proxy_for(x_label)
+
+    def factory(sequence_item):
+        x = DoubleProxy((), x_input_label)
+        x_out = x * Const(2.0)
+        x_out.set_label(x_label)
+        sequence_item.add(x_out)
+        sequence_item.add(x)
```
CharlieCrisp: @icantsayrural here is an example of two items, both with the inner label namespace "x", being added to the same sequence item. This is necessary as otherwise the sequence building can't function properly.

CharlieCrisp: If `get_contents()` were to return a map with just the inner namespace of vertices as the keys then it's not clear or defined whether the proxy or output vertex would be the one keyed by "x".

icantsayrural: Nice. Can we check if the contents have `proxy_for.x_label`?

icantsayrural: Otherwise lgtm!

keanu-python/tests/test_sequence.py
```diff
@@ -2,6 +2,8 @@
 
 import pytest
 
+from keanu import BayesNet
+from keanu.network_io import DotSaver
```
icantsayrural: not used

PR #548, opened 2019-03-27
Better Sequences API part 1

CharlieCrisp:
This PR does two things:
 1. Adds a `sequenceItem.add<Double/Integer/Boolean>ProxyFor()` method that allows proxy vertices to be generated more easily.
 2. Throws exceptions whenever the label of a proxy vertex is changed after construction - this shouldn't ever happen and should allow some issues I've had when using sequences to be found closer to the source of the problem - i.e. when proxy vertices are relabelled.


keanu-project/src/main/java/io/improbable/keanu/templating/SequenceBuilder.java
```diff
@@ -42,10 +45,39 @@
         Sequence build();
     }
 
-    public static VertexLabel proxyFor(VertexLabel label) {
+    public static VertexLabel proxyLabelFor(VertexLabel label) {
         return label.withExtraNamespace(PROXY_LABEL_MARKER);
     }
 
+    public static DoubleProxyVertex doubleProxyFor(VertexLabel label) {
```
CharlieCrisp: @migwellian how would you feel about this method being a non-static method on each `SequenceItem`? This would mean we could actually also call the `sequenceItem.add(proxy)` and remove another step for the developer.

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/nonprobabilistic/BooleanProxyVertex.java
```diff
@@ -28,12 +29,25 @@ public BooleanProxyVertex(VertexLabel label) {
...
+        setLabel(label);
+    }
+
+    @Override
+    public void addScopeToLabel(String scope) {
+        super.setLabel(this.getLabel().withExtraNamespace(scope));
+    }
+
+    @Override
+    public <V extends Vertex<BooleanTensor>> V setLabel(VertexLabel label) {
```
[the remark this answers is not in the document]
CharlieCrisp: I'm not sure that makes sense for a boolean vertex to extend a double vertex. We could put this as a default method in the ProxyVertex interface but we would then have to make sure all ProxyVertices _extend_ Vertex (so they can override .setLabel) which is conceptually different to the current interface for ProxyVertices.

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/nonprobabilistic/BooleanProxyVertex.java
```diff
@@ -28,12 +29,25 @@ public BooleanProxyVertex(VertexLabel label) {
...
+        if (this.getLabel() != null) {
+            throw new RuntimeException("You should not set the label on a Proxy Vertex");
+        }
+        return super.setLabel(label);
     }
 
-    @ExportVertexToPythonBindings
```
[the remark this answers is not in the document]
CharlieCrisp: We definitely want to split the string value into namespaces otherwise the implementation breaks.

keanu-python/keanu/sequence.py
```diff
@@ -108,11 +109,32 @@ def get_last_item(self) -> SequenceItem:
...
-        proxy_label_java = k.jvm_view().SequenceBuilder.proxyFor(label_java)
+        proxy_label_java = k.jvm_view().SequenceBuilder.proxyLabelFor(label_java)
         return proxy_label_java.getQualifiedName()
+
+    @staticmethod
+    def double_proxy_for(label: str, shape: Collection[int] = ()) -> Vertex:
+        """
+        Creates a proxy vertex for the given label
+        """
+        return DoubleProxy(shape, Sequence.proxy_label_for(label))
```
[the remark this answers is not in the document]
CharlieCrisp: This matches the previous implementation - it is split into namespaces by the DoubleProxy constructor which passes it to `_VertexLabel` to do the splitting

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/nonprobabilistic/BooleanProxyVertex.java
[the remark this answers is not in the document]
icantsayrural: Ah, my bad: ** `Vertex`

keanu-python/keanu/sequence.py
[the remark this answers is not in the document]
icantsayrural: Ah, I missed this. Nice :+1:

keanu-project/src/main/java/io/improbable/keanu/templating/SequenceBuilder.java
migwellian: Yes, that's a good idea

keanu-project/src/main/java/io/improbable/keanu/templating/SequenceItem.java
```diff
@@ -54,7 +54,11 @@ public SequenceItem() {
             throw new IllegalArgumentException("Key " + label + " already exists");
         }
         contents.put(label, v);
-        v.setLabel(label);
+        if (v instanceof ProxyVertex) {
+            ((ProxyVertex) v).addScopeToLabel(getUniqueName());
```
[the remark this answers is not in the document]
CharlieCrisp: This was because if we call `setLabel` on a ProxyVertex to add the scope (This is the only time it's acceptable to call `setLabel on a proxy vertex`) then it will throw an exception. The `addScopeToLabel` is a safe way of doing the same thing.

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/nonprobabilistic/BooleanProxyVertex.java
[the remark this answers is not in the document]
CharlieCrisp: Ah I see. It still stands that it would be a bit of a conceptual change from the current implementation which goes more down the 'Composition over inheritance' route.

keanu-project/src/main/java/io/improbable/keanu/templating/SequenceItem.java
migwellian: I don't like the idea of complicating the `Proxy` interface with this additional method. I understand you've put it here because you don't want to allow `setLabel` on a proxy. Instead I think it would be reasonable to replace your logic

        if (this.getLabel() != null) {
            throw new RuntimeException("You should not set the label on a Proxy Vertex");
        }

with

        if (this.getLabel() != null && !this.getLabel().getUnqualifiedName().equals(label.getUnqualifiedName())) {
            throw new RuntimeException("You should not set the label on a Proxy Vertex");
        }

codegen/src/main/java/io/improbable/keanu/codegen/python/vertex/JavaVertexToPythonConverter.java
```diff
@@ -121,6 +122,9 @@ private String toTypedParam(Class<?> parameterClass) {
             return "Collection[int]";
         } else if (Vertex[].class.isAssignableFrom(parameterType)) {
             return "Collection[Vertex]";
+        } else if (VertexLabel.class.isAssignableFrom(parameterType)) {
+            //In python we treat vertex labels as simple strings
```
icantsayrural: nit: I don't think we really need these comments

keanu-project/src/main/java/io/improbable/keanu/templating/loop/Loop.java
```diff
@@ -123,7 +122,7 @@ public static LoopBuilder withInitialConditions(VertexDictionary initialState) {
     }
 
     private void checkIfMaxCountHasBeenReached(SequenceItem item) throws LoopDidNotTerminateException {
-        BooleanVertex stillLooping = item.get(STILL_LOOPING_LABEL);
+        BooleanVertex stillLooping = item.get(LOOP_LABEL);
```
migwellian: why did you change the name?

keanu-project/src/main/java/io/improbable/keanu/vertices/bool/nonprobabilistic/BooleanProxyVertex.java
```diff
@@ -28,12 +29,25 @@ public BooleanProxyVertex(VertexLabel label) {
...
+
+    @Override
+    public void addScopeToLabel(String scope) {
+        super.setLabel(this.getLabel().withExtraNamespace(scope));
+    }
+
+    @Override
+    public <V extends Vertex<BooleanTensor>> V setLabel(VertexLabel label) {
+        if (this.getLabel() != null) {
+            throw new RuntimeException("You should not set the label on a Proxy Vertex");
```
migwellian: I think a more instructive error message is "You should not **change** the label on a Proxy Vertex"

keanu-project/src/test/java/io/improbable/keanu/templating/SequenceBuilderTest.java
```diff
@@ -537,6 +532,12 @@ public void youCanCreateSequenceFromMultipleFactories() {
             .build();
 
         assertThat(sequence.size(), equalTo(5));
+
+        VertexLabel x1InputLabel = SequenceBuilder.proxyLabelFor(x1Label);
+        VertexLabel x2InputLabel = SequenceBuilder.proxyLabelFor(x2Label);
+        VertexLabel x3InputLabel = SequenceBuilder.proxyLabelFor(x3Label);
+        VertexLabel x4InputLabel = SequenceBuilder.proxyLabelFor(x4Label);
```
migwellian: you can inline these now

keanu-project/src/test/java/io/improbable/keanu/vertices/ProxyVertexTest.java
```diff
@@ -0,0 +1,13 @@
+package io.improbable.keanu.vertices;
+
+import io.improbable.keanu.vertices.intgr.nonprobabilistic.IntegerProxyVertex;
+import org.junit.Test;
+
+public class ProxyVertexTest {
+
+    @Test(expected = RuntimeException.class)
```
migwellian: please use a `@Rule ExpectedException` so that you assert on the message too

There's examples in other test files that you can copy.

keanu-python/keanu/sequence.py
migwellian: I'd prefer you to call `k.jvm_view().SequenceBuilder.doubleProxyFor` here. Otherwise you are duplicating logic in java and python.

keanu-python/keanu/vertex/base.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, val_or_ctor: Union[JavaMember, JavaObject], label: Optional[s
             val = typing_cast(JavaObject, val_or_ctor)
 
         super(Vertex, self).__init__(val)
-        if label is not None:
+        if label is not None and self.get_label() is None:
```
migwellian: when/how would `self.get_label()` not be `None`?

keanu-python/tests/test_sequence.py
```diff
@@ -181,6 +177,11 @@ def factory2(sequence_item):
 
     assert sequence.size() == 5
 
+    x1_input_label = Sequence.proxy_label_for(x1_label)
+    x2_input_label = Sequence.proxy_label_for(x2_label)
+    x3_input_label = Sequence.proxy_label_for(x3_label)
+    x4_input_label = Sequence.proxy_label_for(x4_label)
```
migwellian: likewise, you can inline these

keanu-python/keanu/vertex/base.py
icantsayrural: When `label` is a required label for a vertex, it is set twice - by the constructor and here (this is currently how default labels are implemented).

For proxy vertices here it'll throw if you try to set a label after construction.

keanu-project/src/main/java/io/improbable/keanu/templating/loop/Loop.java
CharlieCrisp: no idea. I don't actually remember doing it. Will change it back

keanu-project/src/main/java/io/improbable/keanu/templating/SequenceItem.java
CharlieCrisp: Nice. I didn't want to add a method but couldn't think of a better way of checking. This is much better.

keanu-project/src/main/java/io/improbable/keanu/templating/SequenceItem.java
```diff
@@ -121,4 +124,85 @@ public SequenceItem withExtraEntries(Map<VertexLabel, Vertex<?>> extraEntries) {
...
+     * @param label the label of the corresponding vertex from the previous item.
+     * @param shape the shape of the corresponding vertex from the previous item.
+     * @return a newly created {@link ProxyVertex}
+     */
+    public BooleanProxyVertex addBooleanProxyFor(VertexLabel label, long[] shape) {
+        VertexLabel vertexLabel = SequenceBuilder.proxyLabelFor(label);
+        BooleanProxyVertex newVertex = new BooleanProxyVertex(shape, vertexLabel);
+        this.add(newVertex);
+        return newVertex;
+    }
```
migwellian: you can collapse these into a single generic method:

    public <T extends Vertex<?>> T addProxyFor(VertexLabel label, Function<VertexLabel, T> factoryMethod) {
        VertexLabel proxyLabel = SequenceBuilder.proxyLabelFor(label);
        T newVertex = factoryMethod.apply(proxyLabel);
        this.add(newVertex);
        return newVertex;
    }

Then you can call it like:

    addProxyFor(label, DoubleProxyVertex::new);

Or: 

    addProxyFor(label, (proxyLabel) -> new IntegerProxyVertex(shape, proxyLabel));

etc.

keanu-python/keanu/vertex/base.py
migwellian: Ah I get it. That's quite a subtle point (the distinction between the Java ctor and the Python ctor).

keanu-project/src/main/java/io/improbable/keanu/templating/SequenceItem.java
CharlieCrisp: @migwellian I like this - very elegant. In python, do you think it would be better to remain having three different methods for each type or is there a similar way of wrapping them into one function?

CharlieCrisp: Actually to be honest I'm don't really know how to pass references to methods via Py4J at all

PR #549, opened 2019-03-28
Sequences labels with timesteps

CharlieCrisp:
This PR introduces a change to Sequences such that they can have unique namespace identifiers and also labels which include timesteps. Additionally I've added `getVerticesIgnoringNamespace()` to `BayesianNetwork` with an example use in `SequenceExample.java` to explain its use. 

# Reasoning
## Timesteps in labels
Before, timesteps were only identified by the hashcode of the sequence item. 
This made debugging difficult because it was not obvious what order timesteps are in. For instance when you are taking sample dataframes and you get samples for `134321231.x`, `28384492.x`, etc. and you don't know which represents each timestep.
This is even more of an issue if you use save/load because there is no way of knowing what timestep you are on without the original `Sequence` object.

This change means that labels will be in the form `Sequence_Item_<timestep>.<hashcode>.<label>`, so that you can just use `.getVerticesInNamespace("Sequence_Item_<timestep>")` to retrieve vertices from a timestep.

## Namespace identifiers
Imagine you have constructed a bayesnet from two sequences and you have saved/loaded it from MIR. 
Now you will have two vertices for each timestep, which can only be told apart by their hashcode. You should be able to include information in the labels so you can semantically tell apart the `Sequence_item_0.<hashcode>.x` label for one sequence from the other. By adding a 'identifyingNamespace` you can do this because your labels will be `sequence1.Sequence_item_0.<hashcode>.x` and `sequence2.Sequence_item_0.<hashcode>.x`.


docs/src/test/java/io/improbable/snippet/SequenceExample.java
```diff
@@ -112,10 +114,26 @@ public Sequence buildSequenceTimeSeries() {
...
             .withFactory(factory)
             .build();
 
+        //We can now do a few things:
+        Vertex<?> seedVertex = sequence.getLastItem().get(x1Label);
+        BayesianNetwork network = new BayesianNetwork(seedVertex.getConnectedGraph());
+
+        //Within `network` our vertices will have the labels of the form:
+        //Keanu-Example.Sequence_Item_<<timestep>>.<<unique-hash>>.<<label>>
+        List<Vertex> allXVertices = network.getVerticesIgnoringNamespace(x1Label.getUnqualifiedName());
```
migwellian: needs an extra line of explanation e.g.

    // You can get all the vertices with a particular name, regardless of which SequenceItem they belong to.

keanu-project/src/main/java/io/improbable/keanu/templating/SequenceBuilder.java
```diff
@@ -240,6 +252,18 @@ private VertexLabel getDefaultParentLabel(VertexLabel proxyLabel) {
         }
     }
 
+    /**
+     * This function is best described by how it operates on labels passed to it:
+     *  1. `Sequence_Item_TIMESTEP.HASHCODE.proxy_for.LABEL` -> `proxy_for.LABEL`
+     *  2. `IDENTIFYING_NAMESPACE.Sequence_Item_TIMESTEP.HASHCODE.proxy_for.LABEL` ->  `proxy_for.LABEL`
+     */
```
migwellian: nit: `INDEX` instead of `TIMESTEP`

keanu-project/src/main/java/io/improbable/keanu/templating/SequenceBuilder.java
```diff
@@ -240,6 +252,18 @@ private VertexLabel getDefaultParentLabel(VertexLabel proxyLabel) {
...
+    private VertexLabel getProxyLabel(VertexLabel proxyLabel) {
+        if (this.identifyingNamespace != null) {
+            proxyLabel = proxyLabel.withoutOuterNamespace();
+        }
+        return proxyLabel.withoutOuterNamespace().withoutOuterNamespace();
+    }
```
migwellian: I'm not sure this is gonna be sufficient. Users are allowed to add their own namespaces too - although it's unlikely, it is allowed, and once we do more work on model composition it will become more common.

So you need to handle the case  `Sequence_Item_TIMESTEP.HASHCODE.proxy_for.SCOPED.LABEL`

In other words, your logic needs to be "drop the outer namespaces until you reach the `PROXY_LABEL_MARKER`"

keanu-project/src/main/java/io/improbable/keanu/templating/SequenceItem.java
```diff
@@ -20,9 +20,19 @@
     private static Pattern NAME_REGEX = Pattern.compile(NAME_PREFIX + "-?[\\d]+$");
 
     private Map<VertexLabel, Vertex<?>> contents;
+    private int itemPosition;
+    private int uniqueIdentifier;
```
migwellian: better to call it `sequenceIdentifier` so it's clear that it's unique to the Sequence rather than the SequenceItem.

keanu-project/src/main/java/io/improbable/keanu/templating/SequenceItem.java
```diff
@@ -66,12 +76,23 @@ public SequenceItem() {
         return copyOf(this.contents);
     }
 
+    public int getPosition() {
```
migwellian: Maybe `getIndex` - what do you think?

keanu-project/src/main/java/io/improbable/keanu/templating/SequenceItem.java
```diff
@@ -20,9 +20,19 @@
     private static Pattern NAME_REGEX = Pattern.compile(NAME_PREFIX + "-?[\\d]+$");
 
     private Map<VertexLabel, Vertex<?>> contents;
+    private int itemPosition;
+    private int uniqueIdentifier;
+    private String identifyingNamespace;
```
migwellian: I'd probably just call this `sequenceName` - the fact that it's used as the namespace for a VertexLabel is an implementation detail.

keanu-project/src/main/java/io/improbable/keanu/templating/SequenceItem.java
```diff
@@ -66,12 +76,23 @@ public SequenceItem() {
         return copyOf(this.contents);
     }
 
+    public int getPosition() {
+        return this.itemPosition;
+    }
+
     private String getUniqueName() {
```
migwellian: Should be renamed to `getName` since it's no longer unique.

keanu-project/src/main/java/io/improbable/keanu/templating/SequenceItem.java
```diff
@@ -66,12 +76,23 @@ public SequenceItem() {
...
     }
 
     private VertexLabel scoped(VertexLabel label) {
-        return label.withExtraNamespace(getUniqueName());
+        VertexLabel scopedLabel = label
+            .withExtraNamespace(String.valueOf(this.uniqueIdentifier))
+            .withExtraNamespace(getUniqueName());
+
+        if (this.identifyingNamespace != null) {
+            scopedLabel = scopedLabel.withExtraNamespace(this.identifyingNamespace);
```
migwellian: could you please add a test that checks you get an exception when you call `withExtraNamespace(null)`? It will pass because you'll get an exception from `ImmutableList.builder().add(null)`, I just want to make sure the test is there.

keanu-project/src/test/java/io/improbable/keanu/network/BayesianNetworkTest.java
```diff
@@ -152,6 +152,24 @@ public void networkReturnsVerticesInNamespace() {
...
+    @Test
+    public void networkReturnsAllWithLabelInnerLabel() {
+        BooleanVertex a0 = new BernoulliVertex(0.5);
+        BooleanVertex a1 = new BernoulliVertex(0.5);
+        BooleanVertex b0 = new BernoulliVertex(0.5);
+        BooleanVertex c = new BernoulliVertex(0.5);
+
+        a0.setLabel(new VertexLabel("root", "a", "0"));
+        a1.setLabel(new VertexLabel("root", "a", "1"));
+        b0.setLabel(new VertexLabel("root", "b", "0"));
```
migwellian: `setLabel` returns `this` so you can do it in one line, e.g.:

    BooleanVertex a0 = new BernoulliVertex(0.5).setLabel(new VertexLabel("root", "a", "0"));

keanu-project/src/test/java/io/improbable/keanu/network/BayesianNetworkTest.java
```diff
@@ -152,6 +152,24 @@ public void networkReturnsVerticesInNamespace() {
         assertTrue(verticesInNamespace.containsAll(Arrays.asList(a0, a1, b0)));
     }
 
+    @Test
+    public void networkReturnsAllWithLabelInnerLabel() {
```
migwellian: What does "LabelInnerLabel" mean? Do you mean "SameInnerLabel"?

docs/src/test/java/io/improbable/snippet/SequenceExample.java
```diff
@@ -112,10 +114,26 @@ public Sequence buildSequenceTimeSeries() {
 
         Sequence sequence = new SequenceBuilder<Integer>()
             .withInitialState(dictionary)
+            .withIdentifyingNamespace("Keanu-Example")
             .count(5)
             .withFactory(factory)
             .build();
 
+        //We can now do a few things:
+        Vertex<?> seedVertex = sequence.getLastItem().get(x1Label);
```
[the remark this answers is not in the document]
CharlieCrisp: So I guess I've used it to mean "an arbitrary vertex in the graph, from which we can get the connected graph and construct the bayes net".
Agreed this may be confusing although I think `sequenceOutput` could also be confusing because it's actually not relevant that it's an output in this case.
Perhaps we should even add a `.getBayesNet` method to Sequence?? What do you think?

docs/src/test/java/io/improbable/snippet/SequenceExample.java
```diff
@@ -112,10 +114,26 @@ public Sequence buildSequenceTimeSeries() {
...
+        BayesianNetwork network = new BayesianNetwork(seedVertex.getConnectedGraph());
+
+        //Within `network` our vertices will have the labels of the form:
+        //Keanu-Example.Sequence_Item_<<timestep>>.<<unique-hash>>.<<label>>
```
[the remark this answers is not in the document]
CharlieCrisp: Yep - apologies for all these mentions of 'timesteps'. I caught myself doing this a few times but obviously not enough :)

keanu-project/src/main/java/io/improbable/keanu/templating/SequenceItem.java
CharlieCrisp: Yep I think that's more clear :)

keanu-project/src/main/java/io/improbable/keanu/templating/SequenceItem.java
CharlieCrisp: So this is a test on a `VertexLabel`?

keanu-project/src/main/java/io/improbable/keanu/templating/SequenceBuilder.java
CharlieCrisp: @migwellian I believe in this case the code will just return the label `proxy_for.SCOPED.LABEL`. I think this is actually what we want though.
Because then the `proxy_for.SCOPED.LABEL` will be default parent label as intended?

User defined scopes cant be added to form e.g. `Sequence_Item_TIMESTEP.HASHCODE.SCOPED.proxy_for.LABEL`, so I think this might still work.
Or have I just misunderstood your point?

docs/src/test/java/io/improbable/snippet/SequenceExample.java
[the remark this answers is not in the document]
migwellian: Yes, I think `toBayesNet` would be a useful method.

We already have `model.to_bayes_net()` in python so this would tie in well with that.

keanu-project/src/main/java/io/improbable/keanu/templating/SequenceBuilder.java
migwellian: yes you're right. carry on.

keanu-python/keanu/net.py
```diff
@@ -15,10 +18,12 @@
 
 class BayesNet(JavaObjectWrapper):
 
-    def __init__(self, vertices: Iterable[Any]) -> None:
-        java_vertices = k.to_java_object_list(vertices)
-
-        super(BayesNet, self).__init__(k.jvm_view().BayesianNetwork(java_vertices))
+    def __init__(self, vertices_or_java_object: Union[Iterable[Any], JavaObject]) -> None:
```
CharlieCrisp: @icantsayrural @migwellian I'm currently changing this constructor so I can call `BayesNet(java_bayes_net)` in `Sequence.to_bayes_net()` but is there a better way of doing this?

icantsayrural: I don't think so, this is how we do it usually. It is a bit weird here since `BayesNet` is a public class and we'd usually want to hide the java part as much as possible (the parameter `vertices_or_java_object`)... but I don't think there is a better solution.

In the past, I've tried creating a separate `_BayesNet` class that just takes the java object in the constructor and having `BayesNet` extend that. But this just adds unnecessary complexity for so little gain.

CharlieCrisp: Okay I've gone with a slightly different approach of constructing a JavaObjectWrapper directly and updating the instance's `__class__` from within the `to_bayes_net()` method.

icantsayrural: :+1:

keanu-project/src/main/java/io/improbable/keanu/templating/SequenceBuilder.java
```diff
@@ -23,6 +23,7 @@
     private static final String PROXY_LABEL_MARKER = "proxy_for";
     private VertexDictionary initialState;
     private Map<VertexLabel, VertexLabel> transitionMapping = Collections.emptyMap();
+    private String identifyingNamespace;
```
migwellian: nit: could you please rename this to `sequenceName`, as you did with the same field in `SequenceItem`

PR #559, opened 2019-05-08
Faster tensor creation in Python

CharlieCrisp:
This python uses byte arrays to pass double and integer arrays to Java so that we get a roughly 400x speed up!


keanu-python/keanu/tensor.py
```diff
@@ -51,11 +50,26 @@ def get_tensor_type(self) -> type:
...
 
         ctor = Tensor.__infer_tensor_ctor_from_ndarray(ndarray)
-        values = k.to_java_array(ndarray.flatten().tolist())
+        values = Tensor.__get_java_array_from_ndarray(ndarray)
         shape = k.to_java_long_array(ndarray.shape)
 
         return ctor(values, shape)
 
+    @staticmethod
+    def __get_java_array_from_ndarray(ndarray: numpy_types) -> Union[JavaObject, bytes]:
```
migwellian: is there a way to make it always return `bytes` and never `JavaObject`?

keanu-project/src/main/java/io/improbable/keanu/util/Py4jByteArrayConverter.java
```diff
@@ -40,4 +40,20 @@
         byte[] byteArray = booleanBuffer.array();
         return byteArray;
     }
+
+    public static double[] toDoubleArray(byte[] byteArray, int itemSizeBytes){
```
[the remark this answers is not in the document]
CharlieCrisp: So I was actually using this before and the tests were failing because the size of integers in python was different to what the JVM considers an integer. I therefore used this approach for integers and additionally used it for doubles for consistency and in case the same problem occurred with python floats and differing sizes. 
@migwellian does that make sense?

keanu-python/keanu/tensor.py
CharlieCrisp: I've now done it so it only returns JavaObject, which I think still captures the spirit of this comment?

keanu-project/src/main/java/io/improbable/keanu/tensor/dbl/DoubleTensor.java
```diff
@@ -41,6 +41,7 @@ public static DoubleTensor create(double... values) {
         return create(values, values.length);
     }
 
+
```
icantsayrural:
```suggestion
```

keanu-project/src/main/java/io/improbable/keanu/util/Py4jByteArrayConverter.java
```diff
@@ -40,4 +40,20 @@
...
+        double[] doubles = new double[byteArray.length / itemSizeBytes];
+        for(int i=0;i<doubles.length;i++){
```
icantsayrural:
```suggestion
        for (int i = 0; i < doubles.length; i++) {
```

keanu-project/src/main/java/io/improbable/keanu/util/Py4jByteArrayConverter.java
```diff
@@ -40,4 +40,20 @@
         byte[] byteArray = booleanBuffer.array();
         return byteArray;
     }
+
+    public static double[] toDoubleArray(byte[] byteArray, int itemSizeBytes){
+        double[] doubles = new double[byteArray.length / itemSizeBytes];
```
icantsayrural: Should probably check `itemSizeBytes` is not zero

keanu-project/src/main/java/io/improbable/keanu/util/Py4jByteArrayConverter.java
```diff
@@ -40,4 +40,20 @@
...
+        for(int i=0;i<doubles.length;i++){
+            doubles[i] = ByteBuffer.wrap(byteArray, i*itemSizeBytes, itemSizeBytes).order(ByteOrder.LITTLE_ENDIAN).getDouble();
+        }
+        return doubles;
+    }
+
+    public static int[] toIntegerArray(byte[] byteArray, int itemSizeBytes){
+        int[] ints = new int[byteArray.length / itemSizeBytes];
+        for(int i=0;i<ints.length;i++){
```
icantsayrural:
```suggestion
        for(int i = 0; i < ints.length; i++) {
```

keanu-project/src/main/java/io/improbable/keanu/util/Py4jByteArrayConverter.java
```diff
@@ -40,4 +40,20 @@
...
+    public static double[] toDoubleArray(byte[] byteArray, int itemSizeBytes){
+        double[] doubles = new double[byteArray.length / itemSizeBytes];
+        for(int i=0;i<doubles.length;i++){
+            doubles[i] = ByteBuffer.wrap(byteArray, i*itemSizeBytes, itemSizeBytes).order(ByteOrder.LITTLE_ENDIAN).getDouble();
+        }
+        return doubles;
+    }
+
+    public static int[] toIntegerArray(byte[] byteArray, int itemSizeBytes){
+        int[] ints = new int[byteArray.length / itemSizeBytes];
```
icantsayrural: Should probably check `itemSizeBytes` is not zero

keanu-project/src/test/java/io/improbable/keanu/util/Py4jByteArrayConverterTest.java
```diff
@@ -0,0 +1,44 @@
...
+        byte byte2 = Byte.parseByte("00000011", 2);
+        byte zeroByte = Byte.parseByte("00000000", 2);
+        byte[] bytes = new byte[] {
+            byte1, zeroByte, zeroByte, zeroByte,
+            byte2, zeroByte, zeroByte, zeroByte
+        };
+
+        int[] ints = Py4jByteArrayConverter.toIntegerArray(bytes, 4);
+
+        assertThat(ints.length, is(2));
```
icantsayrural: You can do `assertArrayEquals`

keanu-project/src/test/java/io/improbable/keanu/util/Py4jByteArrayConverterTest.java
```diff
@@ -0,0 +1,44 @@
...
+        byte byte3 = Byte.parseByte("00001000", 2);
+        byte zeroByte = Byte.parseByte("00000000", 2);
+        byte[] bytes = new byte[] {
+            zeroByte, zeroByte, zeroByte, zeroByte, zeroByte, zeroByte, byte2, byte1,
+            zeroByte, zeroByte, zeroByte, zeroByte, zeroByte, zeroByte, byte3, byte1
+        };
+
+        double[] doubles = Py4jByteArrayConverter.toDoubleArray(bytes, 8);
+
+        assertThat(doubles.length, is(2));
```
icantsayrural: You can do `assertArrayEquals`

keanu-python/keanu/tensor.py
migwellian: 👍

keanu-project/src/main/java/io/improbable/keanu/util/Py4jByteArrayConverter.java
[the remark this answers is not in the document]
migwellian: I see - looks like Python (numpy?) ints are 8 bytes.

So how about we have separate methods `toIntegerArray` and `toLongArray` with hard-coded item sizes, and from Python you call `toLongArray`?

keanu-project/src/main/java/io/improbable/keanu/util/Py4jByteArrayConverter.java
```diff
@@ -40,4 +41,53 @@
...
+        for(int i = 0; i < ints.length; i++) {
+            ints[i] = ByteBuffer.wrap(byteArray, i*itemSizeBytes, itemSizeBytes).order(ByteOrder.LITTLE_ENDIAN).getInt();
+        }
+        return ints;
+    }
+
+    public static boolean[] toBooleanArray(byte[] byteArray, int numberOfBooleansInArray) {
+        BitSet bits = BitSet.valueOf(byteArray);
+        boolean[] bools = new boolean[numberOfBooleansInArray];
+        int numberOfBitsToIterateThrough = ((int) Math.ceil((numberOfBooleansInArray - 1.0) / 8.0) + 1) * 8;
```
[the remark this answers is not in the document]
migwellian: +1 - this needs pulling out into a separate function that describes what it does.

keanu-project/src/main/java/io/improbable/keanu/util/Py4jByteArrayConverter.java
```diff
@@ -40,4 +41,53 @@
...
+        }
+
+        int[] ints = new int[byteArray.length / itemSizeBytes];
+        for(int i = 0; i < ints.length; i++) {
+            ints[i] = ByteBuffer.wrap(byteArray, i*itemSizeBytes, itemSizeBytes).order(ByteOrder.LITTLE_ENDIAN).getInt();
+        }
+        return ints;
+    }
+
+    public static boolean[] toBooleanArray(byte[] byteArray, int numberOfBooleansInArray) {
```
migwellian: why are the semantics different here (`numberOfBooleansInArray` instead of `itemSizeBytes`)? You're just passing in `byteArray.size * <size of a byte>`

keanu-project/src/main/java/io/improbable/keanu/util/Py4jByteArrayConverter.java
```diff
@@ -40,4 +41,53 @@
...
+    }
+
+    private int getBigEndianPosition(int index){
+        int byteNumber = index / 8;
+        int bitNumber = index % 8;
+        return byteNumber * 8 + (7 - bitNumber);
+    }
+}
+
+class ArrayConversionException extends Exception {
```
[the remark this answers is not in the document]
CharlieCrisp: Have actually now removed it due to the removal of `itemSizeBytes` :)

keanu-project/src/main/java/io/improbable/keanu/util/Py4jByteArrayConverter.java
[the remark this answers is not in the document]
CharlieCrisp: I've actually realised this was completely unnecessary - what I was actually looking for was `byteArray.length * 8` to get the total number of bits in the array.

keanu-project/src/main/java/io/improbable/keanu/util/Py4jByteArrayConverter.java
```diff
@@ -40,4 +41,62 @@
...
+            longs[i] = ByteBuffer.wrap(byteArray, i * longSizeBytes, longSizeBytes).order(ByteOrder.LITTLE_ENDIAN).getLong();
+        }
+        return longs;
+    }
+
+    /**
+     * This function is intended to be called from the Python API with the result of np.packbits(...).toBytes().
+     * In Python, the result of this is a byte array where each boolean is represented by a single bit (0 for false, 1 for true).
+     * When put into a bytearray, these bits may need to be padded with up to 7 bits to make a whole number of bytes.
+     * Therefore, the function requires the parameter numberOfBooleansInArray to know which bits are padding bits and which are booleans.
```
migwellian: good explanation

keanu-project/src/main/java/io/improbable/keanu/tensor/intgr/IntegerTensor.java
```diff
@@ -41,6 +42,26 @@ static IntegerTensor create(int... values) {
         return create(values, values.length);
     }
 
+    /**
+     * Creates an IntegerTensor from a long[]. Will throw an exception is any value in the long[] cannot be represented as an integer
```
icantsayrural:
```suggestion
     * Creates an IntegerTensor from a long[]. Will throw an exception if any value in the long[] cannot be represented as an integer
```

keanu-project/src/main/java/io/improbable/keanu/tensor/intgr/IntegerTensor.java
```diff
@@ -41,6 +42,26 @@ static IntegerTensor create(int... values) {
...
+     * @param shape shape of the tensor
+     * @return a new IntegerTensor
+     */
+    static IntegerTensor create(long[] values, long... shape) {
+        int[] ints = Arrays.stream(values).mapToInt(Ints::checkedCast).toArray();
+        return IntegerTensor.create(ints, shape);
+    }
+
+    /**
+     * Creates an IntegerTensor from a long[]. Will throw an exception is any value in the long[] cannot be represented as an integer
```
icantsayrural:
```suggestion
     * Creates an IntegerTensor from a long[]. Will throw an exception if any value in the long[] cannot be represented as an integer
```

keanu-project/src/main/java/io/improbable/keanu/util/Py4jByteArrayConverter.java
```diff
@@ -40,4 +41,62 @@
         byte[] byteArray = booleanBuffer.array();
         return byteArray;
     }
+
+    public static double[] toDoubleArray(byte[] byteArray) {
+        int doubleSizeBytes = Double.SIZE / Byte.SIZE;
```
icantsayrural: Can these be constant?

keanu-project/src/main/java/io/improbable/keanu/util/Py4jByteArrayConverter.java
```diff
@@ -40,4 +41,62 @@
...
+     * Therefore, the function requires the parameter numberOfBooleansInArray to know which bits are padding bits and which are booleans.
+     * @param byteArray an array of bytes where booleans (in little endian form) are represented by single bits.
+     * @param numberOfBooleansInArray the number of booleans represented by the bits in the bytearray
+     * @return an array of booleans
+     */
+    public static boolean[] toBooleanArray(byte[] byteArray, int numberOfBooleansInArray) {
+        BitSet bits = BitSet.valueOf(byteArray);
+        boolean[] bools = new boolean[numberOfBooleansInArray];
+        int totalNumberOfBits = byteArray.length * 8;
+        for (int i = bits.nextSetBit(0); i != -1 && i < totalNumberOfBits; i = bits.nextSetBit(i+1)) {
```
icantsayrural: Why do we need this check `i < totalNumberOfBits`? Surely the example loop in https://docs.oracle.com/javase/7/docs/api/java/util/BitSet.html#nextSetBit(int) should be enough?

keanu-project/src/main/java/io/improbable/keanu/util/Py4jByteArrayConverter.java
```diff
@@ -40,4 +41,62 @@
...
+            int position = getBigEndianPosition(i);
+            if (position < numberOfBooleansInArray) {
```
icantsayrural: Should we throw an exception if this is not the case?

keanu-project/src/test/java/io/improbable/keanu/util/Py4jByteArrayConverterTest.java
```diff
@@ -0,0 +1,87 @@
...
+        boolean[] desiredBools = new boolean[] {
+            true, false, true, false, true, false, true, false,
+            true, false, false, false, false, false, false, false
+        };
+
+        assertArrayEquals(bools, desiredBools);
+    }
+
+    @Test
+    public void doesntCreateTooManyBooleansFromPadding() {
```
icantsayrural: Nice

keanu-project/src/main/java/io/improbable/keanu/util/Py4jByteArrayConverter.java
CharlieCrisp: No - in the case that the number of booleans does not fit exactly into n bytes (i.e. the number of bools is not a multiple of 8) then this case can occur completely validly. What it means, is that we are parsing a 'padding bit' and that we should ignore it. Because of the endianness, this could occur whilst there are still further valid bits yet to parse.

CharlieCrisp: @icantsayrural as you pointed out the above is wrong. Any extra padding bits should be zero and therefore not appear here as a `nextSetBit`. The behaviour will now throw an error if any set bits are found that are longer than the number of booleans to be encoded.
This behaviour is also now tested

keanu-project/src/main/java/io/improbable/keanu/util/Py4jByteArrayConverter.java
```diff
@@ -4,9 +4,13 @@
 
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
+import java.util.BitSet;
 
 @UtilityClass
 public class Py4jByteArrayConverter {
+    private final int LONG_SIZE_BYTES = Long.SIZE / Byte.SIZE;
```
icantsayrural: Isn't this just `Long.BYTES`?

keanu-project/src/test/java/io/improbable/keanu/tensor/intgr/Nd4jIntegerTensorTest.java
```diff
@@ -274,6 +274,15 @@ public void doesSetWithMask() {
         assertArrayEquals(expected, matrix.asFlatIntegerArray());
     }
 
+    @Test
+    public void cannotCreateTensorWithLongsThatAreTooLong() {
+        thrown.expect(IllegalArgumentException.class);
+        thrown.expectMessage("Out of range: " + Long.MAX_VALUE);
+
```
icantsayrural:
```suggestion
```

PR #564, opened 2019-06-04
Multinomial too strict on p and fix multinomial broadcast

gordoncaleb:
This PR cleans up the MultinomialVertex. The issues with it before were:

- It was extremely strict on checking that p summed to 1.0
- It disallowed scalar n. This is probably the most common use case.
- It didn't support the usual batch sampling nor batch logProb (see MultinomialVertex.java for docs)

Functional changes:
- The shape of p expected k to be the far left dimension. It is now the far right in order to allow for broadcasting semantics.
- n and p parameter validation was semi-controllable but is now completely toggleable with `vertex.setValidationEnabled(true);`


keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
```diff
@@ -1,116 +1,92 @@
...
+
+        double[] flatP = broadcastedP.asFlatDoubleArray();
+        int[] flatN = broadcastedN.asFlatIntegerArray();
+
+        int sampleCount = flatN.length;
+        int[] samples = new int[numCategories * sampleCount];
+
+        for (int i = 0; i < sampleCount; i++) {
+            int pIndex = i * numCategories;
+            drawNTimes(flatN[i], random, samples, pIndex, flatP, pIndex, numCategories);
```
GeorgeNash: you pass in `pIndex` twice, we don't need that

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
```diff
@@ -1,116 +1,92 @@
...
+        return IntegerTensor.create(samples, shape);
     }
 
-    private static int[] drawNTimes(int n, KeanuRandom random, Double... categoryProbabilities) {
-        int[] categoryDrawCounts = new int[categoryProbabilities.length];
+    private static void drawNTimes(int n, KeanuRandom random, int[] sample, int sampleIndex, double[] p, int pIndex, int pCount) {
         for (int i = 0; i < n; i++) {
-            int index = draw(random, categoryProbabilities);
-            categoryDrawCounts[index] += 1;
+            int index = draw(random, p, pIndex, pCount);
```
GeorgeNash: `indexOfDrawnSample` or something? not clear what this index is

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
```diff
@@ -1,116 +1,92 @@
...
 
-    private static int draw(KeanuRandom random, Double... categoryProbabilities) {
+    private static int draw(KeanuRandom random, double[] p, int pIndex, int pCount) {
         double value = random.nextDouble();
         int index = 0;
         Double pCumulative = 0.;
-        while (index < categoryProbabilities.length) {
-            Double currentP = categoryProbabilities[index++];
+        while (index < pCount) {
+            Double currentP = p[pIndex + index++];
```
GeorgeNash: this is carnage can you do the incrementing separately - and why are we using boxed types?

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
```diff
@@ -123,26 +99,95 @@ private static int draw(KeanuRandom random, Double... categoryProbabilities) {
...
-            String.format("Shape mismatch. k: %s, p: %s",
-                Arrays.toString(k.getShape()),
-                Arrays.toString(expectedShape)),
-            k.getShape(), expectedShape
-        );
-        Preconditions.checkArgument(
-            k.sum(0).elementwiseEquals(this.n).allTrue(),
-            String.format("Inputs %s must sum to n = %s", k, this.n)
+    public DoubleTensor logProb(IntegerTensor x) {
+        if (validationEnabled) {
```
GeorgeNash: I don't really get this validation thing, why not just validate on construction and then only validate the things that change?

e.g: only call:

```
validateBroadcastShapes(x.getShape(), n.getShape(), p.getShape());
validateX(x, n, p);
```

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
```diff
@@ -199,6 +199,28 @@ public static boolean isLengthOne(long[] shape) {
         return (shape1.length >= shape2.length) ? shape1 : shape2;
     }
 
+    public static long[] getBroadcastResultShape(long[] left, long[] right) {
+
+        final long[] highestRank = left.length > right.length ? left : right;
```
GeorgeNash: `shapeOfHighestRank`

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
```diff
@@ -199,6 +199,28 @@ public static boolean isLengthOne(long[] shape) {
...
+        long[] resultShape = Arrays.copyOf(highestRank, highestRank.length);
+
+        int lowestRank = Math.min(left.length, right.length);
+        for (int i = 1; i <= lowestRank; i++) {
```
GeorgeNash: why not call `public static boolean isBroadcastable(long[] left, long[] right)`?

keanu-project/src/main/java/io/improbable/keanu/vertices/intgr/probabilistic/MultinomialVertex.java
```diff
@@ -15,51 +15,110 @@
...
+ * If the number of categories is defined by k, then the shape of p is (a...b, k) where a..b represents any shape of
+ * any rank. For the p as a vector case, a...b is rank 0 and would be just a shape (k). Given that p has a shape of
+ * (a...b, k) then n can have any shape that is broadcastable with a...b. The resulting shape would be the broadcasted
+ * n shape with a...b and end in k.
+ * e.g.
+ * n = [[1, 2],[3, 4]] with shape (2, 2)
+ * p = [[0.2, 0.2, 0.6], [0.5, 0.25, 0.25]] with shape 2, 3
+ * therefore k = 3
+ * and the result shape is (2, 2, 3), which is (2, 2) broadcasted with (2) and k appended.
+ */
```
GeorgeNash: v useful!

keanu-project/src/test/java/io/improbable/keanu/vertices/intgr/probabilistic/MultinomialVertexTest.java
```diff
@@ -47,221 +46,436 @@
...
+    }
+
+    @Test
+    public void itWorksWithAlmostOneProbability() {
         int n = 100;
-        DoubleTensor p = DoubleTensor.create(new double[]{0.01, 0.09, 0.9}, 3, 1);
-        MultinomialVertex multinomial = new MultinomialVertex(n, ConstantVertex.of(p));
-        IntegerTensor samples = multinomial.sample(KeanuRandom.getDefaultRandom());
-        assertThat(samples, hasShape(3, 1));
+        DoubleTensor p = DoubleTensor.create(0.1, new long[]{10});
```
GeorgeNash: this is meant to be almost one? This `p` is exactly `1.0`

gordoncaleb: The sum of those 0.1s is actually `0.9999999999999999` due to floating point summation. This was causing this exact case to throw an exception before.

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
gordoncaleb: Only because that would require iterating through the shape twice. Sometimes you want to know if it's broadcastable without throwing an exception and sometimes you want to get the broadcasted shape or throw an exception. You could check isBroadcastable before getting the broadcastable shape but that would iterate twice, which probably wouldn't be a big deal.

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
gordoncaleb: 👍

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
gordoncaleb: Good call. The Multinomial object is never reused after either sample or logProb so I didn't see the point but I've made the change so that we don't force that.

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
```diff
@@ -1,116 +1,92 @@
...
     }
 
-
     @Override
     public IntegerTensor sample(long[] shape, KeanuRandom random) {
-        TensorShapeValidation.checkTensorsMatchNonLengthOneShapeOrAreLengthOne(shape, n.getShape());
 
-        Tensor.FlattenedView<Integer> nFlattened = n.getFlattenedView();
-        List<DoubleTensor> sliced = p.sliceAlongDimension(0, 0, numCategories);
+        if (validationEnabled) {
```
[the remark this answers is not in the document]
gordoncaleb: They'll often just be constants and summing them all the time could be pointless.

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
```diff
@@ -123,26 +103,93 @@ private static int draw(KeanuRandom random, Double... categoryProbabilities) {
...
+
+        DoubleTensor gammaN = n.plus(1).toDouble().logGammaInPlace();
+        DoubleTensor xLogP = p.log().timesInPlace(x.toDouble()).sum(-1);
+        DoubleTensor gammaXs = x.plus(1).toDouble().logGammaInPlace().sum(-1);
+        return xLogP.plusInPlace(gammaN).minusInPlace(gammaXs);
+    }
+
+    private static void validateProbabilities(DoubleTensor p) {
+        if (p.isScalar()) {
+            throw new IllegalArgumentException("Probabilities must be a vector or a tensor with rank >= 1");
```
migwellian: throw a `io.improbable.keanu.tensor.TensorValueException` instead?

Here and elsewhere in the PR.

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
```diff
@@ -1,116 +1,96 @@
...
-        );
-        CATEGORY_PROBABILITIES_CANNOT_BE_ZERO.validate(p);
-
-        numCategories = p.getShape()[0];
+    private Multinomial(IntegerTensor n, DoubleTensor p, boolean validationEnabled) {
+        k = Ints.checkedCast(p.getShape()[p.getRank() - 1]);
         this.n = n;
         this.p = p;
-    }
+        this.validationEnabled = validationEnabled;
```
migwellian: I liked the old way, with the `DebugTensorValidator`. You could implement all these validation checks with `TensorValidator`s.

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
```diff
@@ -123,26 +103,93 @@ private static int draw(KeanuRandom random, Double... categoryProbabilities) {
...
+
+    private static void validateBroadcastShapes(long[] xShape, long[] nShape, long[] pShape) {
+
+        long[] broadcastResultShape;
+        try {
+            broadcastResultShape = TensorShape.getBroadcastResultShape(
+                pShape, TensorShape.concat(nShape, new long[]{1})
+            );
+        } catch (IllegalArgumentException e) {
+            throw new IllegalArgumentException(
```
migwellian: add `e` as the `cause` argument?

keanu-project/src/test/java/io/improbable/keanu/vertices/intgr/probabilistic/MultinomialVertexTest.java
migwellian: you should express that in the test - `assertThat(p.sum(), not(equalTo(1.0))`

keanu-project/src/main/java/io/improbable/keanu/vertices/intgr/probabilistic/MultinomialVertex.java
```diff
@@ -15,51 +15,110 @@
...
+ * e.g.
+ * n = 5 with shape ()
+ * p = [0.2, 0.2, 0.6] with shape (3)
+ * a sample could return x = [1, 3, 1] with shape (3)
+ * and logProb([1, 3, 1]) would be valid
+ * and logProb([[1, 3, 1], [2, 2, 1]]) would be a batch logProb equivalent to logProb([1, 3, 1]) + logProb([2, 2, 1])
+ * <p>
+ * More complex cases are also acceptable and use broadcasting semantics.
+ * <p>
+ * If the number of categories is defined by k, then the shape of p is (a...b, k) where a..b represents any shape of
```
migwellian:
```suggestion
 * If the number of categories is defined by k, then the shape of p is (a...b, k) where a...b represents any shape of
```

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
migwellian: It would be nice to get rid of the code duplication though. How about if `isBroadcastable` were to call `getBroadcastShape`?

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
gordoncaleb: The previous TensorValidator only did a check on the sum of the probabilities. That meant that you could turn off some of the checks but you couldn't turn them off wholesale. Converting all of these checks to TensorValidator could be done but I think delegating the checks to them doesn't make it any clearer nor add any new/needed feature. We would have to write TensorValidators for tensor shape checking which *might* make the shape validation code few lines but I don't think it's worth the extra complexity and extra classes required.

keanu-project/src/main/java/io/improbable/keanu/distributions/discrete/Multinomial.java
gordoncaleb: I think IllegalArgumentException is more informative in this case. The TensorValueException doesn't really tell me more but instead really tells me less. This is definitely the case for checking hyper-parameter shapes or ranks. I thought about creating a ParameterValidationException but the value added is still pretty minimal.

keanu-project/src/main/java/io/improbable/keanu/tensor/TensorShape.java
gordoncaleb: 👍

PR #568, opened 2019-06-08
SonarQube - String Literals on the left on comparisons

dvmarcilio:
Transforming from `someString.equals("string")`  to `"string".equals(someString)`. This avoids checking for `null` and also prevents `NullPointerException` being thrown.

These fixes multiple SonarQube violations of the rule: [Strings literals should be placed on the left side when checking for equality](https://rules.sonarsource.com/java/RSPEC-1132)

---

Also, there is one fix for SonarQube violation of the rule:[ Strings and Boxed types should be compared using "equals()"](https://rules.sonarsource.com/java/RSPEC-4973)

The rule states that is almost always a mistake to compare two instances of `String` with `==`, because it compares location in memory. `equals()` should be used.


keanu-project/src/main/java/io/improbable/keanu/util/status/StatusBar.java
```diff
@@ -109,7 +109,7 @@ private String formatComponents() {
         StringBuilder sb = new StringBuilder();
         for (StatusBarComponent component : components) {
             String render = component.render();
-            if (render != "") {
+            if (!"".equals(render)) {
```
dvmarcilio, line 111: 2 fixes in one line: 
1. replacing `!=` with `equals()`
2. putting the String literal on the left of the comparison

keanu-project/src/main/java/io/improbable/keanu/util/status/StatusBar.java
```diff
@@ -62,7 +62,7 @@ public StatusBar(PrintStream printStream, ScheduledExecutorService scheduler) {
 
         // Status bar is disabled for testing.
         String disableStatusBar = System.getProperty("io.improbable.keanu.util.status.StatusBar.disableStatusBar");
-        if (disableStatusBar != null && disableStatusBar.equals("true")) {
+        if ("true".equals(disableStatusBar)) {
```
dvmarcilio, line 64: When we move the String Literal to the left, we can avoid the null check. 
if `disableStatusBar` happens to be `null`, the comparison will evaluate to `false`

PR #571, opened 2019-07-05
Update windows queue and use scaler.v2.

stefan-improbable:


efarem:
Can we merge this? Also it might need an updated Windows depending on whether the original bake specified an agent_count of 1.

stefan-improbable:
> Can we merge this? Also it might need an updated Windows depending on whether the original bake specified an agent_count of 1.

@efarem last time I tried this, windows CI was broken. I've merged the newest changes from `develop` and made linux use scaler.v2. Let's see if it passes now.

PR #577, opened 2020-05-05
Fix CI (using existing capability)

wesleyw72:
Update queue. Also stop running on Windows.


ashleydavies (approved):
LGTM so we can land the dependabot PRs

